PR #42, opened 2026-04-17
fix: change fill to z

nk1408:


nk1408:
Schema Migration: fill → z

  Renamed the fill property to z throughout the maidr data structure to align with standard
  charting terminology (x, y, z axes).

  Affected chart types:
  - Stacked/Dodged bars: z represents segment/series name
  - Multi-line plots: z represents series name
  - Boxplot/Violin: z represents category/group
  - Heatmap: z represents cell value label

  Files changed: 11 processor files, 1 man page, 5 test files

  ---
  New Option: maidr.use_bundled

  Added user control over MAIDR.js source selection:

  options(maidr.use_bundled = TRUE)   # Always use local bundled files
  options(maidr.use_bundled = FALSE)  # Always use CDN
  options(maidr.use_bundled = NULL)   # Auto-detect (default)

  Use case: Ensures reproducible offline builds when users need guaranteed local file usage
  regardless of internet connectivity.

PR #48, opened 2026-07-24
fix: comprehensive bug and performance sweep across Base R and ggplot2 pipelines

jooyoungseo:
> ### ⚠️ Update after review — some claims below are no longer accurate
>
> A second review round (see the PR comments) verified the claims in this description
> against rendered output and found several that did not hold. The code has been
> corrected; this note corrects the description. Everything not listed here still stands.
>
> - **"Auto-show is now actually wired up" — reverted, not shipped.** Wiring it up broke
>   the most basic Base R idiom: `show()` clears the recorded calls and closes the temp
>   device, so firing it after every top-level expression made `plot(x, y)` followed by
>   `abline(h = 1)` fail with *"plot.new has not been called yet"*. Reproduced in a real
>   interactive session. The wiring was removed; Base R auto-display remains unwired, as
>   before this PR. Re-enabling it needs a non-destructive render path — tracked in #53.
> - **`plot(y ~ x, subset = ...)` was not actually fixed by the arg-forcing change.** It
>   still failed, for a different reason: `plot.formula()` resolves `subset` in
>   `parent.frame()`, which the wrapper displaces. `boxplot(y ~ g, subset = ...)` failed
>   the same way. Both now work via a retry evaluated in the caller's frame.
> - **"scatter point colour/group values are emitted" held only for unfaceted plots.**
>   Faceted scatter emitted raw hex codes (`#F8766D`) instead of category names.
> - **Faceted heat maps reported panel 1's cell values in every panel**, and faceted box
>   plots carried their own quartiles under panel 1's category names.
> - **`matplot(m)` collapsed a multi-series matrix into one fabricated series**
>   (column 2 vs column 1), with series 1's values announced as x coordinates.
> - **A trailing `par(mfrow = c(1, 1))` reset collapsed the whole multi-panel grid**, so
>   the idiomatic cleanup after a 2x2 figure dropped three of the four panels.
> - **Dodged bars crashed on expression aesthetics** such as `aes(fill = factor(cyl))`,
>   and the `stat = "count"` path emitted bars that ggplot2 never draws.
> - **The unused `curl` import was NOT dropped.** `curl` is still in `Imports` and is
>   still used, at `R/html_dependencies.R:107`. The only `DESCRIPTION` change in this
>   branch is adding `tibble` to `Suggests`. Nothing needed dropping — the claim is what
>   was wrong, not the code.
> - **Verification numbers are stale.** The suite is now **2762 pass, 0 fail** (was 2708),
>   and `R CMD check` had an unstated-dependency WARNING that is now fixed.
>
> Also changed: `R-CMD-check` and `test-coverage` now run on pull requests. They were
> `push: main` only, which is why this change reached review with no test run behind it.
>
> **Follow-up issues opened from this review:** #52 (violin inside patchwork emits an
> empty subplot), #53 (auto-display needs a non-destructive `show()`), #54 (stacked/dodged
> series order vs SVG rect order — needs a check against the bundled maidr.js before
> anyone changes the R side), #55 (one `segments()`/`arrows()` overlay drops all accessible
> data for the figure), #56 (`render_maidr({ plot(...) })` renders blank — the new NULL
> short-circuit hides the new Base R auto-detection), #57 (`maidr_off()` leaves recorded
> calls on the device), #58 (internet probe cached for the whole session), #59 (NSE replay
> captures the caller frame by reference), #60 (unverified audit leads, tracking),
> #61 (fixes claimed here that still have no regression test).

---

## Summary

A deep review of the whole package (multi-agent audit with adversarial cross-verification of every finding, plus empirical end-to-end validation against the actual exported SVG DOM) surfaced 47 confirmed correctness and performance defects. This PR fixes them. Every alignment-sensitive fix was verified by rendering real plots through the full pipeline and rank-matching the emitted `maidr-data` values against the DOM geometry (rect heights/widths, grayscale fills, `use` element positions).

### User-visible crashes fixed

- **Faceted boxplot / violin / histogram / smooth / heatmap / stacked / dodged plots crashed** with `unused arguments` — seven processors did not accept the facet-path arguments. They now accept them, filter data per panel, and scope selectors to the panel grob.
- **`curve(sin(x))` and `plot(y ~ x, subset = ...)` errored (or silently recorded wrong values)** because wrappers force-evaluated `...` before the original call. Arguments are now forced only after the original call succeeds, with an NSE fallback recording expressions plus the caller environment for faithful replay.
- **`lines(v)` / `plot(v)` crashed and `plot(x, type = "l")` produced `NA` data** from blind positional access (`args[[2]]` grabbing `"l"`). Arguments are now resolved the way the plotting functions match them.
- **Dodged `geom_bar(position = "dodge")` with `stat = "count"` always crashed**; tibble inputs crashed violins and emitted malformed dodged-bar values (1x1 tibbles instead of numbers).
- **`show(as_widget = TRUE)` always errored for Base R plots**; `render_maidr()` errored on NULL reactives; `hist(x, plot = FALSE)` was recorded as a plot call, injecting phantom layers into the next render.

### Silently wrong accessible output fixed

- **Base R barplot announcements/highlights were misaligned with the drawn bars** whenever categories weren't already alphabetical (the processor re-sorted data while the SVG kept call order). Data, drawing, and replay are now consistent; `barplot(horiz = TRUE)` emits `orientation = "horz"`.
- **Matrix `barplot()` without explicit `beside` was typed as simple "bar"** instead of stacked (barplot's default *is* stacked), and **`legend.text` swatch rectangles leaked into stacked/dodged highlight selectors**.
- **Base R heatmap cell highlighting was vertically mirrored** (gridSVG emits rects bottom-up; the frontend heuristic assumes top-down) — fixed with per-cell selector grids. `heatmap()` also ignored its dendrogram reordering, `image()` was transposed, and positional matrix args produced an empty layer.
- **Faceted line plots with Date/continuous x emitted NA for every point** (built positions were treated as category indices); **faceted bar panels were labeled from the whole dataset** instead of the panel's own scale.
- **Probability histograms announced counts instead of densities**; `right=`/`include.lowest=` were dropped from the recomputation.
- **Box plot outlier highlights (both pipelines) assumed lower outliers are drawn first**; they are drawn in data order, so interleaved outliers highlighted the wrong points.
- **Nested patchwork layouts (`(p1 | p2) / p3`) dropped panels** from the accessible output.
- **maidr-data JSON silently rounded all numerics to 4 decimal digits** (jsonlite default); **scatter point colour/group values were never emitted**; **multiline plots with 10+ series mapped series to the wrong polylines**.
- Base R multipanel: `layout()` grids fell into the single-panel branch, plots beyond the grid were silently dropped, empty cells serialized as invalid `{}`, per-panel `axis()` formats were applied globally, and `orientation`/`domMapping` fields were dropped.
- `$` partial matching crashed layout extraction on `plot(..., subset = ...)`; scale label mapping keyed labels by `seq_along` instead of break positions; facet_grid column variables were missing from subplot titles; format configs were mis-attached in facet output; histogram/smooth processors extracted the *first* similarly-shaped layer instead of their own; ggplot2 heatmaps mapped positions by data-appearance order instead of factor level order.

### Behavior / integration fixes

- **knitr**: `maidr_off()` now really disables RMarkdown interception; PDF/LaTeX output keeps its figures; repeated `maidr_on()` can no longer cause infinite hook recursion; non-HTML output uses the stored original hook; iframe height auto-resize now works in RMarkdown.
- **`chartSeries()` is now recorded when quantmod loads after maidr** (the packageEvent hook was a silent no-op against the sealed namespace).
- CDN URL pinned to the bundled MAIDR.js version instead of `@latest`; `maidr_set_fallback()` no longer resets unspecified settings; plot IDs no longer consume RNG state (preserving `set.seed()` reproducibility); iframe HTML is UTF-8-safe; unsupported data-bearing overlays (`polygon()` etc.) trigger the documented fallback; stale temp devices are cleaned up safely.

### Performance

- ggplot2 renders build the plot **once** instead of three times; faceted/patchwork paths reuse the build instead of rebuilding per layer; the boxplot processor cached 4 redundant builds; auto-print reuses the support-check orchestrator.
- Base R renders memoize the replayed gtable; candlestick SVG post-processing parses the document once instead of five times; multi-MB bundled assets are read once per session; per-plot network probes removed.

## Verification

- Full testthat suite: **2762 pass, 0 fail** (baseline 2704 before this branch; 2708 at first review; the rest are regression tests added during review).
- `R CMD check`: clean on all four CI platforms.
- Empirical DOM alignment suite: named/unnamed/horizontal/stacked/dodged barplots, `image()`, `heatmap()`, probability histograms, boxplot outliers, mfrow grids — emitted data verified rank-consistent with the exported SVG geometry, cross-checked against the bundled maidr.js selector contracts.
- Faceted plots for all seven previously-crashing layer types (facet_wrap and facet_grid, Date/numeric/discrete x) and nested/flat patchwork layouts verified end-to-end.

Known limitation documented in-code: faceted violins are skipped (rendered visually, not interactively) because their multi-layer result and KDE coordinate injection don't support facet panels yet. The same guard also catches patchwork leaves — see #52.


jooyoungseo:
Worked through the review. I verified each correctness claim against rendered output rather than by reading, since several of these are silent-wrong-data bugs where the code reads fine. Three were real and are fixed in `e537576`, `f9c76c0`, `940577a`; the rest are either already filed, already handled, or need a decision I shouldn't make alone.

## Fixed

**Faceted scatter falls back to raw hex — confirmed, and I had this wrong earlier.** I checked this during my own review and concluded it was a non-issue, but I compared the wrong key: I read `fill` on the faceted plot (absent) and `color` on the unfaceted one. Reading `color` on both shows the reviewer is right:

```
unfaceted  -> f | 4 | r        hex=FALSE
faceted    -> #00BA38 | #F8766D | #619CFF    hex=TRUE
```

`nrow(original_data) == nrow(layer_data)` is false for every faceted plot. Now compares against the *unfiltered* built rows and subsets the source column by the indices the panel kept. Both cases emit `f | 4 | r`.

**Faceted heat map pulls fill from the wrong panel — confirmed, and worse than described.** It isn't only the same-(x,y)-in-two-panels case; because the lookup takes `score_val[1]`, *every* panel reports panel 1's values. Two-panel tile plot, panel 2 should hold 100–400:

```
before   panel[1][1] points [[3,4],[1,2]]      panel[1][2] points [[3,4],[1,2]]
after    panel[1][1] points [[3,4],[1,2]]      panel[1][2] points [[300,400],[100,200]]
```

Values now come from the panel's rows; axis levels still come from the whole data set, since ggplot2 assigns discrete positions from the full scale.

**`base_r_state_tracking.R:112` counts empty layout cells — confirmed.** `layout(matrix(c(1,0,2,3), 2))` reported 4 panels. Fixed to match the `mat[mat > 0]` treatment already applied in `base_r_plot_grouping.R`.

Regression tests added for all three. Suite: **2762 pass, 0 fail** (was 2708 at the start of this review).

## Already filed as issues

- **Faceted violins skipped** → #52. I also found the guard catches *patchwork* leaves, which emit a zero-layer subplot. I tried narrowing the guard to facets only; the result is a degenerate `type: violin` layer with no data and no selectors, because the patchwork combiner can't expand the processor's `multi_layer` result either. Reverted rather than ship that — it needs the combiner work, which is the same root cause as the facet gap.
- **Dodged/stacked ordering** → #54, with your non-alphabetical fill-level case added as a second repro. Three orders diverge once levels aren't alphabetical: `extract_data` alphabetical, `reorder_layer_data` reverse-alphabetical, ggplot2 factor-level order. I deliberately did **not** change it: the emitted data is series-major while the DOM rects are x-major for *both* stacked and dodged, and that predates this PR. If the frontend flattened and zipped, every such chart shipped today would already be misaligned — far more likely it re-groups the nodes itself, in which case "fixing" the R side breaks all of them. Needs a read of the bundled maidr.js first.
- Overlay fallback scope → #55, auto-display → #53.

## Not acting on

**Duplicate side-effect evaluation on the retry path** — correct as described, and it's my code. The retry only runs after the direct call has already thrown, which previously meant a hard error (`plot(y ~ x, data = d, subset = g == 1)` failed outright). So the trade is "an argument with a side effect may evaluate more than once in a call that used to fail" against "the call works". I'll take that, but it deserves the note, so it's in the roxygen for `retry_call_in_caller_frame()`.

**NSE `call_env` captured by reference in a `for` loop** — plausible and worth a test, but note the retry changed the reachability: for `subset = grp == g` where `g` is a loop variable and `grp` a data column, `list(...)` now succeeds in the caller frame, so values get recorded and `call_env` stays `NULL`. I'd rather leave this than fix it speculatively.

**Version bump to `0.4.0.9000`** — skipping; that looks like release-tooling territory, maintainer's call.

**Shared helpers for the repeated grob-sort and panel-resolution idioms** — agreed, but that's a refactor across ~12 files with no behaviour change, and this PR is already 80 files. Better as its own PR.

## One correction to the review

The claim that `R/maidr_widget.R:15` documents bundled-by-default while the code probes the CDN is worth checking, but I didn't get to verifying it and don't want to assert either way.

CI is green on all four R CMD check platforms plus test-coverage — the latter two never ran on PRs before this branch added the `pull_request:` trigger, which is how the sweep reached review unverified in the first place.

---

jooyoungseo:
You were right about `maidr_widget.R:15`, and I was wrong twice getting there. Fixed in `1833a43`.

I tried to check it by running the code rather than reading it, and got a false negative — twice:

```
maidr_widget(p, use_cdn = NULL)  saved via saveWidget  -> no remote URL   # wrong
create_maidr_iframe("<svg/>", use_cdn = NULL)          -> no "http" at all # wrong
```

Both measurements were looking at the wrong bytes. The iframe payload is embedded as a **base64 `data:` URI**, so no `http` substring survives in the outer string, and `saveWidget` JSON-escapes it on top of that. Decoding the payload gives the answer directly:

```
iframe head:  <iframe ... src="data:text/html;base64,PCFET0NUWVBFIGh0bWw+...
decoded length: 2985
  https://cdn.jsdelivr.net/npm/maidr@3.73.0/dist/maidr.css
  https://cdn.jsdelivr.net/npm/maidr@3.73.0/dist/maidr.js
```

And the branch is exactly where you traced it:

```
create_standalone_html(NULL)   -> 2985 chars, jsdelivr CDN
create_standalone_html(TRUE)   -> 2985 chars, jsdelivr CDN
create_standalone_html(FALSE)  -> 3342594 chars, bundled inline
```

`maidr_internet_available()` is `TRUE` in this container, so `NULL` took the CDN branch — the documented "bundled" default does not hold for the widget path.

I took **option (a)**, correcting the doc rather than the behaviour. `show()`, `save_html()` and `create_maidr_html()` all document `NULL` as auto-detect; `maidr_html_dependencies()` is the lone hardcoded-bundled case and carries its own written rationale about a past squished-render bug. So the widget was the outlier in its docs, not in its behaviour, and flipping the behaviour would be a product call about RMarkdown payload size that isn't mine to make. The new text also names the divergence explicitly so the next reader doesn't have to re-trace it.

Suite still **2762 pass, 0 fail**.

On the two items you list as still open that I'm deliberately leaving:

- **Dodged-bar ordering asymmetry** — I did run your non-alphabetical case. Three orders diverge: `extract_data()` alphabetical, `reorder_layer_data()` reverse-alphabetical, ggplot2 factor-level order (`(a,zeta) (a,alpha) (b,zeta) (b,alpha)`). Full numbers are in #54. Still not changing it here: the same series-major-vs-x-major mismatch exists for **stacked** bars too and predates this PR, so if the frontend flattened and zipped, every stacked/dodged chart shipping today would already be misaligned. That points to the frontend re-grouping the nodes itself, in which case "fixing" the R side breaks all of them. It needs a read of the bundled maidr.js first, which is what #54 asks for.
- **NSE `call_env` by reference** — agreed it's real for `subset = grp == g` in a loop. Noting the retry narrowed its reach: where `g` resolves in the caller frame, `list(...)` now succeeds and values are recorded with `call_env` left `NULL`. It bites when the expression mixes a data column with a loop variable. I'd rather leave it than snapshot frames speculatively, since a shallow copy has its own failure modes.

Thanks for the persistence on the `use_cdn` one — I'd have shipped it wrong on the strength of a bad measurement.

---

PR #49, opened 2026-08-04
feat(step): support step plots for ggplot2 geom_step and base R type="s"

jooyoungseo:
## ✅ Merge order — 3 of 3 done

**Unblocked.** The vendored bundle is `maidr-4.0.0`, which carries the step trace, so a step layer no longer hits `Invalid trace type: step`.

- [x] **xability/maidr#723 merged** — `TraceType.STEP` is on `maidr` `main`.
- [x] **A release containing it is published** — `v4.0.0`, 2026-08-10 15:24 UTC. It arrived on the Monday `schedule: cron '0 15 * * 1'` run, as expected.
- [x] **`chore: update MAIDR bundle to v4.0.0`** (`56d6818`) landed on `main` and is merged here, replacing `inst/htmlwidgets/lib/maidr-3.75.1/`. It was produced by this repo's own `update-maidr-bundle.yml`, which verifies the download against the integrity hash published in the npm registry metadata — rather than by a hand-rolled commit inside this PR, which would have put a major-version bundle jump under a step-plot review.

**Verified against the vendored file, not the commit message.** `inst/htmlwidgets/lib/maidr-4.0.0/` exists, `TraceFactory` contains `case f.STEP:return new Br(e)`, the enum is `` STEP=`step` ``, and `stepDirection` appears throughout. Worth recording: the obvious check, `grep '"step"'`, returns **zero** — the bundle uses backticks — so that check alone would have read as a false negative and is not evidence of absence.

**And verified end to end in Chromium**, which the R test suite structurally cannot cover: the R tests prove the emitted JSON is right, not that the JS reads it. A factor-`y` hypnogram rendered against this bundle announces

```
Hour is 1, Sleep stage is Awake
Hour is 2, Sleep stage is REM
... (8 in total) ...
No more data to display
```

— **eight announcements for eight samples, not fifteen for the stairstep vertices**, and the *level names* rather than the numeric level codes. No page errors, and the highlight element resolves against the emitted selector.

`v4.0.0` is a major bump. Its breaking change is `NavigateCallback` receiving `null` when no data point is active; r-maidr never sets `onNavigate` outside the vendored bundle, so this package is not exposed to it.

## Merging `main` while waiting on that release

`main` moves several times a day and this branch has taken it every time, so every merge is documented in a thread comment. Most are mechanical — `NEWS.md` conflicts on nearly all of them and is a keep-both, folded under one `New Features` and one `Bug Fixes` heading rather than left as two of each. The six that needed a real decision are summarised below. (An earlier revision of this description counted the merges here; the count went stale faster than it could be edited, and the per-merge sections are the part worth reading anyway.)

### `1e02fa0` — #63's patchwork rewrite

#63 replaced the single-layer rebuild in `R/ggplot2_patchwork_utils.R` with one that expands a `multi_layer` result and then carries the processor's remaining fields generically — but **deliberately only for expanded results**:

```r
# Restricted to expanded results so single-layer patchwork payloads
# keep their current shape.
if (expanded) { for (field_name in names(sub)) { ... } }
```

A step layer is a **single-layer** result, so it never reaches that carry. Taking #63's version wholesale would have dropped `stepDirection` from every patchworked step plot — the exact regression the explicit carry exists to prevent, reintroduced invisibly. So I took #63's structure verbatim and re-added a narrow carry for the non-expanded path only:

```r
if (!expanded && !is.null(sub$stepDirection)) {
  layer_entry$stepDirection <- sub$stepDirection
}
```

### `fb836d9` — #87's facet rewrite, and the opposite answer

#87 added the same generic-carry mechanism to the **facet** path. Here the right answer was the reverse: #87's carry is *unconditional*, and `stepDirection` is not in its exclusion list, so it subsumes the explicit lift this branch had. I took main's version, and `3a9b2be` then removed the now-dead `first_step_direction()` helper, its `man/` page and its unit test.

I did not conclude that from reading the exclusion list. The end-to-end test passes on the merged tree, **and I confirmed it is not vacuous** — neutering #87's carry so it copies nothing makes it fail on the `stepDirection` expectation specifically.

`R/grob_utils.R` also conflicted in the fourth merge, and was not a keep-both either: main had **deleted** the free `find_panel_grob()` next to a function this branch added. Resolved as main's deletion plus our addition, after confirming main had removed every call site.

### `bcdbeea` — #103's selector rewrite, and a second deletion to take

#103 replaced `generate_selectors()` wholesale. `find_all_polyline_grobs()` is gone; the panel's polylines now come from `layer_polyline_grobs()`, which skips entire grob trees named after another layer's geom (`geom_smooth.gTree.N`) so a smooth's three curves stop shifting every position by three. This branch's copy of `find_all_polyline_grobs()` had no callers left after that, so I took the deletion — the same shape as the `R/grob_utils.R` resolution above.

What did **not** change is why this branch generalised the position counter. `GeomStep` draws through `GeomPath$draw_panel()`, which returns a bare `polylineGrob` carrying grid's auto-generated `GRID.polyline.N` name — no geom prefix — so a `geom_step()` lands in #103's candidate list exactly as a `geom_line()` does, and `other_geom_grob_prefixes()` skips only *named* trees, never a bare grob. Counting only `"line"` layers would still index the wrong polyline for both layers of a mixed plot. So `line_layer_position()` keeps delegating to `polyline_layer_position()`, and the mixed line+step test that pins it passes on the merged tree.

`DESCRIPTION` also conflicted there and lists pie and step plots. In `inst/examples/ggplot2_all_plot_types_example.R`, #104 wrapped the candlestick block in `if (have_tidyquant)`; the step test sits outside that guard, since step needs no tidyquant.

### `86ed549` — #113's positional Base R arguments, landing next to the step work

#113 matches every recorded Base R call against the definition R dispatched to before the processors read it. Three things followed that the textual merge could not produce on its own:

1. **A test of this branch asserted the old behaviour.** It expected `plot(x, y, type = "b")` to read as `"line"`. #113 deliberately moved `"b"` to the points branch — it draws its segments with a gap at every symbol, which gridSVG exports as `"brokenline"`, a name the line selector cannot address, so that layer used to come out with no highlight at all. The expectation is updated to `"point"`; the test still guards what it was written for, that the step branch claims only `"s"` / `"S"` and leaves every other non-`"p"` type where the branches around it put them.

2. **#113's own comment had stopped describing the code below it.** It listed `"c"`, `"h"`, `"s"` and `"S"` together as types with "nothing to point at until a trace type exists for them". A trace type now does exist for `"s"` / `"S"`, and `BaseRStepLayerProcessor` overrides the grob search to the `-step-` / `-Step-` names gridGraphics gives them, so those two are addressable rather than unpointable. Narrowed the sentence to `"c"` and `"h"`.

3. **`plot(x, y, "s")` is described for the first time.** The patched functions are declared `function(...)`, so a positional third argument never reached `args$type` and a positional stairstep read as points. With #113's matching sitting in front of detection it arrives as `type = "s"`. I verified this end-to-end through the orchestrator rather than inferring it from the NEWS entry — `type` `"step"`, `stepDirection` `"hv"`, six points for six samples, selector `#graphics-plot-1-step-1\.1 polyline` — and pinned it with a test, because reverting the matching sends the layer back to `"point"` silently, with no error anywhere.

### `2990b3f` — #116 fixing this branch's workaround out from under it

#116 tags every R6 method docblock in the package with `@description`, which is exactly what this branch had done for `line_layer_position()` alone — so the same lines conflicted. The resolution is neither keep-both nor take-theirs. I kept this branch's text, because it carries the step-specific reason the position counter is generalised, and deleted the paragraph explaining *why* the tag is there. That explanation was accurate while this branch was the only thing fixing the problem; leaving it now would document a repo-wide fix as a local workaround.

### `a670271` — #120 turning markdown on in roxygen

#120 adds `Roxygen: list(markdown = TRUE)` to `DESCRIPTION`, so every `.Rd` is regenerated with markdown interpreted rather than passed through — and `man/Ggplot2LineLayerProcessor.Rd` and `man/maidr-package.Rd` conflicted. Neither was hand-merged. When the *generator's own settings* change, regenerating is the only resolution that can be right: any hand-merge would produce a file roxygen would immediately disagree with.

The five `man/` pages this branch adds needed no source edits, because they were already written in the same markdown the rest of the package used — they simply come out with `\code{}` and `\emph{}` now instead of literal backticks and asterisks. Confirmed rather than assumed: no literal backtick survives in any of them.

## Description

A step plot holds its value across an interval and then jumps, rather than interpolating between samples the way a line does. The motivating case is a **hypnogram** — sleep stage against time, an ordinal factor on y that is held for a stretch of the night and then changes. Drawing it as a line tells a blind user the sleeper glided continuously through the intermediate stages, which is not what the data says.

Worth knowing what this replaces: `geom_step()` currently falls through `detect_layer_type()` to `"unknown"`, which is not an error path but a **silent-degradation** one — `has_unsupported_layers()` → `should_fallback()` renders the whole plot as a static, inaccessible PNG. So the tests here assert on the emitted JSON `type`, not merely that rendering succeeded.

## Changes Made

**Detection.** `GeomStep` → `"step"`, placed *before* the `GeomLine`/`GeomPath`/`GeomMA` branch and testing `class(...)[1]` — `GeomStep` inherits `GeomPath`, so an `inherits()`-style check would swallow step into `"line"`. Base R needed two sites: the high-level `plot` branch maps any non-`"p"` type to `"line"`, and the low-level `lines` branch never inspected `args$type` at all.

**Processors.** `Ggplot2StepLayerProcessor` and `BaseRStepLayerProcessor`, both inheriting their line counterpart so data extraction, x-label recovery and selector generation are reused rather than copied. They emit one point per **data sample**, never one per stairstep vertex — the JS `StepTrace` derives transitions and run lengths by comparing consecutive `y` values, so vertex-level data would double every level and misreport both. It maps the rendered polyline's surplus vertices back onto the samples itself, which I confirmed live: a 10-point series renders **19 / 19 / 20** vertices at `hv` / `vh` / `mid`, exactly the `2n-1` and `2n` shapes it handles.

**Direction.** `hv` / `vh` / `mid` from `layer$geom_params$direction` (a `draw_panel` formal, not `layer$params`), defaulting to `hv` because that is ggplot2's own documented default — known, not guessed. Base R maps `type="s"` → `hv` and `type="S"` → `vh`, and **omits** the field when `type` is absent or unrecognised rather than defaulting, matching the contract's "omit when genuinely unknown".

**Faceted and patchwork step plots keep their direction**, by two different mechanisms now: the facet path uses main's generic field carry, the patchwork path needs this branch's narrow explicit carry because its generic one is scoped to expanded results. See the merge sections above.

## Two bugs found by executing the code, not by reading it

1. **A factor `y` aborted serialisation entirely.** `ggplot_build()` returns `y` classed `mapped_discrete`, which `jsonlite` cannot encode — `No method asJSON S3 class: mapped_discrete`. That killed payload generation for the entire motivating use case. `y` is now coerced to a bare number, which is also what the wire contract wants since `y` drives sonification, braille and the min/max range; the level *name* rides alongside as `label`. Pinned by a `toJSON` assertion, and now confirmed in the browser — the hypnogram announces "Awake" / "REM" / "N1", not `5` / `4` / `3`. (A factor-`y` `geom_line()` has the same latent bug; left alone as out of scope.)

2. **Selector indexing collided on mixed plots.** `line_layer_position()` counted only layers typed `"line"`, while the panel search returned *every* polyline in it. A plot combining `geom_line()` with `geom_step()` resolved the step layer's position to `NULL`, so it fell through to the first-polyline path and collided with the line layer — both layers highlighting the same geometry. Now generalised via `polyline_layer_position()` to count every polyline-producing type. The panel search has since been replaced by #103's `layer_polyline_grobs()`, which narrows the population to polylines no geom-named grob tree claims — a bare `GeomStep` grob is still in it, so the generalisation is still what makes the mixed case resolve.

**Label recovery design note:** rather than duplicating the line processor's grouping and NA-drop logic to track row indices, the level-code → level-name lookup is keyed on the *built* `y` value and constructed from the full unfiltered columns. That is correct regardless of which rows survive NA-dropping, panel filtering or group splitting, and it lets `extract_data()` simply delegate to `super$extract_data()` — preserving the line processor's NA-filtering rationale by inheritance rather than by copy.

## Testing

`testthat::test_local()` after the latest merge — **0 failures, 0 errors, 1 skip** (pre-existing: `test-ggplot2-orchestrator.R:89`). The count has fallen from 7 to 3 to 1 across these merges as main's newer tests cover cases that previously skipped. CI's own R-CMD-check reports its testthat figure independently, so these numbers are not local-only.

The browser walk-through described at the top is the one check the suite cannot do: `testthat` proves the emitted JSON is right, not that the bundled JS reads it. Both now hold.

**Retracting an earlier claim about CI coverage.** This description previously said "this repo's CI has no `R CMD check` or testthat job". That was true for *pull requests* when written and is now false: R-CMD-check existed all along but fired only on push to `main`, and #48 added the `pull_request` trigger. Both the test suite and `R CMD check` now run on this PR across macOS, Windows and two Linux R versions.

**On lintr — correcting an earlier claim.** This description previously said lintr was "clean apart from three items". That was understated. `lintr::lint_package()` cannot run on this repo at all — it dies with `subscript out of bounds` inside `validate_lint_object`, which reproduces on untouched files and is pre-existing. Linting file by file instead, `R/ggplot2_step_layer_processor.R` reports **two** `indentation_linter` items, at the continuation lines of two multi-line `if` conditions. I have deliberately left them: that method is a direct structural mirror of the parent class's `get_original_x_column`, whose identical continuation lines are flagged the same way. Matching the method it parallels seemed better than silencing the linter on the copy and leaving the original.

`man/` is regenerated with roxygen2 pinned to **7.3.2** to match `RoxygenNote` — the default 8.0.0 rewrites all 206 `.Rd` files — and, since #120, with `Roxygen: list(markdown = TRUE)`. Pages that drift with zero source edits on `main` itself are reverted rather than carried, so `man/BaseRAdapter.Rd` and `man/dot-maidr_chartseries_ta_warned.Rd` are byte-identical to main's copies. The eight pages that do differ are all step work: two new processor classes, three new helper topics, the two line-processor pages the step classes hook into, and two lines in `man/maidr-package.Rd` adding step to the supported-plot lists. `NAMESPACE` is unchanged, as expected: processors carry `@keywords internal`.

**The roxygen problem this branch worked around is fixed upstream now.** Earlier revisions of this description tracked a bogus-`\keyword` count on `man/Ggplot2LineLayerProcessor.Rd` — 29, then 14 — and credited part of that drop to this branch tagging one method with `@description`. #116 has since done exactly that repo-wide, and the page now carries a single `\keyword{internal}`. There is no count left to track and no workaround left for this branch to own; what it still carries at that spot is the *text*, not the tag. Every `man/` page this branch adds already used `@description`, so #116 needed nothing from them.


jooyoungseo:
Three findings, three different outcomes — thanks, the second one was my error.

**1. `build_level_lookup()` masking `names`** — the runtime-error part of this is **not correct**, but the underlying complaint is fair and I've acted on it.

R resolves a name in *call position* as a function, looking past non-function bindings, and `names(lookup) <- ...` is a call to `` `names<-` `` — neither is shadowed by a local character vector. Verified directly:

```r
f <- function() {
  lookup <- c("a", "b"); names <- c("X", "Y")
  names(lookup) <- c("k1", "k2")
  lookup[!duplicated(names(lookup))]
}
f()   #  k1  k2
      # "a" "b"
```

So step-layer processing and JSON generation for factor/character `y` were never broken — that path is also covered by the existing `toJSON` regression test, which passes. Renamed to `level_names` anyway in ce0dd1e, with a comment saying why: it *reads* like a bug even though it isn't, and that costs a reviewer time every time.

**2. Facet/patchwork carrying `stepDirection`** — **correct, and my mistake.** The description was written from an earlier revision and never updated after `stepDirection` propagation was added to those paths. The behaviour is intentional and the description was stale, not the code. I've rewritten that section and flagged the correction inline.

**3. The `\keyword{}` explosion** — **correct, and worse than it looks: it predates this PR.** `man/Ggplot2LineLayerProcessor.Rd` already carried **29** bogus keywords on `main`.

The cause is that an untagged method title inside an R6 class block is folded into the class block, where `@keywords internal` splits the accumulated prose into one `\keyword{}` per word. Every untagged method title in the file feeds it — `Process the line layer with actual SVG structure` is where `\keyword{Process}`, `\keyword{SVG}`, `\keyword{actual}` come from. My longer title pushed it from 29 to 47.

Tagging the one method this PR touches with `@description` takes the file to **15**, below where it started. I stopped there rather than retagging the rest of the file: the remaining entries belong to methods this PR does not touch, and fixing them is a change about roxygen hygiene, not about step plots. Worth its own issue if you'd like one.

`testthat::test_local()` after all three: **3136 passing, 0 failures, 0 errors, 7 skips.**

---

jooyoungseo:
### Third merge of `main` — `366d55e`

`main` moved `0cd9a1d` → `74c4ebd` while this waited on the release, taking four PRs: #67 (violin box statistics / zero-width render), #68 (knitr device clearing), #69 (time-boxed internet probe) and #70 (startup message). The PR went `dirty`; this resolves it.

**Only `NEWS.md` conflicted**, and only as both-sides-added-bullets under `## Bug Fixes` — kept both, ours first. Nothing else needed a decision: none of the four touched a file this branch changes.

```
NEWS.md                                    |  33 +   <- the only conflict
R/ggplot2_violin_layer_processor.R         | 393 +-
R/html_dependencies.R                      |  30 +-
R/knitr_support.R                          |   6 +-
R/ggplot2_system_init.R, R/svg_utils.R     |   9 +-
man/ + tests/                              | 637 +-
```

`git diff --name-only HEAD~1 HEAD | grep -i step` returns nothing — the step sources came through the merge untouched.

**The thing worth re-checking here is the patchwork carry**, since #67 rewrote ~390 lines of `Ggplot2ViolinLayerProcessor` and touched `test-violin-patchwork.R`, and it was #63's rewrite of that same area that silently dropped `stepDirection` last time. Confirmed still in place at `R/ggplot2_patchwork_utils.R:631`:

```r
if (!expanded && !is.null(sub$stepDirection)) {
  layer_entry$stepDirection <- sub$stepDirection
}
```

Suite after the merge: **0 failures, 0 errors, 7 skips** (all pre-existing). The 18 items in the warnings section are the long-standing `aesthetics were dropped during statistical transformation` notes from the candlestick patchwork tests, unrelated to this branch.

**Still blocked on the same thing.** maidr's latest release is `v3.75.1` (2026-08-03), which predates #723, so the bundled JS would still throw `Invalid trace type: step`. Next scheduled release run is Monday 2026-08-10 ~15:00 UTC.

*(The "Two merges of `main`" heading in the description above is now one behind — this is the third.)*

---

jooyoungseo:
### Fourth merge of `main` — `a6d9b6c`

`main` moved `74c4ebd` → `335b033`, taking **eight** PRs: #71, #72, #73, #74, #76, #77, #78, #79. The PR went `dirty` again; this resolves it. Two conflicts, and the second one needed an actual decision rather than a "keep both".

**`NEWS.md`** — both-added-bullets under `## Bug Fixes`, as in all three previous merges. Kept both, ours first.

**`R/grob_utils.R`** — this one was not a both-sides-added conflict. #76 **deleted** the free `find_panel_grob()` from that file (replaced by a recursive gtable walk), and this branch had added `polyline_layer_position()` immediately below it, so git could not tell which lines the deletion targeted.

Resolved as *main's deletion **plus** our addition*, after checking that the deletion is genuinely complete rather than assuming it:

- every processor class carries its own `self$find_panel_grob` **method**, which is untouched — only the free function went;
- `main` has no remaining reference to the free function anywhere in `R/`, `tests/` or `man/` (`man/find_panel_grob.Rd` is gone too);
- this branch's one free-function call site, `ggplot2_heatmap_layer_processor.R:218`, was already removed by main's own version of that file in the auto-merge.

Keeping our side wholesale would have re-introduced a function with no callers and an orphaned man page; taking main's side wholesale would have deleted `polyline_layer_position()`, which the line processor still calls at `ggplot2_line_layer_processor.R:609` and which the mixed `geom_line()` + `geom_step()` selector fix depends on. Both are verified in place after the merge.

**The patchwork carry survived the auto-merge**, which was the thing to watch: #78 (`make layer ids unique across a composition`) and #76 both rewrote `R/ggplot2_patchwork_utils.R` — the same file whose rewrite in #63 silently dropped `stepDirection` last time. Confirmed still present at line 661:

```r
if (!expanded && !is.null(sub$stepDirection)) {
  layer_entry$stepDirection <- sub$stepDirection
}
```

`git diff --name-only HEAD~1 HEAD | grep -i step` returns nothing — the step sources came through untouched.

Suite after the merge: **0 failures, 0 errors, 3 skips**. Skips are down from 7 because main's new tests cover cases that previously skipped; nothing was silenced here.

**Still blocked on the same thing.** maidr's latest release is `v3.75.1` (2026-08-03), predating #723, so the bundled JS would still throw `Invalid trace type: step`. Next scheduled release run is Monday 2026-08-10 ~15:00 UTC.

*(The "Two merges of `main`" heading in the description is now two behind — this is the fourth.)*

---

jooyoungseo:
### Fifth merge of `main` — `fb836d9`, plus a cleanup it made possible (`3a9b2be`)

`main` moved `335b033` → `474d126`, taking #75, #84, #85, #86, #87. Conflicts were `NEWS.md` (both-added bullets, kept both) and **`R/ggplot2_facet_utils.R`**, which is the interesting one.

#### The facet conflict was #63's patchwork problem again, with the opposite answer

#87 added a **generic field carry** to the facet path — the same mechanism the patchwork path already had, copying every processor field not in an explicit exclusion list, so that `orientation`, `violinOptions` and `domMapping` stop being dropped. It landed exactly where this branch's explicit `stepDirection` lift sits.

When #63 did this to the *patchwork* path, taking main's version wholesale would have broken the feature, because its carry was scoped `if (expanded)` and a step layer is a single-layer result — so I re-added a narrow carry. **Here the answer is the reverse.** #87's carry is unconditional, and `stepDirection` is not in its exclusion list:

```r
if (field_name %in% c(
  "id", "type", "selectors", "data", "title", "axes",
  "labels", "multi_layer", "layers"
)) next
```

So main's mechanism subsumes the explicit lift, and I took main's version — step stops being a special case on that path.

I did not conclude that from reading the list. The end-to-end test `"A faceted step plot keeps its stepDirection"` passes on the merged tree, **and I confirmed it is not vacuous**: neutering #87's carry so it copies nothing makes that test fail on the `stepDirection` expectation specifically.

```
── Failure ('test-ggplot2-step-layer-processor.R:582'): A faceted step plot …
Expected `layer$stepDirection` to equal "vh".
```

#### The follow-up commit

That leaves `first_step_direction()` with no production caller — only its own unit test and its generated Rd. `3a9b2be` removes the function, `man/first_step_direction.Rd` and the test, rather than keeping a helper alive for the sake of testing itself. The behaviour stays pinned by the end-to-end test above; git history holds the helper if the exclusion list ever grows to include `stepDirection`.

#### Everything else checked

- **Patchwork carry** — survived the auto-merge, still at `R/ggplot2_patchwork_utils.R:675`. Both `stepDirection` paths therefore remain covered, one by main's mechanism and one by ours.
- **`polyline_layer_position`** — still defined in `R/grob_utils.R` and called from `R/ggplot2_line_layer_processor.R`, despite #86 reworking selector emission in that file.
- `git diff --name-only HEAD~1 HEAD | grep -i step` returns nothing for the merge commit — no step source was touched by it.

Suite after both commits: **0 failures, 0 errors, 3 skips**.

**Still blocked on the same thing.** maidr's latest release is `v3.75.1` (2026-08-03), predating #723. Next scheduled release run is Monday 2026-08-10 ~15:00 UTC.

*(The "Two merges of `main`" heading in the description is now three behind — this is the fifth.)*

---

jooyoungseo:
Correcting a claim in this PR's description now that it has been checked against merged `main`.

The description says, of the `mapped_discrete` serialisation failure:

> (A factor-`y` `geom_line()` has the same latent bug; left alone as out of scope.)

**That does not reproduce.** On `ea17360` a factor-`y` `geom_line()` serialises and renders end to end with no error:

```r
stages <- factor(c("Awake","REM","N1","N2","N3","N2"),
                 levels = c("N3","N2","N1","REM","Awake"), ordered = TRUE)
df <- data.frame(t = 1:6, stage = stages)
p <- ggplot(df, aes(t, stage, group = 1)) + geom_line()

b <- ggplot2::ggplot_build(p)
class(b$data[[1]]$y)          #> "mapped_discrete" "numeric"
jsonlite::toJSON(b$data[[1]]$y)  #> [5,4,3,2,1,2]      -- no error
save_html(p, tempfile(fileext = ".html"), use_cdn = FALSE)  #> succeeds
```

`mapped_discrete` *is* on the built column, but it also inherits `numeric`, so `jsonlite` encodes it without complaint. I hit that crash while building the step processor and generalised it to `geom_line()` without testing `geom_line()`. That was an unverified inference and the description should not have stated it as fact.

There **is** a real gap on that path, just a different and quieter one — no crash, a wrong announcement:

| | `geom_line()` | `geom_step()` |
|---|---|---|
| point keys | `x, y` | `x, y, label` |
| first point | `{"x":"1","y":5}` | `{"x":"1","y":5,"label":"Awake"}` |

A blind reader navigating a factor-`y` line chart hears **"5"** where the axis shows **"Awake"**. The level-name recovery this PR built is on `Ggplot2StepLayerProcessor` only; the line processor never gained it.

Filing that as its own issue with this reproduction, since it is a real accessibility defect rather than the crash the description claimed.

---

PR #69, opened 2026-08-07
fix(assets): time-box the cached internet probe

### LINKED ISSUE #58: Internet probe is cached for the whole session, so one transient result pins asset loading
Found while reviewing #48. The caching is deliberate and documented; the missing invalidation is the part worth revisiting.

`maidr_internet_available()` (`R/html_dependencies.R:101-110`) probes once and caches for the life of the process:

```r
cached <- .maidr_asset_cache$internet
if (!is.null(cached)) return(cached)
result <- tryCatch(curl::has_internet(), error = function(e) FALSE)
.maidr_asset_cache$internet <- isTRUE(result)
```

The rationale in the roxygen is sound — `curl::has_internet()` can block for seconds on an offline machine, and probing per plot is wasteful. But there is no invalidation, so the first probe decides for the rest of the session, in both directions:

- **Probe fails transiently** (flaky DHCP, VPN reconnecting, a proxy hiccup) → every subsequent render inlines the multi-MB bundle, even once the machine is online. Correct output, much larger documents.
- **Probe succeeds, then the machine goes offline** (laptop leaves wifi, on a train) → renders keep pointing at the CDN and the plots are dead in the browser. This is the direction that actually hurts, because the failure is silent and the user is offline precisely when they can't debug it.

This matters more now that #48 pinned the CDN to the bundled version and routed the widget/iframe path through `create_standalone_html()`, which resolves `use_cdn = NULL` via this function (see the `use_cdn` discussion on #48).

Options, roughly in increasing cost:

1. Time-box the cache (re-probe after N minutes) so a stale result self-heals.
2. Cache only the positive result and re-probe on failure, on the theory that "offline" is the state more likely to change during a session.
3. Leave it and document the behaviour on the user-facing `use_cdn` parameter, so anyone who cares can pass `use_cdn = FALSE` explicitly.

Related: `DESCRIPTION` still declares `curl` in Imports and this is its only use. #48's description claims the unused `curl` import was dropped, but the import is still there and still used — the claim is what is wrong, not the code.

jooyoungseo:
Closes #58.

`maidr_internet_available()` probed `curl::has_internet()` once and cached the answer for the life of the process, with no invalidation. Confirmed by mocking the probe with a call counter: two calls produced exactly **1** probe, and the cache entry had no expiry field at all.

So the first answer decided CDN-versus-inline for the whole session, in both directions:

- **Probe fails transiently** (flaky DHCP, VPN reconnecting, proxy hiccup) → every later render inlines the multi-MB bundle. Correct output, much larger documents.
- **Probe succeeds, then the machine goes offline** → renders keep emitting CDN references and the plots are dead in the browser. This is the direction that hurts, because it is silent and the user is offline precisely when they cannot debug it.

## The change

The result now expires after five minutes (`MAIDR_INTERNET_CACHE_TTL`), with the timestamp stored alongside the value.

Five minutes keeps the performance rationale the roxygen already gave — the probe stays per-*session*, not per-plot, and a knit of a many-plot document finishes well inside 300s, so it still costs one probe (occasionally two) per render. But a wrong answer can no longer outlive the connectivity it described by more than five minutes. Shorter (30–60s) would start probing repeatedly during a slow knit; an hour is barely different from "forever" for a typical session.

Two shapes also force a re-probe rather than trusting the cache: a missing timestamp (what the old implementation wrote, so an in-flight upgrade behaves), and a future-dated one (a backwards clock jump).

## Verification

`tests/testthat/test-html-dependencies.R` is new — 8 tests, 26 assertions, all mocked via `local_mocked_bindings(has_internet = ..., .package = "curl")`. **No test touches the network**, and each counts probes, so a silent real call shows up as a wrong count. Covers reuse inside the window with `probes == 0`, re-probe outside it, cold-probe timestamping, stale-success self-heal to FALSE, missing timestamp, future timestamp, and probe-throws-error.

Confirmed it is a real regression test by reverting the fix and re-running: **9 failures/errors appear**.

Full suite on this branch: **3469 passed, 0 failed, 0 errors, 35 skipped** (all pre-existing — missing `{tidyquant}`, Base R detection).

## Review notes

**On `MAIDR_INTERNET_CACHE_TTL` being SCREAMING_SNAKE_CASE** — it does get an `object_name_linter` note, but so does the constant 98 lines above it in the same file:

```
R/html_dependencies.R:4:1: style: [object_name_linter] ...
MAIDR_VERSION <- "3.75.1"
--
R/html_dependencies.R:102:1: style: [object_name_linter] ...
MAIDR_INTERNET_CACHE_TTL <- 300
```

That is the existing convention for top-level constants in this file and the note is accepted for it today, so renaming only the new one would make the file inconsistent with itself. Left as-is. (`lintr` is not run in CI here — the `Lint workflows and shell scripts` job is `shellcheck` + `actionlint` — and `lintr::lint_package()` is broken in this repo, so linting is done file by file.)

**On a persistently future-dated clock** — every call then falls into the `age < 0` branch and re-probes, permanently defeating the cache. That is fail-open: extra probes, never a wrong answer. Left unhandled deliberately; that is the direction I want it to fail.

**On the `curl` import** — the issue observes that #48's description claimed the unused `curl` import was dropped, but `curl` is still in `DESCRIPTION` Imports and still used: one live call site, `curl::has_internet()`. The claim was wrong, not the code. The import is untouched here.


jooyoungseo:
On the `MAIDR_INTERNET_CACHE_TTL` naming — checked, and it does get flagged, but so does the constant 98 lines above it in the same file. Ran `lintr::lint("R/html_dependencies.R")`:

```
R/html_dependencies.R:4:1: style: [object_name_linter] Variable and function name style should match snake_case or CamelCase.
MAIDR_VERSION <- "3.75.1"
--
R/html_dependencies.R:102:1: style: [object_name_linter] Variable and function name style should match snake_case or CamelCase.
MAIDR_INTERNET_CACHE_TTL <- 300
```

So SCREAMING_SNAKE_CASE for a top-level constant is the existing convention in this file, and the note is accepted for it today. Renaming only the new one would make the file inconsistent with itself; renaming both is a wider change than this PR's scope. Leaving as-is.

Worth adding that `lintr` is not run in CI here — the `Lint workflows and shell scripts` job is `shellcheck` + `actionlint`, and `lintr::lint_package()` is broken in this repo (pre-existing), so linting is done file by file by hand. The note is informational either way.

On the persistently-wrong-clock edge case: agreed, and agreed it is fail-open. A clock stuck in the future makes every call re-probe, which costs an extra `curl::has_internet()` per render but never returns a wrong answer. That is the direction I would want it to fail, so I have left it.

`devtools::test()` and lint were run locally — full suite 3469 passed, 0 failed, 0 errors, 35 skipped on this branch, and the new test file is lint-clean.

---

Issue #66, opened 2026-08-06
Patchwork layers miss axis number formatting and share layer ids across subplots


jooyoungseo:
Found while working on #52. Pre-existing on `main`, out of scope there.

## 1. `format_config` is never threaded through the patchwork path

`Ggplot2PlotOrchestrator$process_layers()` and `process_faceted_plot()` both do:

```r
private$.format_config <- extract_format_config(built)
```

and `combine_layer_results()` / the facet combiner then apply it via `attach_axis_format()` and run `validate_axes()`. `process_patchwork_plot()` does neither — it builds a minimal layout and calls `process_patchwork_plot_data()`, which has no `format_config` parameter at all.

So a `scale_y_continuous(labels = maidr::label_currency())`-style wrapper is honoured on a single plot and on a faceted plot, and silently ignored the moment the same plot is put in a patchwork: values are announced unformatted. Patchwork layers also skip the `validate_axes()` contract check the other two paths run.

## 2. Layer ids repeat across subplots

Every patchwork subplot numbers its layers from 1, so a 2x2 composition of single-layer plots emits `maidr-layer-1` four times. The ids are unique within a subplot but not within the figure, which means they collide in any figure-wide map the frontend keys by layer id — the number-formatter map in particular, which is the same feature (1) breaks.

(#52 added a `-N` suffix for layers expanded from one processor result, e.g. `maidr-layer-1-1` / `maidr-layer-1-2` for a violin. That makes them unique within their subplot; it does not address the cross-subplot collision.)

## Suggested direction

Thread `format_config` into `process_patchwork_plot_data()` → `process_patchwork_panel()` and apply the same `attach_axis_format()` + `validate_axes()` treatment the other two paths use, so all three paths share one contract. For the ids, either qualify them with the subplot's row/col or switch the frontend's figure-wide lookups to key on (subplot, layer) — worth confirming against the bundled maidr.js which of the two it expects before changing the R side.

PR #72, opened 2026-08-08
fix(patchwork): apply each leaf's axis number format

jooyoungseo:
Addresses the first half of #66. **Does not close it** — see below.

`Ggplot2PlotOrchestrator$process_layers()` and `process_faceted_plot()` both extract a format config and then apply it through `attach_axis_format()`, running the `validate_axes()` contract check on the way. `process_patchwork_plot()` did neither — it built a minimal layout and called `process_patchwork_plot_data()`, which had no `format_config` parameter at all.

So a `scale_y_continuous(labels = maidr::label_currency())`-style wrapper was honoured on a single plot and on a faceted plot, and silently ignored the moment the same plot went into a composition. Values were announced unformatted — currency read as a bare number, percentages without their sign. Patchwork layers also skipped the `validate_axes()` check the other two paths run.

Now threaded through `process_patchwork_plot_data()` → `process_patchwork_panel()` with the same `attach_axis_format()` + `validate_axes()` treatment, so all three paths share one contract. Extracted **per leaf** from that leaf's own build, since each leaf is its own ggplot with its own scales — a composition where one leaf formats currency and another formats percent keeps both.

## What this does NOT do

Part 2 of the issue — **layer ids repeating across subplots** — is untouched. Every patchwork subplot still numbers its layers from 1, so a 2×2 of single-layer plots emits `maidr-layer-1` four times.

That half needs a decision this PR does not make. The issue itself says to confirm against the bundled maidr.js whether the frontend expects ids qualified by subplot, or expects figure-wide lookups keyed on (subplot, layer), *before* changing the R side — because getting it backwards would break the frontend rather than fix it. I would rather land the verified half than guess at the other.

#66 stays open for it.

## Verification

`tests/testthat/test-patchwork-utils.R` gains coverage for the threading. Confirmed load-bearing rather than decorative by reverting `R/ggplot2_patchwork_utils.R` to `main` and re-running: **4 failures appear**.

Full suite on this branch: **3496 passed, 0 failed, 0 errors, 35 skipped** (all pre-existing — missing `{tidyquant}`, Base R detection).

---


jooyoungseo:
Confirming CI, since both review rounds asked for it: **all 11 checks are green** on `5451423`.

| check | result |
|---|---|
| ubuntu-latest (release) | success |
| ubuntu-latest (oldrel-1) | success |
| windows-latest (release) | success |
| macos-15-intel (release) | success |
| test-coverage | success |
| pkgdown | success |
| claude-review | success |
| Lint workflows and shell scripts | success |
| CodeQL / Analyze (actions) / Analyze (javascript-typescript) | success |

Worth recording why it was red before, because the review round guessed at a cause and the log disagreed. It was not the tests and not a flake — `R CMD check` runs with `error-on: "warning"`, and the failing step was:

```
  '::' or ':::' import not declared from: 'scales'
Status: 1 WARNING
Error: R CMD check found WARNINGs
```

The test suite itself was passing on all four platforms the whole time (`[ FAIL 0 | WARN 18 | SKIP 7 | PASS 3680 ]`, identical warning count to the green PR #71). The new axis-format tests call `scales::label_dollar()`/`scales::label_percent()`, and `scales` reaches this package only transitively through ggplot2 — enough to run, not enough to declare. `0b926d3` adds it to `Suggests`; both Ubuntu jobs that were failing went green on that commit alone.

Two notes on the earlier review:

- The nit on the fourth test's title was right, and it was right for a reason worth keeping. That test does not exercise number formats at all — what it guards is the `validate_axes()` call this PR newly put on the patchwork path. A gate that throws has one failure mode, aborting a composition that used to render, so it now says so: `"the axes gate passes every leaf of an unformatted composition"`, with a comment explaining why it deliberately uses plain scales.
- Re-confirming the scope: this is half of #66. The layer-id half (ids repeating across subplots) is deliberately not here — it needs the bundled `maidr.js` inspected first to determine whether the frontend wants ids qualified by subplot or figure-wide lookups keyed on `(subplot, layer)`, and that is not something to guess at. The PR body and NEWS entry are both scoped accordingly, and this PR does not close the issue.

---

PR #76, opened 2026-08-08
fix(patchwork): resolve nested leaf panels via the recursive gtable walk

### LINKED ISSUE #64: Non-violin geoms in a nested patchwork emit selectors that resolve to nothing
Found while working on #52. Pre-existing on `main`; not caused by that work and not fixed by it.

## What happens

In a nested patchwork, a leaf's layer is emitted with a selector that addresses no element in the exported SVG, so the layer is announced but nothing highlights.

```r
b1 <- ggplot(mpg, aes(class)) + geom_bar()
b2 <- ggplot(mpg, aes(drv)) + geom_bar()
s  <- ggplot(mpg, aes(displ, hwy)) + geom_point()

maidr::save_html((b1 | b2) / s, "out.html")
#  subplot[1][1] bar    selectors=1  resolving=0
#  subplot[1][2] bar    selectors=1  resolving=0
#  subplot[2][1] point  selectors=1  resolving=1
```

A flat `b1 | b2` is fine. It is nesting that breaks it, and a 2x2 `(b1|b2)/(s|b2)` breaks every cell.

## Why

Each processor carries its own `find_panel_grob()` that scans only the **top level** of the gtable:

```r
idx <- which(grepl(paste0("^", panel_ctx$panel_name, "\\b"), gt$layout$name))
```

A nested patchwork row keeps its panels inside a child gtable (`patchwork-table-N`); the top level holds only a `panel-nested-patchwork-N` placeholder. So the lookup either finds nothing or resolves to a different leaf's panel. Panel names are also not unique across nesting levels — `(v|b)/c` yields `panel-1, panel-2, panel-2`, and both halves of a 2x2 contain a `panel-1` — so matching by name is ambiguous by construction.

Affected: `R/ggplot2_candlestick_layer_processor.R:328` and the private `find_panel_grob` methods in the boxplot, point, and heatmap processors, plus `R/grob_utils.R::find_panel_grob()`.

## Suggested fix

#52 added `find_gtable_panel_grob(gt, panel_ctx)` in `R/ggplot2_patchwork_utils.R`, which recurses into child gtables and addresses panels by discovery position rather than by name. The violin processor delegates to it. Migrating the other processors onto the same helper would fix this class of bug in one place.

It was left out of #52 deliberately: it changes which panel those geoms resolve to under nesting, so the checked-in `inst/output/*.html` and any candlestick+volume examples need re-baselining, and that deserves its own review rather than riding along with a violin fix.

jooyoungseo:
Closes #64.

## Reproduced, and it is worse than the issue describes

Rendering `(b1 | b2) / s` and resolving each emitted selector against the exported SVG reproduces the report exactly. But the failure is not uniform — the ten affected geoms split into **two** modes, and the issue only names one of them:

```
bar     selectors=0                 point   selectors=0
box     selectors=0                 hist    selectors=1  resolving=0
                                    smooth  selectors=1  resolving=0
                                    stack   selectors=1  resolving=0
                                    dodge   selectors=1  resolving=0
```

Bar, point, box, line, heatmap and candlestick emit **no selector**. Histogram, smooth, stacked bar and dodged bar instead fall through to a **fabricated** selector (`#geom_rect\.rect\.1\.1 rect`) that matches nothing. The second is the more dangerous of the two — it is literally the issue's title, and the payload looks healthy while the layer highlights nothing.

A flat `b1 | b2` was never affected. It is nesting that breaks it, and a 2×2 `(b1|b2)/(s|b2)` breaks every cell.

## What changed

`find_gtable_panel_grob()` in `R/ggplot2_patchwork_utils.R` already recurses into child gtables and addresses panels by discovery position rather than by name — the violin processor delegates to it. All ten remaining processors now route through the same helper: bar, point, boxplot, heatmap, candlestick, line, histogram, smooth, stacked bar, dodged bar. The private `find_panel_grob` methods in boxplot, point and candlestick became one-line delegations, matching the shape the violin processor already had.

`R/grob_utils.R::find_panel_grob()` — the last item the issue names — lost its only caller (heatmap) and is removed along with its tests. Leaving a nesting-blind, name-matching panel finder in the codebase is how this bug comes back.

After the fix, every cell of every nested layout resolves to **its own** panel: bar-a → 7 marks (7 `class` levels), bar-b → 3 (3 `drv` levels), point → 234 (`nrow(mpg)`), each matching that layer's own data length.

## Verification

I verified this independently of the agent that drafted it, rather than taking the report at face value.

- **Bite check.** Reverting `R/` to `main` and re-running `test-nested-patchwork-selectors.R` gives `[ FAIL 20 | WARN 0 | SKIP 0 | PASS 32 ]`, with failures in all five nested blocks. With the fix: `[ FAIL 0 | PASS 87 ]`.
- **Full suite:** `[ FAIL 0 | WARN 2 | SKIP 31 | PASS 3689 ]` after merging current `main` (which now carries #71, #74 and #77).
- **The removal is safe.** No bare (non-`self$`) call to `find_panel_grob` survives anywhere in `R/` or `tests/`; it was never in `NAMESPACE`; the remaining matches are all R6 methods of the same name on individual processors.
- **`roxygen2::roxygenise()` is in sync** — after regenerating, the only drift is the three perennially-stale unrelated pages, which are untouched here.
- **No undeclared test dependency.** The new test file uses only `ggplot2`, `jsonlite`, `xml2` and `testthat`, all declared. No `withr`.

The flat-patchwork guard test deliberately still passes with the fix reverted: flat was never broken, and that test exists so a future change cannot "fix" nesting by breaking flat.

## `inst/output/*.html` — no re-baselining needed

The issue warned this might be required. It is not. Both patchwork baselines are **flat**, not nested — `example_patchwork_2x2_ggplot2.html` is `p_line + p_bar1 + p_bar2 + p_line_extra + plot_layout(ncol = 2)`, and `example_candlestick_ma_volume_ggplot2.html` is `p_price / p_volume`. Re-rendering the patchwork example with and without the fix produces **byte-identical** normalized `maidr-data`.

## Not verified locally — worth a CI glance

**Candlestick.** `tidyquant` is not installed in this environment, so all 19 candlestick tests skip. The processor's `find_panel_grob` now delegates like the others, and its old name-match and the new index lookup agree on a flat composition (both resolve to the k-th panel in gtable layout order, and names are unique when un-nested) — which is what the candlestick+volume example is. That equivalence was demonstrated empirically for the flat bar/line/point case but not for candlestick itself. CI has `tidyquant`, so it covers this.

## Follow-up split out

Four processors (histogram, smooth, stacked bar, dodged bar) still contain the fabricated-selector fallback itself; this PR fixes the common trigger, not the fallback. That is now tracked in #83, with the per-geom measurements showing it is **not** a uniform `return(list())` fix — the guess is never right for histogram/stacked/dodged, but for smooth it genuinely equals the real id in a simple plot.

## Retracted

~~An earlier version of this description claimed six Base R baselines carry pre-existing unresolved selectors.~~ **That claim was wrong and is withdrawn.** It came from an audit that resolved selectors by stripping a leading `#` and cutting at the first space, which cannot parse the `g#…\.1 > polyline:nth-child(1)` and `polygon[id^='…']` forms Base R actually emits. Re-measured by loading each file in real Chromium and calling `document.querySelectorAll()`: **one** unresolved selector in **one** baseline (`example_boxplot_horizontal_base_r`), and re-rendering those figures with current code gives **zero** unresolved everywhere. There is nothing to fix and no issue was filed. Details in the comment thread below.


jooyoungseo:
On the fabrication-fallback finding — you're right that it's still reachable in those four, and I went and measured it rather than reasoning about it. The result is more interesting than either of us assumed, and it changed my mind twice.

**First measurement (wrong).** I rendered all four geoms in one R session and checked whether each fabricated id resolved:

```
histogram real=#geom_rect\.rect\.2\.1 rect    fabrication=geom_rect.rect.1.1     resolves=0
stacked   real=#geom_rect\.rect\.55\.1 rect   fabrication=geom_rect.rect.1.1     resolves=0
dodged    real=#geom_rect\.rect\.136\.1 rect  fabrication=geom_rect.rect.1.1     resolves=0
smooth    real=#GRID\.polyline\.218\.1\.1     fabrication=GRID.polyline.1.1.1    resolves=0
```

I read that as "the fallback never helps, so removing it is free" and ripped all four out. That reading was wrong. **gridSVG numbers ids sequentially across the session**, so `.55.1`, `.136.1` and `.218.1.1` are large only because three other plots rendered first. The measurement was confounded by its own setup.

**Second measurement (fresh R process per geom, which is what a user actually gets):**

```
histogram real=#geom_rect\.rect\.2\.1 rect  guess=geom_rect.rect.1.1    guess_exists=0
stacked   real=#geom_rect\.rect\.2\.1 rect  guess=geom_rect.rect.1.1    guess_exists=0
dodged    real=#geom_rect\.rect\.2\.1 rect  guess=geom_rect.rect.1.1    guess_exists=0
smooth    real=#GRID\.polyline\.1\.1\.1     guess=GRID.polyline.1.1.1   guess_exists=1
```

So the four are **not** the same case:

- For histogram, stacked bar and dodged bar the guess is never right — the first rect grob is `.2.1`, not `.1.1`, even in the simplest possible plot. There the fallback is pure harm: it converts a detectable empty selector list into an undetectable wrong one.
- For **smooth** the guess is *identical to the real id* in a simple single-layer plot. Removing it there would turn a selector that happens to work into an empty one. That is a real regression risk, not a cleanup.

Given that, I've **reverted the removal out of this PR** rather than shipping three-quarters of it. Two reasons: this PR is scoped to nested-patchwork panel resolution and the nested case is fixed for all ten regardless of the fallback; and a change that's clearly right for three geoms and clearly risky for the fourth deserves to be reviewed as its own diff, not folded into a fix for a different bug. Your "worth a follow-up" framing was the right call.

Worth carrying into that follow-up: the fix is probably not a uniform `return(list())`. Smooth needs its lookup to stop failing, or a fallback that is *checked* against the document rather than assumed — the other three can just return empty.

On the style note: agreed that `test-nested-patchwork-selectors.R` should qualify with `testthat::` like the rest of the suite. Leaving it for now since the file is otherwise final and CI is green on it, but noting it here so it isn't lost.

The candlestick point is the one I'd still like CI to answer, for the reason you give — `tidyquant` is absent locally so all 19 candlestick tests skip here.

---

jooyoungseo:
**Correction to this PR's description.** The "Out of scope, pre-existing" section claims six Base R baselines carry unresolved selectors. **That claim is wrong and should be disregarded** — it was a measurement artifact, and I am striking it rather than leaving it to mislead whoever picks it up.

The audit behind it resolved selectors by stripping a leading `#` and taking everything up to the first space. Base R selectors are not that shape:

```
g#graphics-plot-1-segments-3\.1 > polyline:nth-child(1)
polygon[id^='graphics-plot-1-polygon-1.1']
```

Child combinators, `:nth-child()`, attribute prefix matches and the `g#` type-plus-id form all fail that parser, so it reported "unresolved" for selectors that are in fact fine.

Re-measured by loading each file in real Chromium and calling `document.querySelectorAll(sel)` — the same thing the frontend does:

| baseline | selectors | unresolved |
|---|---|---|
| `example_boxplot_horizontal_base_r` | 14 | **1** |
| `example_boxplot_vertical_base_r` | 14 | 0 |
| `example_density_plot_base_r` | 1 | 0 |
| `example_dodged_bar_plot_base_r` | 1 | 0 |
| `example_histogram_base_r` | 1 | 0 |
| `example_histogram_density_base_r` | 2 | 0 |
| `example_multipanel_2x2_mfrow_base_r` | 4 | 0 |

So it is one selector in one baseline, not 13 across six. And re-rendering those same figures with current code gives **0 unresolved everywhere**, including the horizontal boxplot — that baseline predates changes since (14 selectors then, 10 now), so even the single hit is a stale-file artifact rather than a live defect.

Nothing in this PR changes as a result; the correction is only to the "pre-existing problems" note in the description. No separate issue will be filed for it, because there is nothing to fix.

---

Issue #80, opened 2026-08-08
Dodged bar with stat = "count" emits ragged series, desyncing the highlight mapping


jooyoungseo:
Found while working on #27/#28. Reproduced against current `main`.

## What happens

A dodged bar chart whose fill/x combinations are not all populated emits **series of different lengths**, because zero-count cells are dropped rather than emitted as zeros.

```r
library(ggplot2); library(maidr)
maidr::save_html(ggplot(mpg, aes(class, fill = drv)) + geom_bar(position = "dodge"), "out.html")
```

Emitted payload:

```
type: dodged_bar   n series: 3
series 1  z=4  n=5  ys = 12,3,33,4,51
series 2  z=f  n=4  ys = 35,38,11,22
series 3  z=r  n=3  ys = 5,9,11
```

Against the truth:

```
    2seater compact midsize minivan pickup subcompact suv
4         0      12       3       0     33          4  51
f         0      35      38      11      0         22   0
r         5       0       0       0      0          9  11
```

Each series' **values** are right — they are exactly that row's non-zero entries. The problem is the shape: 5, 4 and 3 elements, and the columns no longer line up across series. Series 1 covers `compact, midsize, pickup, subcompact, suv`; series 2 covers `compact, midsize, minivan, subcompact`. Position 3 means `pickup` in one series and `minivan` in the next.

## Why it matters

The bundled `maidr.js` regroups the flat node list column-major and assumes a rectangular `barValues`:

```js
for (let col = 0, k = 0; col < barValues[0].length; col++)
  if (domMapping?.groupDirection === "forward")
    for (let s = 0; s < barValues.length; s++)      out[s][col] = nodes[k++];
  else
    for (let s = barValues.length - 1; s >= 0; s--) out[s][col] = nodes[k++];
```

It walks `barValues[0].length × barValues.length` = **5 × 3 = 15** node slots. Only **12** rects exist (5 + 4 + 3). So `k` runs past the end and every assignment after the first ragged column is shifted — the highlight lands on the wrong bar, and the tail gets `undefined`.

This is the same class of defect as #54, except #54 concluded the *rectangular* case is correct by construction. It is the raggedness that breaks it, so the contract documented in that work still holds — it just is not being met here.

## Notes for whoever picks this up

- `geom_col()` with an explicitly complete data frame is unaffected; the trigger is a `stat = "count"` cross-tabulation with structurally empty cells.
[...]

Issue #83, opened 2026-08-08
Four processors fabricate a selector when the grob lookup fails, instead of emitting none


jooyoungseo:
Split out of #76, where it was raised in review and deliberately deferred so it could be reviewed as its own diff.

## The pattern

When the panel or rect grob lookup comes up empty, four processors invent an id rather than returning `list()`:

- `R/ggplot2_histogram_layer_processor.R` — hardcoded `"geom_rect.rect.1.1"`
- `R/ggplot2_smooth_layer_processor.R` — hardcoded `"#GRID\\.polyline\\.1\\.1\\.1"` (two call sites)
- `R/ggplot2_stacked_bar_layer_processor.R` — `paste0("geom_rect.rect.", get_layer_index(), ".1")`
- `R/ggplot2_dodged_bar_layer_processor.R` — same, when `panel_grob` is `NULL`

Bar, point, boxplot, line, heatmap and candlestick all return `list()` in the same situation. `find_gtable_panel_grob()`'s own doc states the principle: *"the caller can tell an empty selector list apart from a wrong one, a user cannot."* A fabricated selector that matches nothing is the worse failure — the payload looks healthy while the layer highlights nothing, which is exactly the shape of #64.

## It is not a uniform fix — measured per geom

The obvious change is "replace all four with `return(list())`". That is right for three of them and **wrong for the fourth**. Measured in a **fresh R process** per geom (this matters: gridSVG numbers ids sequentially across the session, so measuring several plots in one process inflates the ids and makes every guess look wrong):

| geom | real id | fabrication | fabrication exists? |
|---|---|---|---|
| histogram | `geom_rect.rect.2.1` | `geom_rect.rect.1.1` | no |
| stacked bar | `geom_rect.rect.2.1` | `geom_rect.rect.1.1` | no |
| dodged bar | `geom_rect.rect.2.1` | `geom_rect.rect.1.1` | no |
| **smooth** | `GRID.polyline.1.1.1` | `GRID.polyline.1.1.1` | **yes — identical** |

For histogram, stacked and dodged the guess is never right even in the simplest single-layer plot, so it is pure harm and can be replaced with `list()`. For smooth the guess *is* the real id in a simple plot, so removing it would turn a selector that happens to work into an empty one. Smooth needs either a lookup that stops failing, or a fallback that is **validated against the document** before being emitted, rather than assumed.

## Reachability

[...]

PR #84, opened 2026-08-08
fix(ggplot2): split a grouped smooth into one series per curve

### LINKED ISSUE #81: geom_smooth(aes(colour = g)) flattens every group into one series
Found while working on #27/#28. Reproduced against current `main`.

## What happens

A grouped smooth draws one curve per group but is described as a single undifferentiated series.

```r
library(ggplot2); library(maidr)
set.seed(2)
df <- data.frame(x = rep(1:10, 3), y = rnorm(30), g = rep(c("a", "b", "c"), each = 10))
maidr::save_html(
  ggplot(df, aes(x, y, colour = g)) + geom_smooth(se = FALSE, method = "loess", formula = y ~ x),
  "out.html"
)
```

Emitted payload:

```
type: smooth
n series: 1
data[[1]]: 240 points
distinct z values: (none)
```

ggplot2 draws **three** curves at 80 points each. The payload concatenates all 240 points into one series and carries no `z` at all, so the three groups are indistinguishable — a reader navigating the layer walks from the end of one curve straight into the start of the next with nothing announced in between.

## Comparison

The line processor handles the equivalent case correctly: `geom_line(aes(colour = g))` emits one series per group with a per-point `z`, and (after #75) a `z` label taken from the legend title. The smooth processor does not do the group split at all.

## Suggested direction

`Ggplot2SmoothLayerProcessor` needs the same group handling `Ggplot2LineLayerProcessor` has — split `built_data` by `group`, emit one series per group with `z` set to the group's value, and resolve the legend title for `axes$z$label`. `resolve_legend_label()` in `R/axes_utils.R` (added in #75) is the shared helper for that last part, so this should be able to reuse rather than duplicate it.

Note #75 is open and adds that helper; this issue is not fixed by it.

jooyoungseo:
Closes #81.

## Before / after

The issue's reproduction, `ggplot(df, aes(x, y, colour = g)) + geom_smooth(se = FALSE, method = "loess")` — measured one plot per fresh R process, selectors resolved with Playwright/Chromium `document.querySelectorAll`:

| | before | after |
|---|---|---|
| series | 1 | 3 |
| points | `[240]` | `[80, 80, 80]` |
| z values | none | `a`, `b`, `c` |
| `axes$z$label` | absent | `"g"` (or the `labs(colour = )` title) |
| selectors | 1 → an **80-vertex** polyline against 240 data points | 3 → one per group, each resolving to exactly 1 polyline of exactly 80 vertices |

## The selectors were already wrong, not just the data

The old single selector pointed at one group's polyline while the layer carried all three groups' points — 240 data points against 80 vertices, highlighting only group 3. That is the same class of mismatch as #64 and #80, so this fixes both halves.

`GeomSmooth` has a `draw_group`, so the layer's grob tree holds an equal run of children **per group**: `[polyline] × 3` for `se = FALSE`, `[ribbon gTree, polyline] × 3` for `se = TRUE`, `[ribbon gTree] × 3` for `geom_density()`. The fix chunks the layer's grob children by group and takes the last polyline in each chunk — the same "the curve is the last polyline drawn" rule the single-curve path already used, applied once per group instead of once per layer. It reduces exactly to the old behaviour when there is one group, and falls back to the old path whenever the child count is not a multiple of the group count.

Two details the chunking had to get right, both confirmed empirically rather than assumed: with `se = TRUE` the ribbon's outline polyline is `stroke="none"` while the fitted line carries the group colour, so picking the ribbon would have been *silently* wrong; and for `geom_density()` the curve **is** the ribbon's inner polyline, so a naive "skip ribbons" rule would have broken density. Selector search is scoped to the layer's own `geom_*` grob tree, which keeps a sibling `geom_line()` polyline and a second `geom_smooth()` out of the chunking.

## Shared with the line processor rather than duplicated

`resolve_legend_label()` is reused unchanged. The line processor's four group helpers moved into a new `R/series_group_utils.R` — `resolve_series_group_mapping()`, `data_has_series_groups()`, `resolve_series_group_names()`, `attach_series_group_axis()` — and `Ggplot2LineLayerProcessor`'s methods became pure delegations, so both processors share one implementation.

One deliberate deviation: smooth probes `fill` after `colour`, because `geom_density(aes(fill = g))` is an ordinary way to split these geoms while a line has no fill.

## Verification

Independently re-verified rather than taken on trust.

- **Bite check**: reverting `R/` to `main` and deleting the new file, keeping the tests, gives `[ FAIL 20 | WARN 0 | SKIP 0 | PASS 472 ]` in `test-ggplot2-smooth-layer-processor.R`. The four ungrouped tests correctly still pass — they are no-regression guards for the single-curve path, not bug detectors.
- **Full suite**: `[ FAIL 0 | WARN 2 | SKIP 31 | PASS 3929 ]`, against a `main` baseline of 3879. Same 2 pre-existing warnings, same 31 skips — so the line processor's existing tests survive the refactor.
- Also verified after the fix: `se = TRUE` grouped, `geom_point + smooth`, `geom_line + smooth`, two grouped smooths in one plot, layer-level `aes(colour = )`, `labs(colour = )`, `aes(group = )` (emits `Series 1..3` with no z label, matching the line processor), `facet_wrap`, and patchwork.

## One claim in the original write-up was wrong, and is corrected here

The first draft of the NEWS entry and the `group_aes()` comment both said `geom_area()` splits per group the same way. **It does not.** `Ggplot2Adapter` types a layer as `smooth` only for `GeomSmooth` or `StatDensity`, and `geom_area()` is neither — rendering one emits **no `maidr-data` payload at all**, so it never reaches this processor. Both places are corrected in `187b699`, and the comment now states which two geoms actually arrive here.

`geom_density()` is genuinely covered and stays: verified for both `aes(colour = g)` and `aes(fill = g)`, 3 series and 3 selectors each, `z` label `g`.

## Noted, not fixed here

With `geom_line(aes(colour = g)) + geom_smooth(...)`, the **line** layer emits 3 series but only 1 selector. Confirmed against `origin/main`, so it predates this change — a separate line-processor defect worth its own issue.

`lintr::lint_package(".")` crashes with an internal lintr error (`subscript out of bounds` in `validate_lint_object`); pre-existing and unrelated, so files were linted individually and come back clean.

---


jooyoungseo:
Both points were right, and the second one caught me correcting an over-claim with another over-claim. Fixed in `e597d01`.

**The fallback now emits nothing instead of one wrong selector.** You're right that falling through to the single-curve path reintroduces exactly the defect this PR fixes — one selector against `n_series` series. I'd framed it as graceful degradation; it isn't, it's the wrong-selector failure with a different trigger. It now returns `list()`, which matches the principle `find_gtable_panel_grob()` already documents and that #83 is applying to four other processors: *the caller can tell an empty selector list apart from a wrong one, a user cannot.* Collapsing the data back to one merged series would also be internally consistent, but it throws away the per-group `z` values that are the actual accessibility win here, so empty selectors is the better trade.

Pinned by a test that forces the give-up (3 curves, a claimed group count of 4), so a future ggplot2 rendering change that trips the branch fails loudly.

**`geom_area(stat = "density")` does reach this processor** — I checked rather than took it on reading, and it splits correctly:

```
type: smooth   series: 3   selectors: 3   zlabel: g
```

So my earlier correction was too absolute. The comment now says what is actually true: a *default* `geom_area()` uses `StatAlign` and never arrives here, but `geom_area(stat = "density")` is typed `smooth` and does. There's a test for that too now.

For the record, the sequence on this one claim was: the original NEWS said `geom_area()` splits (wrong — a default one emits no payload at all); I corrected it to "never reaches this processor" (also wrong — only true for the default stat); it now states both halves. Worth spelling out because the failure mode each time was asserting a general rule from one measured case.

**Redundant guard**: left as-is. `attach_series_group_axis()` does re-check, but the outer `data_has_series_groups()` in `attach_group_axis()` guards the `resolve_group_mapping()` call above it too, so dropping it would change what runs, not just what is checked.

Full suite after these changes: `[ FAIL 0 | WARN 2 | SKIP 31 | PASS 3934 ]`.

---

jooyoungseo:
`b5341c2` merges `main` in. Flagging it because the merge was **not** clean: `generate_selectors()` in `R/ggplot2_smooth_layer_processor.R` was rewritten by both this PR and #86, and git resolved it by interleaving the two around the `return(list())` / `collect_all_polyline_grobs` text they share — the same anchoring pathology that has bitten the test files on this branch before. Naive marker removal would have produced a syntactically valid but semantically wrong method, so it was reconstructed by hand.

The resolved method, in order:

```r
generate_selectors = function(plot, gt = NULL, panel_ctx = NULL,
                              built = NULL, panel_id = NULL) {
  if (is.null(gt)) {                        # #86: no gtable, no N to read off
    return(list())
  }

  n_series <- self$series_group_count(plot, built, panel_id)
  if (n_series > 1L) {                      # this PR: grouped-curve path
    grouped <- self$grouped_curve_selectors(plot, gt, panel_ctx, n_series)
    if (!is.null(grouped)) return(grouped)
    return(list())                          # give up rather than emit one selector for n curves
  }

  ...                                       # #86: single-curve tail, fabricated ids removed
}
```

The `is.null(gt)` guard is placed first deliberately. It is not just #86's branch relocated: `grouped_curve_selectors()` already returns `NULL` for a `NULL` gtable, and this PR's give-up then returns `list()`, so both paths agreed on the answer — hoisting the guard makes that one statement instead of two, with no behaviour change either way.

Verification, all after the merge:

- **Independently reproduced.** I resolved the conflict here from the markers, and separately in a throwaway worktree that merged all four PRs (#84/#85/#86/#87) in a different order. The two resolutions are **byte-identical** — `diff` reports no difference in either `R/ggplot2_smooth_layer_processor.R` or `tests/testthat/test-ggplot2-smooth-layer-processor.R`.
- **Bite check on the merged tree.** Reverting only `R/ggplot2_smooth_layer_processor.R` to `origin/main` — which already carries #86's fix but not this one — and re-running `test-ggplot2-smooth-layer-processor.R`: **FAIL 21 | PASS 472**. The grouped-curve work is still load-bearing after the merge, not accidentally satisfied by #86's changes.
- **Full suite on the merged tree:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4197`. Same 2 pre-existing warnings.
- `NEWS.md` conflicted as usual; resolved by keeping every entry (118 bullets, none dropped).

---

PR #88, opened 2026-08-08
fix: ship KaTeX and stop declaring the placeholder stylesheet

jooyoungseo:
## What this fixes

maidr **3.75.1** split KaTeX out of `dist/maidr.css` into `dist/maidr-math.css`. `maidr.css` became a 406-byte placeholder, published only so that `<link>` tags written before the split keep resolving — upstream's README stopped instructing anyone to link it.

This package tracked that release (the bundle on `main` is already 3.75.1) without picking up the change, so today:

- `maidr.css` is still declared by both HTML dependencies, the htmlwidgets manifest, and the standalone document builder — a request that cannot change what the page looks like.
- **`maidr-math.css` is not bundled at all.** `maidr.js` fetches it at runtime, resolved against the URL it was itself loaded from, so the local bundle path 404s and LaTeX in AI chat responses renders unstyled. The failure is silent and narrow: only a reader who opens the chat and receives mathematics ever sees it.

## What changed

**`.github/scripts/fetch-maidr-bundle.sh`** installs `maidr-math.css` instead of `maidr.css`, and refuses a maidr older than 3.75.1 with a sentence rather than tar's "Not found in archive".

**The font stripping moved into that script.** It was `tools/trim-maidr-css.R`, which the comments asked maintainers to re-run after every refresh — but `update-maidr-bundle.yml` calls only the fetch script, so every scheduled update silently re-inflated the package until someone noticed by hand. It now happens during the fetch and is verified rather than assumed: the fonts have to be gone *and* KaTeX's rules have to have survived, or the script fails. `tools/trim-maidr-css.R` is deleted as a result — flagging that explicitly in case you would rather keep it.

The trade is the one this package already documented for CRAN, applied to the file upstream moved the fonts into: **19 kB of layout rules instead of 368 kB**, with glyphs falling back to system fonts while spacing, fractions, radicals and delimiters render correctly.

**No stylesheet is declared anywhere now.** Both dependencies and the manifest drop it; `create_standalone_html()`'s CDN branch drops the `<link>`. The offline branch is the one exception and the reason the inlined stylesheet does not simply disappear — its script is inline, so it has no URL for the runtime to resolve against — and it now inlines `maidr-math.css` where it inlined `maidr.css` before.

`inst/COPYRIGHTS`, `cran-comments.md` and `NEWS.md` follow the file rename and the moved stripping step.

## Verification

- `maidr_local_assets()`, `maidr_html_dependencies()` (both modes) and `maidr_inline_asset_tags()` exercised directly against the repo tree: the maths stylesheet resolves beside `maidr.js`, neither dependency declares a stylesheet, and the inlined `<style>` is 18.7 kB carrying `.katex` rules with no `@font-face`.
- The refreshed bundle was installed by the fetch script itself, so the npm tarball was verified against its published sha512.
- New tests in `tests/testthat/test-html-dependencies.R` pin all of the above, including that the manifest and the dependency agree.
- **Not run locally:** `R CMD check` and the testthat suite — this environment has R but not the package's dependency tree. CI covers both.

The same upstream change is handled for py-maidr in xability/py-maidr#291.

---
_Generated by [Claude Code](https://claude.ai/code/session_0195xCq4AQ6273n8yWCGv3t1)_


jooyoungseo:
Both review rounds addressed; closing out the two items left open.

**shellcheck — already covered by CI, and green.** The `Lint workflows and shell scripts` job runs it on every push and passed on `3682566` ([job](https://github.com/xability/r-maidr/actions/runs/31278883949/job/93156847181)), including the +61-line growth in `fetch-maidr-bundle.sh`. Worth noting because the one construct that would plausibly have tripped it — `grep -qx ... <<<"$TARBALL_FILES"` — is there deliberately: under `set -o pipefail` a `tar | grep -q` pipeline reports *failure on success*, because grep exits at the first match and tar dies of SIGPIPE. I hit exactly that while writing it (the script refused to bundle 3.75.1 for not shipping a file it does ship), which is why the listing goes through a variable.

**`R CMD check` / testthat** — still CI's call, running now across all four platforms. I did exercise the changed functions directly against the tree rather than leaving them entirely to CI: `create_standalone_html()` on both branches, and `maidr_local_assets()` / `maidr_html_dependencies()` / `maidr_inline_asset_tags()`. All six of the new `test-svg-utils.R` assertions pass that way, and the two strings the offline branch asserts *absent* (`cdn.jsdelivr.net`, `@font-face`) appear nowhere in the 1.9 MB inlined `maidr.js` either, so they are not passing by accident.

**On the `sed` assumption** — agreed it rests on a single-line minified file, and it is worth saying why that is load-bearing rather than incidental: `sed` works a line at a time, so a multi-line stylesheet would strip only the blocks that fit on one line and leave the rest. That is precisely why the two verification `grep`s follow it and fail the script: the fonts have to be gone *and* `.katex` has to have survived. A future KaTeX build that ships unminified fails the first check loudly instead of quietly shipping 368 kB into a CRAN package.

The alternative — `perl -0pe`, which would be slurp-mode and line-agnostic — would add a dependency the script does not currently have (bash, curl, jq, openssl, tar), and Git Bash on Windows is where that bites. Happy to switch if you would rather not depend on the minification.

---

PR #90, opened 2026-08-08
chore(examples): regenerate the ggplot2 baselines

jooyoungseo:
The HTML under `inst/output/` predates several merged fixes, so the artifacts a reader downloads describe behaviour the package no longer has. This is the deliberate regeneration pass.

## What was stale, and how I know

Two of the differences are merged fixes finally showing through. Both were measured, not inferred.

**`example_patchwork_2x2_ggplot2.html` shipped four layers all called `maidr-layer-1`.** #78 made composition layer ids unique; the baseline still displayed the collision that PR fixed.

```
before: maidr-layer-1, maidr-layer-1, maidr-layer-1, maidr-layer-1   (1 distinct of 4)
after : maidr-layer-1-1-1, maidr-layer-1-2-1, maidr-layer-2-1-1, maidr-layer-2-2-1
```

Resolved in headless Chromium (`page.setContent` + `document.querySelectorAll`, not string surgery):

```
maidr-layer-1-1-1  line  data=8  selectors=1 matched=[1]
maidr-layer-1-2-1  bar   data=5  selectors=1 matched=[5]
maidr-layer-2-1-1  bar   data=5  selectors=1 matched=[5]
maidr-layer-2-2-1  line  data=8  selectors=1 matched=[1]
```

**`example_boxplot_horizontal_ggplot2.html` gave two of setosa's four outliers no highlight target.** `ggplot_build()` puts 2 below the lower fence and 2 above the upper one, and the baseline's `data` listed all four — but its `selectors.lowerOutliers` / `selectors.upperOutliers` held one entry each, so two announced outliers pointed at nothing.

```
ggplot2 ground truth  group 1 (setosa): 4 outliers -> 1.1, 1, 1.9, 1.9 | lower: 2 | upper: 2
before                box0 selectors: lowerOutliers=1  upperOutliers=1
after                 box0 selectors: lowerOutliers=2  upperOutliers=2
```

*(An earlier revision of this description said the baseline "announced two of four". That was loose — the values were all announced; it was the highlight targets that were missing. Corrected after the review pass caught it.)*

Two further differences, both faithful:

- `example_multiline_plot_ggplot2.html` gains `"z": {"label": "Strategy"}`, matching the plot's own `labs(color = "Strategy")`, with each of the three series carrying its own `z`.
- `example_violin_plot_ggplot2.html`'s first curve simplifies to 14 retained KDE levels instead of 15 (28 points instead of 30). That is RDP output under the unchanged 15-level cap, deterministic across three fresh R processes, and the emitted y range still matches the drawn polygon exactly — `21.4000..33.9000` against `ggplot_build()`'s `21.4000..33.9000`, with no duplicate levels.

**Also in every diff, and not an R-side behaviour change:** the bundled asset link moves from `lib/maidr-3.67.0/` to `lib/maidr-3.75.1/`, because `MAIDR_VERSION` in `R/html_dependencies.R:4` was already bumped to `3.75.1` on `main`. Expected for a "make the baselines match current `main`" pass — flagging it so nobody diffing these goes looking for a third behaviour change that isn't there.

The three faceted baselines change only in their layer ids, which embed `Sys.time()` and `Sys.getpid()` by design (`generate_unique_id()`, `R/svg_utils.R:22`). Those files can never be byte-stable across regenerations; that is a property of the id scheme, not something this pass introduces.

## Whole-set verification

Every baseline in the diff, before and after, loaded in Chromium and every selector resolved with a real CSS engine. A layer's `selectors` is a string, a list of strings, or — for `box` and `violin_box` — a `BoxSelector` object keyed `iq`/`q2`/`min`/`max`/`lowerOutliers`/`upperOutliers`; the string leaves are what gets resolved. (My first pass handed the whole object to `querySelectorAll` and reported three "invalid" selectors on the box and violin files. That was my measurement error, not a defect, and the numbers below are from the corrected run.)

```
                                          before            after
unresolved selectors                        0                 0
invalid selectors                           0                 0
patchwork distinct layer ids               1/4               4/4
```

Full suite on this branch: `FAIL 0 | WARN 2 | SKIP 31 | PASS 4197`, the same two pre-existing box-plot notch warnings. No test reads `inst/output/`, so nothing here is load-bearing for CI — the artifacts are documentation.

## Candlestick baselines are NOT regenerated

`tidyquant` does not build in this environment — `survival` fails to compile, taking `clock`, `lava`, `prodlim`, `ipred`, `recipes` and `timetk` with it. Tests 19-20 were skipped and `example_candlestick_ggplot2.html` / `example_candlestick_ma_volume_ggplot2.html` are left exactly as they are rather than guessed at. Anyone with `tidyquant` installed can run the generator and pick those two up.

## Two changes to the generator, both needed to run it

- `library(devtools)` → `library(pkgload)`. `load_all()` lives in pkgload; devtools only re-exports it. Without this the script cannot run in an environment that has the package-loading tooling but not the whole devtools stack.
- `library(tidyquant)` was a hard call, so its absence aborted the entire script and blocked the other eighteen baselines. It is now optional; tests 19-20 and their summary lines announce themselves as skipped. With tidyquant present the behaviour is unchanged.

Also: `inst/output/lib/` — the 1.9 MB htmlwidgets asset bundle `save_html()` drops next to the examples — is now gitignored. It was previously untracked *and* unignored, so anyone running the generator saw it as new work.


jooyoungseo:
Both points taken.

**Bundle version.** Verified — you're right, and it belongs in the PR body:

```
before: lib/maidr-3.67.0
after : lib/maidr-3.75.1
R/html_dependencies.R:4  MAIDR_VERSION <- "3.75.1"   (already on main)
```

Every baseline's asset link moves for that reason, independent of the two R-side deltas I itemised. Nobody diffing these should go hunting for a third behaviour change — noted below and in the PR body.

**Naming.** Fixed in `9c6ec15`: `HAVE_TIDYQUANT` → `have_tidyquant`. Confirmed the six remaining `line_length_linter` hits in that file are all pre-existing lines, none from this diff.

One correction to your outlier note, since it changes what the fix was. You wrote that the old baseline's `data` already listed 4 outliers and only the `selectors` arrays were short. Checking again, that is right and my PR body was loose about it: the mismatch was between `data` and `selectors`, not a change in the announced values. So the user-visible defect was that two of setosa's four outliers had **no highlight target**, while all four were still announced. I'll say it that way rather than "announced two of four".

---

jooyoungseo:
Right on both counts, and thank you for diffing against the merge base — that is what surfaced the second one.

**Stray `lib/`.** Same artifact, same cause: an ad-hoc render run at the worktree root rather than through the generator, so `save_html()` dropped the bundle beside the working directory instead of beside `inst/output/`. Removed in `251d264`, and `.gitignore` now covers both landing sites rather than just the one I'd noticed:

```
inst/output/lib/
/lib/
```

**`main` moving under the branch.** You flagged the KaTeX commit as unrelated; it is not — #88 stops the package declaring a stylesheet at all, which changes every baseline's `<head>`. So the branch as it stood would have merged already-stale artifacts. Merged `main` in and re-ran the generator (`a72814b`, `deef015`):

```
before  <link href="lib/maidr-3.67.0/maidr.css">
after   (no stylesheet declared)
script  lib/maidr-3.75.1/maidr.js
```

Payload structure is identical to the previous regeneration — same patchwork id fix, same multiline `z` axis, same violin RDP level count — so nothing about the substantive claims changes. Re-swept in Chromium across all eighteen: **0 unresolved, 0 invalid**, patchwork ids `4/4`.

Also took your earlier naming note: `have_tidyquant` is snake_case now.

---

PR #91, opened 2026-08-08
fix(ggplot2): omit the selector field when a layer has no highlight target

### LINKED ISSUE #89: Regression from #86: an empty selector list kills the whole figure
**This is a regression I introduced in #86 (`b9a1cb1`), merged today.** Filing it against myself so the fix has a reference.

#86 made histogram, stacked bar, dodged bar and smooth return `list()` instead of a fabricated selector when the grob lookup finds nothing. That was the right call for the payload — but `list()` serialises to `[]`, and `[]` is not a value the bundled frontend can take.

## What breaks

`document.querySelectorAll([])` throws. An empty array stringifies to `""`, and an empty selector is a `SyntaxError`, measured in Chromium:

```
empty array     THREW: SyntaxError — Failed to execute 'querySelectorAll' on 'Document': The provided selector is empty.
one selector    ok, matched 0
empty string    THREW: SyntaxError — Failed to execute 'querySelectorAll' on 'Document': The provided selector is empty.
```

`maidr.js` 3.75.1 hands `layer.selectors` straight to `selectAllElements`, guarded only by `if (!n) return null` — and `[]` is truthy. The throw happens inside the trace constructor, so the **figure never initialises**: no announcement, no sonification, no braille, no keyboard entry. The chart still renders visually, so nothing signals the author that it is dead.

## Reachable input

`facet_wrap(~g, drop = FALSE)` over a factor with an unused level — the same input #86's own tests use.

```r
d <- data.frame(x = c("A","B","A","B"), y = c(10,20,30,40),
                fill = c("u","v","u","v"),
                g = factor(c("p","p","q","q"), levels = c("p","q","r")))
ggplot(d, aes(x, y, fill = fill)) +
  geom_bar(stat = "identity", position = "stack") +
  facet_wrap(~g, drop = FALSE)
```

## Measured, before and after

Each file rendered, loaded from disk in Chromium, then driven with the keys a reader would use (click, <kbd>Enter</kbd>, arrows, <kbd>PageDown</kbd>), capturing `pageerror` and `console.error`:

| render | `e23f860` (pre-#86) | `474d126` (post-#86) |
|---|---|---|
| `geom_histogram()` + empty facet | clean | **PAGEERROR** `querySelectorAll ... selector is empty` |
| `geom_bar(position="stack")` + empty facet | clean | **PAGEERROR** same |
| `geom_bar(position="dodge")` + empty facet | clean | **PAGEERROR** same |
| `geom_smooth()` + empty facet | clean | clean |

Before #86 these were clean only because the fabricated selector was a *non-empty* string that matched nothing — useless, but harmless to parse. So #86 traded "one panel highlights nothing" for "the entire chart is inaccessible", on the three rect layers.

The smooth case stays clean because its trace type does not take the `Segmented` path that reaches `querySelectorAll` with the raw list.

## Which neutral value the frontend actually accepts

Same page, three payload variants, only the `maidr-data` attribute edited:

| payload | result |
|---|---|
| `"selectors":[]` | figure never initialises |
| `selectors` key absent | **clean** |
| the layer removed entirely | **clean** |

So "no selector" has to be expressed by **omitting the key**, or by not emitting the layer at all — never by an empty list.

## Fix

Two parts, because there are two distinct situations:

1. **A panel that drew nothing should not produce a layer.** `combine_facet_layer_data()` (`R/ggplot2_facet_utils.R:304`) does `combined_data <- c(combined_data, list(result$data))`, which wraps a zero-length `list()` into a one-element list. That both creates the empty-selector layer and emits `"data":[[]]`. Skipping zero-length results there makes the panel come out with `layers = list()`, which the frontend already tolerates — the Base R `layout()` path emits zero-layer cells today and loads without error.

2. **A layer that has data but genuinely resolved no selector must omit the key.** This is #84's grouped-smooth give-up and #86's guards on a non-facet input. `selectors = list()` must not reach the payload as `[]` anywhere.

A frontend guard (`if (!n || n.length === 0) return null` in `mapToSvgElements`) is worth having too, but it belongs in xability/maidr and does not help anyone on the currently bundled 3.75.1, so the R side has to be correct on its own.

## Related, found while measuring this

An empty facet panel also emits `"data":[[]]` — a one-element array whose element is empty. On a box plot that survives initialisation and then announces `"cat is undefined, lower outlier(s) v is undefined"`, with `Failed to set the 'value' property on 'AudioParam': The provided float value is non-finite.` from the sonification. Same root cause in the combiner, so part 1 above fixes it.

jooyoungseo:
Closes #89. **This is a regression I introduced in #86 (`b9a1cb1`), merged earlier today.**

#86 replaced four fabricated selectors with `list()`. That was the right answer for the payload and the wrong value on the wire: it serialises to `[]`, the frontend hands `layer.selectors` straight to `document.querySelectorAll()`, and an empty array stringifies to `""` — a `SyntaxError`, raised inside the trace being constructed.

## Measured

Headless Chromium, each page loaded from disk and driven with the keys a reader uses (click, <kbd>Enter</kbd>, arrows, <kbd>PageDown</kbd>), counting `[id^=maidr-highlight-]` elements that appear during navigation:

| render | `e23f860` | `474d126` (#86) | this branch |
|---|---|---|---|
| `geom_histogram()` + empty facet | 1 highlight | **0 + pageerror** | 1 highlight |
| `geom_bar(position="stack")` + empty facet | 1 highlight | **0 + pageerror** | 1 highlight |
| `geom_bar(position="dodge")` + empty facet | 1 highlight | **0 + pageerror** | 1 highlight |
| `geom_smooth()` + empty facet | 1 highlight | 1 highlight | 1 highlight |

The page error is `Failed to execute 'querySelectorAll' on 'Document': The provided selector is empty.` Reachable input is `facet_wrap(~g, drop = FALSE)` over a factor with an unused level — the same shape #86's own tests use. The chart still renders visually, so nothing signals the author.

Note on wording: the static wrapper markup (`#maidr-figure-*`) is present either way, so "the figure never initialises" would overstate it. What is measurably gone is the highlighting — navigation produces no highlight element at all.

I isolated the cause by editing **only** the `maidr-data` attribute of one broken page and reloading:

| payload | result |
|---|---|
| `"selectors":[]` | 0 highlights + pageerror |
| `selectors` key absent | clean |
| the layer removed entirely | clean |

So "no selector" must be expressed by omitting the key, never by an empty list. The smooth case stays clean throughout because its trace type does not take the path that reaches `querySelectorAll` with the raw list — which is why #86's own end-to-end test, written against smooth, did not catch this.

## The fix

**`drop_empty_selectors()`** runs over the payload in `set_maidr_data_attr()` — the single point every payload passes through — and removes a `selectors` entry that has no entries. A `BoxSelector` object is left alone; only a zero-length list goes. Doing it here rather than in each processor keeps `list()` as the honest return value for a processor whose grob lookup found nothing.

**`combine_facet_layer_data()`** no longer wraps a processor's empty result into one empty series. That was emitting `"data":[[]]` for an empty facet level; a reader entering that panel was told it held a box plot and then heard `cat is undefined, lower outlier(s) v is undefined`, with the sonification throwing `Failed to set the 'value' property on 'AudioParam': The provided float value is non-finite.` The panel now carries no layers — the same shape the Base R `layout()` path already emits for a cell with no plot, which loads clean today.

```
subplot ...-1: 1 layer(s)  type=box dataLen=2
subplot ...-2: 1 layer(s)  type=box dataLen=2
subplot ...-3: 0 layer(s)          <- was 1 layer with data [[]]
```

## Verification

**Bite checks**, each fix reverted to `origin/main` independently and `test-empty-selector-payload.R` re-run:

| reverted | result |
|---|---|
| both | **FAIL 7** \| PASS 19 |
| `R/svg_utils.R` only | **FAIL 1** \| PASS 20 |
| `R/ggplot2_facet_utils.R` only | **FAIL 1** \| PASS 24 |
| neither | FAIL 0 \| PASS 25 |

Both are load-bearing; neither subsumes the other.

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4224` against a `main` baseline of 4197. Same two pre-existing box-plot notch warnings.

**roxygen** regenerated; only `man/drop_empty_selectors.Rd` is new, and the three perennially-stale pages were reverted. **lintr** clean on both changed files and the new test (the two remaining hits are pre-existing: `ggplot2_facet_utils.R:90` signature indentation, untouched here).

## A test that was pinning the bug

`test-fabricated-selector-guard.R` — which I added in #86 — asserted **three** selector entries, one of them empty. It was pinning the broken shape. Rewritten to two, with the reason recorded beside it, plus an explicit assertion that no layer may carry an empty selector list.

## Not done here

A frontend guard (`if (!n || n.length === 0) return null` in `mapToSvgElements`) is worth having, but it belongs in xability/maidr and would not help anyone on the currently bundled 3.75.1, so the R side has to be correct on its own.

---


jooyoungseo:
Good catch, and it was exactly what you guessed. `save_html()` drops its htmlwidgets asset bundle next to whatever file it writes; I rendered the Chromium test pages into the worktree root, and `git add -A` swept `lib/` into the commit. Confirmed byte-identical to the tracked copy before removing it:

```
maidr.js   root=31e4cd117be36734  inst=31e4cd117be36734  same=YES
maidr.css  root=f31bba9dc597fe7d  inst=f31bba9dc597fe7d  same=YES
```

`b7b14a8` removes it and gitignores `/lib/` so the next render cannot repeat it. The diff is now 7 files, +303/-11 — down from 1244 added lines.

Full suite re-run after the cleanup: `FAIL 0 | WARN 2 | SKIP 31 | PASS 4224`, unchanged.

---

PR #101, opened 2026-08-08
fix(ggplot2): survive an NA in a faceted stacked bar's facet column

### LINKED ISSUE #92: Faceted stacked bar: an NA in the facet column makes save_html() throw
Verified by rendering, from the #60 Medium leads. Worse than the lead described: it does not merely inject `NA` rows, it aborts the export.

**Where:** `R/ggplot2_stacked_bar_layer_processor.R:84-95`, the `==` comparison on line 88. (Filed as line 98 in #60; drifted.)

## Reproduce

```r
df <- data.frame(
  g    = c(rep("a", 4), rep("b", 4), NA, NA),
  cat  = c("x","x","y","y", "x","x","y","y", "x","y"),
  fill = c("u","v","u","v", "u","v","u","v", "u","v"),
  val  = c(10, 30, 20, 40,  11, 31, 21, 41, 99, 98)
)
p <- ggplot(df, aes(x = cat, y = val, fill = fill)) +
  geom_bar(stat = "identity", position = "stack") + facet_wrap(~g)
maidr::save_html(p, file = tempfile(fileext = ".html"))
```

```
Error in order(group_data[[x_col]]) : argument 1 is not a vector
Calls: save_html ... <Anonymous> -> <Anonymous> -> lapply -> FUN -> order
```

No file is produced.

## Mechanism

`original_data[original_data[[facet_var]] == group, ]` selects with `NA` for every row whose facet value is `NA`, and `[` turns an `NA` index into an all-`NA` row. Replaying the filter per panel:

```
PANEL 1  facet_groups = a    rows kept:  6  (all-NA rows: 2)
PANEL 2  facet_groups = b    rows kept:  6  (all-NA rows: 2)
PANEL 3  facet_groups = NA   rows kept: 10  (all-NA rows: 10)
```

For the real levels the injected rows are usually absorbed by `split()`, which drops `NA` groups — so `a` and `b` survive *when the NA rows sort last*. Move them to the top of the frame and the `setNames(as.character(original_data[[fill_col]]), built_data_layer$fill)` alignment shifts, and **every** panel dies:

```
PANEL 1 ( g = a )  -> ERROR: argument 1 is not a vector
PANEL 2 ( g = b )  -> ERROR: argument 1 is not a vector
PANEL 3 ( g = NA ) -> ERROR: argument 1 is not a vector
```

Causality confirmed by patching only that filter in a throwaway copy (NA-safe `%in%` plus an explicit `is.na()` branch): the identical NA-rows-first plot then renders with correct per-panel payloads — `a`: `v` 30/40, `u` 10/20; `b`: `v` 31/41, `u` 11/21; the `NA` panel: `{"x":"x","y":99,"z":"u"}`.

Controls: no `NA` in the facet column renders fine. `stat = "count"` with `NA` in the facet column renders fine — that branch never touches `original_data`. So this is scoped to the `stat = "identity"` / `geom_col()` path.

## Severity

Any faceted stacked or filled bar built with `geom_col()` whose facet column contains a single `NA` — routine in survey and panel data — produces **no accessible output at all**, with an error that names nothing the user recognises. There is no workaround they could discover.

## Fix sketch

Replace the `==` subset with an NA-safe predicate: `%in%` for ordinary levels, `is.na(original_data[[facet_var]])` when the panel's own facet group is `NA`, then subset with the resulting logical. Independently, the `stat = "identity"` branch should stop assuming `nrow(original_data) == nrow(built_data_layer)` — the `setNames()` silently pads names with `NA` on any length mismatch — and the `fill_groups[[as.character(fill_value)]]` lookup should guard against `NULL` rather than falling into `order(NULL)`.

jooyoungseo:
Closes #92.

`original_data[original_data[[facet_var]] == group, ]` selects with `NA` for every row whose facet value is missing, and `[` turns an `NA` index into an all-`NA` row. Those rows then reach `order()`, which refuses them.

## Before / after

Rendered through the full pipeline, both row orderings, because the ordering changes which panels die:

```
                  origin/main                              this branch
NA rows LAST   -> ERROR: argument 1 is not a vector      -> OK
NA rows FIRST  -> ERROR: argument 1 is not a vector      -> OK
```

After, both orderings produce the same payload (only the session-sequential grob ids differ):

```
panel title=a    series=2   [[{"x":"x","y":30,"z":"v"},{"x":"y","y":40,"z":"v"}],
                              [{"x":"x","y":10,"z":"u"},{"x":"y","y":20,"z":"u"}]]
panel title=b    series=2   [[{"x":"x","y":31,"z":"v"},{"x":"y","y":41,"z":"v"}],
                              [{"x":"x","y":11,"z":"u"},{"x":"y","y":21,"z":"u"}]]
panel title=NA   series=1   [[{"x":"x","y":99,"z":"u"}]]
```

I reproduced both the failure and the fix independently of the authoring pass.

## What the NA panel announces: `"NA"`

Decided from what ggplot2 draws, not from taste. `built$layout$layout$g` holds `NA_character_`, the strip's `titleGrob` label is `NA_character_`, and grid renders that as the literal two glyphs. The exported SVG's `<text>` nodes read `"a" "b" "NA" "x" "y" …`, so `"NA"` is what a sighted reader sees on that strip; `Missing` or `Unknown` would announce a word that is not on the chart. No code change was needed for this — the existing `paste(facet_groups, collapse = " & ")` already coerces — but the test pins it **against the exported SVG text** rather than against a chosen literal.

## The three sub-problems the issue asked about

**The `==` filter.** New internal `facet_group_rows(values, group)` in `ggplot2_facet_utils.R`: `%in%` for ordinary levels, `is.na()` when the panel's own group is `NA`, always a logical with no `NA`s. A column literally holding the string `"NA"` stays distinct from a missing value.

**(a) The row-count assumption — genuinely needed, and reachable with no facet at all.** Added `rows_aligned <- nrow(original_data) == nrow(built_data_layer)` to the `stat = "identity"` guard; a mismatch falls through to the built-data branch. On `ggplot(big, …) + geom_col(data = small)`, main announces `{"x":"z","y":60}` / `{"x":"z","y":50}` — a category the layer never drew — because `setNames()` padded the names with `NA`. After the fix the payload is exactly the four drawn bars.

**(b) `order(NULL)`.** `stacking_order` is filtered to fill levels `split()` actually produced.

**One more, found while fixing the above.** Rows ggplot2 could not position stay in the built data with `x`/`ymin`/`ymax` = `NA` (ggplot2 4.0.3 does not delete them). They poisoned `min(x)`, so `x == min(x)` answered `NA` everywhere and the stacking order came back empty. They are dropped before `min(x)` is read, which turns the no-facet `geom_col()`-with-a-missing-`y` crash into correct drawn-consistent output.

## Verification

**Bite check:** the new test file copied onto a pristine `origin/main` worktree — **0 of 6 blocks pass, 0 of 39 expectations pass**. Two expectation failures in block 1 (`expect_no_error`); the other five blocks error with `argument 1 is not a vector`.

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4283` with `main` merged in.

**Baseline note:** `origin/main` measures `PASS 4217` in this environment, not the 4197 in the issue-era notes — that figure predates #84 and #88.

**Selectors resolved in Chromium** (`page.setContent` + `document.querySelectorAll`, not string surgery):

```
panel[0][0] title="a"  series=2 points=4   "#geom_rect\.rect\.2\.1 rect" -> 4
panel[0][1] title="b"  series=2 points=4   "#geom_rect\.rect\.4\.1 rect" -> 4
panel[0][2] title="NA" series=1 points=1   "#geom_rect\.rect\.6\.1 rect" -> 2
```

The real panels line up exactly. The NA panel's 1-against-2 gap is a pre-existing defect, described below. No empty selector list anywhere, per #89.

roxygen regenerated (one genuine new page, `man/facet_group_rows.Rd`; the three perennially-stale pages reverted). lintr identical to baseline on both touched R files, 0 on the new test. No new test dependencies.

## Found here, not fixed

1. **The dodged processor has the byte-identical `==` filter** (`ggplot2_dodged_bar_layer_processor.R:148-158`). It does not crash — it silently emits an empty series for the NA panel, under **both** stats. Same root cause, one line with the same helper, which is why `facet_group_rows()` lives in `ggplot2_facet_utils.R`. Left out because it is a different user-visible symptom needing its own regression test and its own decision about what that panel announces. Filed separately.
2. **`NA` in the *fill* column.** ggplot2 draws that segment grey with an `"NA"` legend key; the payload omits it. The `stacking_order` guard makes this at least consistent — previously it crashed when the NA segment sat in the first bar and was silently omitted otherwise; now it is always a silent omission, so the rest of the chart stays accessible. Announcing it properly means deciding the `z` label and re-checking rect alignment.
3. **A fill level absent from the first x category is dropped**, faceted or not, because `stacking_order` is derived only from the bar at `min(x)`. Confirmed pre-existing and unchanged here, with no `NA` involved: `data.frame(cat = c("x","y","y"), fill = c("u","u","v"), val = c(99,10,98))` emits only the `u` series on main and on this branch. That is why the NA panel above shows 1 point against 2 rects — and the issue's own stated expected payload for that panel has the same gap. Fixing it means choosing between omitting the cell and scoring it `0`, which the dodged processor's comment argues opposite ways for `count` and `identity`.

---


jooyoungseo:
Good catch on the coverage gap — you're right that two of the three fixes here are reachable with no facet at all and were only demonstrated in the PR body. `1024292` pins both, and both bite on `origin/main`:

```
BITE on origin/main (whole file): FAIL 4 | PASS 0
  :220  "a layer whose own data is shorter…"        -> ERROR
  :250  "rows ggplot2 could not position…"          -> 2 FAILURES
```

Full suite now `FAIL 0 | WARN 2 | SKIP 31 | PASS 4291`.

**Writing the second test corrected something in my own PR description.** I had asserted the announced values would be exactly the ones ggplot2 positioned. They are not. The unpositionable cell is still announced, as `(y, 0, u)`:

```
BUILT  y: 40,30,0,40   x: 1,1,NA,2
SERIES (x,30,v) (y,40,v)
SERIES (x,10,u) (y,0,u)
```

ggplot2 gives that row `x = NA` and draws no rect, so a value of `0` goes out for a bar that is not on screen. This change only stops the row poisoning `min(x)` — it does not filter the cell out. Whether such a cell should be announced as `0` or omitted is the same open question the dodged processor's comment argues opposite ways for `count` and `identity`, so the test pins the current behaviour explicitly rather than asserting it is right. I have said so in the test and would not want the PR to read as if it settled that.

On the `length(group) != 1L` nit: you're right that it is unreachable from the only call site today. I kept it and said why in a comment — a zero-length or vector group degrading to "the missing panel" is better than recycling silently through `%in%`.

The redundant `!is.na()` in the `stacking_order` filter: agreed it is redundant, and I left it for the reason you gave.

---

PR #103, opened 2026-08-08
fix(ggplot2): give a grouped geom_line() one selector per curve

### LINKED ISSUE #95: A grouped geom_line() beside geom_smooth() emits 3 series but 1 selector, so the line layer loses highlighting entirely
Verified by rendering. This is the line-layer twin of the smooth bug fixed in #84 — **#84 did not cover it**, confirmed by diffing `R/ggplot2_line_layer_processor.R` across the merge (its edits to that file are pure helper extraction).

**Where:** `R/ggplot2_line_layer_processor.R:473-480`

```r
if (length(poly_ids) > 1L &&
  !is.null(line_layer_position) &&
  line_layer_position <= length(poly_ids)) {
  pid <- poly_ids[line_layer_position]
  ...
  return(list(paste0("#", escaped)))
}
```

The branch assumes one `GRID.polyline.N` grob per line *layer* — correct for the candlestick + N × `geom_ma` case it was written for. A grouped `geom_line()` puts its curves as *children of one grob* (`GRID.polyline.1.1.{1,2,3}`), while `geom_smooth()` contributes three separate grobs. So `length(poly_ids) == 4 > 1`, the branch fires, and returns a single selector pointing at the first child.

## Reproduce

```r
set.seed(42)
df <- data.frame(x = rep(1:8, 3), g = factor(rep(c("g1","g2","g3"), each = 8)),
  y = c(1:8+rnorm(8,0,.3), 10+1:8*.5+rnorm(8,0,.3), 20-1:8*.4+rnorm(8,0,.3)))
ggplot(df, aes(x, y, colour = g)) + geom_line() +
  geom_smooth(method = "lm", formula = y ~ x, se = FALSE)
```

```
LAYER 1 type=line    length(data)=3  length(selectors)=1
   series z: g1, g2, g3
   selector[1]: #GRID\.polyline\.1\.1\.1
LAYER 2 type=smooth  length(data)=3  length(selectors)=3
```

All three line polylines exist in the DOM; the payload names one:

```
GRID.polyline.1.1.1  stroke=rgb(248,118,109)   <- the only one emitted (g1)
GRID.polyline.1.1.2  stroke=rgb(0,186,56)      <- g2, never referenced
GRID.polyline.1.1.3  stroke=rgb(97,156,255)    <- g3, never referenced
```

## The symptom is *no* highlight, not a misplaced one

The bundled multiline class opens with a hard precondition, verbatim from `inst/htmlwidgets/lib/maidr-3.75.1/maidr.js`:

```js
mapToSvgElements(n){if(!n||n.length!==this.lineValues.length)return null;...
```

With `lineValues.length === 3` and `n.length === 1` this returns `null`, so `highlightValues` is null and every point in all three series falls to the out-of-bounds state. The visual highlight for the line layer is dead, not misplaced. (The comment at `:470-471` names this precondition; the code below it violates it.)

## Scope

Each measured in a fresh R process:

| plot | line layer |
|---|---|
| `geom_line(aes(colour=g))` alone | 3 series / 3 selectors — OK |
| `+ geom_point()` | 3 / 3 — OK |
| `+ geom_smooth(se=FALSE)` | **3 / 1** |
| `geom_smooth(se=FALSE) + geom_line(aes(colour=g))` | **3 / 1** (order-independent) |
| `+ geom_smooth(se=TRUE)` | **3 / 1** |
| ungrouped `geom_line() + geom_smooth()` | 1 / 1 — OK |

Both conditions are needed: a grouped line layer *and* a second polyline-producing layer in the panel. Pre-existing since `d1a374b`.

## Severity

On one of the most common exploratory ggplots — grouped series with a trend overlay — the line layer loses visual highlighting completely. A blind user navigating the three lines gets audio, text and braille, but nothing on screen moves, so a sighted collaborator cannot follow along and screen-magnifier users lose their position anchor. It degrades silently, and the smooth layer *beside* it now highlights correctly, so the two layers of one plot behave differently with no explanation.

## Fix sketch

The branch conflates "one polyline grob per line layer" with "one polyline grob per series". Resolve *this* layer's grob first, then enumerate that grob's children, rather than indexing the flat panel-wide `poly_ids` by layer position — emit one selector per child sized to the number of series the layer produced. #84 already built this machinery for the smooth processor (`geom_grob_prefix()` and the per-group chunking in `R/series_group_utils.R`), including the right failure mode: when the children cannot be lined up with the groups, emit *no* selector rather than a wrong-length list. Reusing those helpers here would close this and keep the two processors consistent. The `find_poly` walk at `:446-462` also wants checking — it matches `^GRID\.polyline\.\d+$`, so it only ever sees top-level grob names, not the children the selectors ultimately address.

Note the "emit no selector" path must omit the key rather than send an empty list — see #89.

jooyoungseo:
Closes #95. This is the line-layer twin of the smooth-layer bug #84 fixed; #84 did not cover it, confirmed by diffing `R/ggplot2_line_layer_processor.R` across that merge — its edits there were pure helper extraction.

## The frontend precondition, read directly

`inst/htmlwidgets/lib/maidr-3.75.1/maidr.js`, **byte offset 689026**, inside the multiline trace class:

```js
mapToSvgElements(n){if(!n||n.length!==this.lineValues.length)return null;
  const r=this.mapViaDomElements(n);return r||this.mapViaPathParsing(n)}
```

and at **byte offset 37988** in its base class:

```js
get highlight(){return this.highlightValues===null||this.isInitialEntry
  ?this.outOfBoundsState:{empty:!1,elements:this.highlightValues[this.row][this.col]}}
```

`1 !== 3` → `highlightValues` null → every point in all three series reports `outOfBoundsState`. So the symptom is **no highlight at all**, not a misplaced one.

## Root cause — and a correction to the issue

The issue points at the `panel_ctx` branch. That branch had the same defect, but for a plain single-panel ggplot the orchestrator passes **no** `panel_ctx`, so the reproduction actually went through the twin at `:503-530` (`find_all_polyline_grobs`). Both descended into `geom_smooth.gTree.N` and both indexed a flat panel-wide list by line-layer position. Collapsed onto one path.

Measured grob tree: a grouped `geom_line()` is a **bare** `polyline GRID.polyline.1` with `id = 1,2,3` — one grob, three curves — as a direct child of the panel gTree, while `geom_smooth()` is `geom_smooth.gTree.6` wrapping `GRID.polyline.2/3/4`. `GeomPath$draw_panel` returns `polylineGrob(...)` with no `ggname()`, which is why a line layer's grob carries no geom prefix and only draw order identifies it.

On the `find_poly` walk at `:446-462`: `^GRID\.polyline\.\d+$` is the right pattern — the `.1.k` children are gridSVG output, not grobs, and the pattern is what keeps theme grid lines out. The defect was descending into *other layers'* trees.

## Before / after

`data` / `selectors` for the line layer, fresh R process per row. I re-measured this independently of the authoring pass:

| plot | before | after |
|---|---|---|
| grouped `geom_line()` alone | 3 / 3 | 3 / 3 |
| `+ geom_point()` | 3 / 3 | 3 / 3 |
| `+ geom_smooth(se = FALSE)` | **3 / 1** | **3 / 3** |
| `geom_smooth(se = FALSE) + geom_line()` | **3 / 1** | **3 / 3** |
| `+ geom_smooth(se = TRUE)` | **3 / 1** | **3 / 3** |
| ungrouped line + smooth | 1 / 1 | 1 / 1 |
| two grouped `geom_line()` layers | **6 / 2** | **6 / 6** |
| grouped line + smooth under `facet_wrap` | **6 / 4** | **6 / 6** |

The last two are outside the issue's table; same defect, same fix. Faceted panels get distinct grobs (`GRID.polyline.1.1.*` vs `.2.1.*`), so panel scoping still holds.

## Browser verification

Rendered with `save_html(use_cdn = FALSE)`, loaded over `file://` in Chromium with the real bundle, driven by keyboard. The highlight circle's centre matched a curve vertex at distance **0.00** each step:

| navigation | highlighted polyline | stroke |
|---|---|---|
| ArrowRight (g1) | `GRID.polyline.1.1.1` | `rgb(248,118,109)` |
| ArrowUp (g2) | `GRID.polyline.1.1.2` | `rgb(0,186,56)` |
| ArrowUp (g3) | `GRID.polyline.1.1.3` | `rgb(97,156,255)` |

The same script against `origin/main` produces **no highlight element at all** at every step.

## On issue #89

Checked rather than assumed: `selectAllElements` is `document.querySelectorAll(e)`, so `[]` → `""` → SyntaxError *for traces that reach it*. The multiline trace does not — `mapToSvgElements([])` returns null at the length guard first. Loading the figure three ways (normal, `selectors: []`, key deleted) gave zero page errors in all three. That matches what I measured for #89 independently: only the `Segmented` rect traces died. `list()` is kept as the give-up value, matching what #84 does in the smooth processor, and #89's `drop_empty_selectors()` strips it centrally before serialisation.

## Verification

**Bite check**, the new tests copied onto pristine `origin/main`: `FAIL 8 | ERROR 1 | PASS 142`. Five of eight new blocks fail there. One of those is a weaker bite than the rest and I want to be precise about it: the "no selector" failure-mode test **errors** with `unused argument (n_series = 7)` rather than failing behaviourally, because `n_series` is a new parameter. The three blocks that pass on main are the deliberate guards (grouped line alone, ungrouped line + smooth, two line layers over a bar).

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4303` with `main` merged in.

**Merge with #100.** Both PRs rewrote parts of `R/ggplot2_line_layer_processor.R` and both appended a section to its test file after the *same* `# ====` rule, so git swallowed the second separator. Reconstructed by hand, `parse()`-checked, 16 section rules and 53 tests with no duplicate names. Then re-ran **both** PRs' measurements on the merged branch to confirm they coexist — #95's table above and #100's transformed-scale table both still hold.

## Candlestick + N × geom_ma — not verified, and I will not claim it

The guarding test is `"candlestick + 2 MA layers in same panel collapse into 2 layers"` (`tests/testthat/test-ggplot2-candlestick-patchwork.R:75`), which asserts `length(line_layer$selectors) == 2`. **It SKIPPED — tidyquant is not installed here** — as do all five tests in that file.

A structural proxy was added that needs no tidyquant: `geom_col() + geom_line() + geom_line()`, several single-curve line overlays in one panel over a non-polyline layer, asserting each line layer resolves to its own grob with one distinct selector. That passes. The reasoning it should carry over: `GeomMA` extends `GeomLine`, so each MA is a bare direct-child polyline with one `id` group, `line_layer_position()` indexes them as before, and the curve-count check is `1 == 1`; the skip-set only ever removes candidates, and candlestick layers never entered the count. That is reasoning, not measurement.

## Found here, not fixed

1. **The `grob_id` faceted branch** still returns one `.1.1` selector regardless of series count. It has **no non-NULL caller anywhere in `R/`** — only a test reaches it — so it was left rather than change a dead path.
2. **One-point groups.** `GeomPath` silently drops groups with fewer than two points from the grob while `extract_data()` still emits them as a series. The curve-count check then emits no selector where main emitted one wrong one. That is the better failure, but it is a behaviour change on that edge.
3. `find_main_polyline_grob()` was already dead before this change and was left alone. `find_all_polyline_grobs()` **was** removed — this change is what orphaned it, and its semantics ("all polylines in the panel, descending into other layers' trees") are precisely the bug.

---


jooyoungseo:
Good catch on the stale comment — and I'd argue it belongs in this PR rather than a follow-up, since this change is what made it untrue. `b6cfa58` rewrites it.

The old text said the generator discovers *all* polyline grobs in the panel and hands every one of N line layers the same length-N set — which is exactly what the dedupe existed to undo. That is no longer the input shape. The new text says so, and says what the trim/pad is still for, which has not changed: the frontend needs one selector per merged series.

I left the dedupe in place rather than removing it. It is now a backstop rather than the mechanism, and removing it would be a behaviour change on a path this PR has no measurement for.

Full suite after the doc change: `FAIL 0 | WARN 2 | SKIP 31 | PASS 4303`, unchanged. lintr on `ggplot2_patchwork_utils.R` is 1, against 15 on main's copy of the same file — the difference is `object_usage_linter` noise from linting main's version outside the package, not anything this diff introduces.

---

jooyoungseo:
You're right, and the "trim/pad" wording was mine from `b6cfa58`, not pre-existing — so it belongs in this PR. `d49c6b3` fixes it and says what actually happens:

> There is deliberately no pad: a short list fails that precondition and the frontend drops the layer's highlight rather than aiming it at the wrong curve.

That is the same fail-closed choice the processor itself makes, so it is worth stating rather than leaving as an absence.

Chasing it also turned up a second stale comment I had missed — the inline note above the dedupe still carried the exact claim I had just corrected in the roxygen block:

```r
# Dedupe selectors (panel_ctx path returns the same set for each line layer
# in the panel), preserving discovery order.
```

Fixed in the same commit. It now says the dedupe should be a no-op post-fix and why it is kept anyway: a duplicate would otherwise consume a series' slot in the trim.

Comment-only, no executable change; lintr unchanged at 1.

---

PR #104, opened 2026-08-09
feat(pie): support ggplot2 and Base R pie charts

jooyoungseo:
## Description

Adds pie chart support to both plotting systems: `geom_col()`/`geom_bar()` + `coord_polar("y")` for ggplot2, and `graphics::pie()` for Base R.

Companion to xability/maidr#767, which adds `TraceType.PIE` to the renderer. **That PR must ship first** — the JS `TraceFactory` throws on an unknown trace type, so a pie layer is only renderable once the bundled JS catches up. Every assertion here is against the emitted JSON payload and the exported SVG, which are correct today.

### ggplot2

`Ggplot2Adapter$detect_layer_type()` gains the package's **first coordinate-system logic** — it never read `plot_object$coordinates` before, and there was no coord handling anywhere in `R/`. `coord_radial()` is handled alongside `coord_polar()` (it does not inherit `CoordPolar`), and `theta = "x"` is explicitly *not* a pie: that is a coxcomb/rose, which keeps its existing bar behaviour.

Under polar coordinates the panel holds a single `polygonGrob` named `geom_rect.polygon.<N>` — **not** the `geom_rect.rect.<N>` both bar processors grep for, which is why every pie resolved to an empty selector list before this change.

Two details that bite:

- Slice labels live on the **fill** aesthetic, not `x` (`x` is the literal `""`), so they resolve through the scale's `get_labels()`.
- Labels are indexed **by** the built group id, not by position among the ids present — otherwise a facet panel missing a category shifts every remaining label by one. There is a test for exactly that.
- `panel_params` under `coord_polar` has no `$x`/`$y` (its keys are `r.range`, `theta.range`, …), so nothing copied from the bar processor may touch them.

Selectors return `list()` rather than a fabricated grob id when a lookup comes up empty — grid's counter is session-wide, so a guessed name silently lands on another panel's wedges (#83).

### Base R

**Base R `pie()` could not be exported by gridSVG at all before this PR.** `grid.echo()` emits pie's label text grobs with `vjust = NA`, and gridSVG's `justTovjust()` throws `missing value where TRUE/FALSE needed` out of `devGrob.text`. Reproduced on gridSVG 1.7.7 / R 4.3.3:

```r
pie(c(A = 3, B = 5, C = 2))   # grid.export() -> FAIL
pie(c(3, 5, 2), labels = NA)  # grid.export() -> OK
barplot(c(A = 3, B = 5, C = 2))  # OK
```

`repair_na_text_justification()` rewrites only the NA components of text grobs, immediately before the existing `grid.export()` call site. Scoped so valid justification passes through untouched, and verified to produce **byte-identical PostScript** — `0.5` is exactly what grid resolves NA to for the `just = "centre"` these grobs declare — so there is no visual change. It is documented as an upstream gridSVG/gridGraphics incompatibility that should be dropped if gridSVG ever handles NA itself.

After the repair, wedges export as `graphics-plot-<p>-polygon-<k>.1.1`, one polygon per wedge, index-aligned to the data. This is structurally unlike `barplot`, whose n bars all live inside one grob — a pie needs n selectors.

`pie` was already classified HIGH and already stubbed, exported and wrapped; the only gap was `detect_layer_type` returning `"unknown"`. Wedge order is left in recorded-call order — the SVG is replayed from the same arguments, so any re-sort would desynchronise data from the polygon ids.

Also fixed while in the file: `BaseRProcessorFactory$get_available_processors()` was omitting `BaseRHeatmapLayerProcessor`.

### Known breakage, fixed

`tests/testthat/test-panel-scoped-fallback.R` used `pie()` as its canonical *unsupported* high-level call. It is supported now, so that is repointed at `mosaicplot()` — verified empirically that mosaicplot/dotchart/stem/pairs still route to `"unknown"` before relying on it. A stale comment in `base_r_plot_orchestrator.R` asserting pie cannot survive SVG export is removed, since it no longer can be true.

## Testing

`testthat::test_local()` — **FAIL 0 | WARN 2 | SKIP 96 | PASS 3624**, ending "Nice code."

- The 2 warnings are pre-existing `boxplot(notch = TRUE)` warnings in `test-base-r-orchestrator.R`, unrelated to pie.
- The 96 skips are `{patchwork}`/`{tidyquant}` not being installed in this environment, plus one long-standing `skip()`.

New: `test-ggplot2-pie-layer-processor.R` (25 blocks) and `test-base-r-pie-layer-processor.R` (22 blocks), plus a pie case added to the fabricated-selector guard and a new `test-svg-utils.R` section that reproduces the upstream gridSVG bug and asserts the repair fixes it. Both processors are verified end-to-end through `save_html()`, with selectors resolved against the real exported SVG via xml2, and index alignment checked geometrically (`pie(c(A=100, B=1, C=1))` → polygon-1 has 197 vertices vs 3 and 3).

Docs updated: README's supported-types table, `DESCRIPTION`, `R/maidr-package.R`, `pkgdown/extra/llms.txt`, `NEWS.md`, the getting-started vignette, the examples article, and `inst/examples/{ggplot2,base_r}/pie.R` so `run_example("pie")` works.

`man/` regenerated with `roxygen2::roxygenise()` pinned to 7.3.2 to match `RoxygenNote` (`devtools` will not install here — `textshaping`/`ragg` need harfbuzz headers). `NAMESPACE` came out byte-identical; the pie processors are internal R6 classes.

## Notes for the reviewer

- Two `man/` diffs are unrelated to pie but are genuine `document()` output from stale checked-in files: `maidr-package.Rd` (maintainer fields now matching `DESCRIPTION`) and `dot-maidr_chartseries_ta_warned.Rd` (a `\code{}` reverting to a backtick, since the package has roxygen markdown off). Flagging so they don't read as hand-edits — if either is unwanted, the fix belongs in the source, not in `man/`.
- Neither implementation rejects negative values. ggplot2 reports `ymax - ymin`, i.e. the magnitude the wedge actually subtends; base R defers to `pie()`'s own rejection. I preferred honesty-to-the-drawing over inventing a rejection path with no precedent here, but say the word if a hard reject is wanted.
- The patchwork leaf selector path is not exercised — `{patchwork}` is not installable in this environment. The facet tests cover the same `find_gtable_panel_grob()` branch.

---


jooyoungseo:
All three points addressed in 6464beb. The bullseye one was a real bug and the "low-risk edge case" turned out to be real too.

## 1. Bullseye misclassification — fixed

Reproduced first:

```r
df <- data.frame(ring = rep(c("R1","R2"), each = 3), sub = rep(c("a","b","c"), 2), val = 1:6)
p <- ggplot(df, aes(x = ring, y = val, fill = sub)) + geom_col() + coord_polar("y")
Ggplot2Adapter$new()$detect_layer_type(p$layers[[1]], p)   # "pie"  ← wrong
```

`is_pie_coord()` now also requires the layer's built `x` to hold at most one distinct value **per panel**. Reading built data rather than the mapping expression is what makes `aes(x = "")`, a one-level factor, a constant character column and `aes(x = 1)` all count alike. A rejected layer falls through to the existing dodge/stack/bar checks — deliberately not `"unknown"`, so those charts keep exactly the behaviour they have on `main`.

Verified classifications: idiomatic pie → `pie`; 2-ring bullseye → `stacked_bar`; **1-ring bullseye → `pie`** (it really is one); bullseye without fill → `bar`; coxcomb → `stacked_bar`; `coord_radial(theta="y")` → `pie`; polar histogram → `hist`; cartesian cases unchanged. I also diffed the full emitted payload for bullseye/coxcomb/fill-less-bullseye against a detached worktree at `0fd2de4` — byte-identical `type`, data, selectors and axes, so nothing that works today changed.

One judgement call to flag: the ring test is **per panel**, not global. A plot faceted so each panel holds one x category is therefore a pie (one per panel), which I think is right and consistent with "a single-ring bullseye is a pie". A global test would call it a bar. Only that contrived middle case differs.

Roxygen note added stating multi-ring/bullseye polar bars are out of scope, plus a NEWS.md amendment.

## 2. `\keyword{}` pollution — root cause found, fixed at source

The cause is **R6 method docblocks that open with untagged title text instead of `@description`**. Reproduced in a throwaway package: with `#' Initialize the demo adapter` above a method, roxygen2 7.3.2 emits `Undocumented R6 method` and scatters that sentence's individual words into the *class-level* block as `\keyword{}` entries — and a title following an already-tagged block gets glued onto the previous `@return` instead, which is why `can_handle`'s Returns had "Detect the type of a single layer…" appended to it. Adding `@description` makes both symptoms vanish.

So I fixed the source, not the generated file: `@description` added to every R6 method docblock in `ggplot2_adapter.R` and `base_r_adapter.R`. Both man pages now carry exactly one `\keyword{internal}` and the Returns bleed is gone. `git diff --stat man/` shows only the three intended files, `NAMESPACE` is unchanged, and a second `roxygenise()` is idempotent.

**Scoping note:** the same root cause affects 15 more `.Rd` files already polluted on `main` (`BaseRLineLayerProcessor`, `Ggplot2PlotOrchestrator`, `ProcessorFactory`, …). I fixed only the two files you named to keep this diff scoped. The sweep is mechanical but touches ~15 source files — I'll file it as a follow-up issue rather than smuggle it in here.

## 3. `aes_groups = list("fill", "x")` — you were right, it does misbehave

Not just theoretical. Written and run:

```r
ggplot(df, aes(y = units, fill = fruit)) + geom_col(aes(x = "")) + coord_polar("y")
```

resolved to `aes = "x"`, giving wedge labels `1, 2, 3` instead of `Apples, Bananas, Cherries`, and an x-axis label of `"x"` instead of `"fruit"`. The shared helper loops mapping-outer/aesthetic-inner, so it probes the layer for *both* fill and x before consulting the plot.

Fixed in `resolve_slice_mapping()` by calling the helper twice — `list("fill")` first, `list("x")` only as fallback — so fill is exhausted at both levels before x is considered. The shared helper's semantics are untouched. Tests pin all three arrangements.

---

`testthat::test_local()` → **FAIL 0 | WARN 2 | SKIP 96 | PASS 3652**. The 2 warnings are the same pre-existing `boxplot(notch = TRUE)` ones; the 96 skips are missing optional suggests.

One cost I should name: `draws_single_ring()` calls `ggplot_build()` during classification. It only runs for a bar/col layer already known to be under a polar coord with `theta == "y"`, so nothing non-polar pays for it, but a pie now builds once more per `detect_layer_type()` than before. I didn't add caching — the adapter has no per-plot lifetime to hang one off, and inventing one looked like more risk than the saving is worth. Happy to revisit if you disagree.

---

PR #108, opened 2026-08-09
fix(base-r): report quantmod attach-order masking instead of misblaming the user

### LINKED ISSUE #97: chartSeries() is not recorded when quantmod is attached after maidr, and NEWS claims otherwise
Verified against a real `R CMD INSTALL` of the package into a private library, so namespace sealing and `library()` semantics are genuine rather than a pkgload artifact. From the #60 Medium leads.

**Where:** `R/base_r_wrapper_exports.R:101-110`, with the hook at `R/ggplot2_system_init.R:29`.

## Measured

```
MODE: maidr_first          # library(maidr); library(quantmod)
search: .GlobalEnv | package:quantmod | package:TTR | package:xts
chartSeries resolves to  : <environment: namespace:quantmod>
== maidr::chartSeries    : FALSE
chartSeries call         : drew ok
n recorded calls         : 0
save_html                : SAVE ERROR: No Base R plots detected. Please create a plot first (e.g., barplot(), plot()).

MODE: quantmod_first       # library(quantmod); library(maidr)
== maidr::chartSeries    : TRUE
n recorded calls         : 1
save_html                : OK
emitted layer type: candlestick | n data: 20

MODE: ns_only              # library(maidr); loadNamespace("quantmod")
== maidr::chartSeries    : TRUE
chartSeries call         : CALL ERROR: no applicable method for `@` applied to an object of class "name"
n recorded calls         : 0
save_html                : SAVE ERROR: No Base R plots detected.
```

## Mechanism

Plain masking, and neither existing mitigation addresses it. The `setHook(packageEvent("quantmod", "onLoad"), ...)` calls `wrap_function("chartSeries")`, and `wrap_function()` assigns into `asNamespace("maidr")` — **maidr's own namespace, never quantmod's**. When `library(quantmod)` attaches at search position 2, ahead of `package:maidr`, an unqualified `chartSeries(...)` binds to `quantmod::chartSeries` and the recording stub is never entered.

`NEWS.md:307` says *"`chartSeries()` calls are now recorded even when 'quantmod' is loaded after 'maidr'"*. That is false for the attach case, which is the documented flow — `vignettes/articles/examples.qmd:578` shows a bare `chartSeries(TST, type = "candlesticks", ...)`.

The namespace-only case is a separate, second break: the hook *does* fire and the stub *is* reached, but the call then dies inside quantmod with `no applicable method for `@` applied to an object of class "name"` (traceback: `quantmod::chartSeries(...) -> sapply(chob@passed.args$TA, function(x) eval(x@call))`).

## Severity

The most natural script — `library(maidr); library(quantmod); chartSeries(SPY)` — produces a plain inaccessible graphic and then a hard `stop()` at export: *"No Base R plots detected. Please create a plot first"*. No candlestick sonification at all, and the error actively misleads by suggesting nothing was drawn. Reordering the two `library()` calls fixes it, which is undiscoverable.

## Fix sketch

Masking cannot be solved from maidr's own namespace. Either patch quantmod's binding directly — in the `packageEvent("quantmod", "attach")` hook, not just `"onLoad"`: `unlockBinding()` / `assignInNamespace()` plus a matching assign into the attached `package:quantmod` environment, so the wrapper wins regardless of search order — or, if patching a foreign package is unacceptable, drop the false NEWS claim, have `.onAttach` detect a later-attached quantmod and say plainly that quantmod must be loaded first (or `maidr::chartSeries()` called explicitly), and make the "No Base R plots detected" error name this case when `package:quantmod` precedes `package:maidr` on the search path. The `@`-slot crash on the namespace-only path needs a separate look at how the stub's `...` forwarding reaches `chob@passed.args$TA`.

## Incidental, measured while here

`save_html()`'s replay of a recorded `chartSeries(x, type = "candlesticks", TA = NULL)` drops the explicit `TA = NULL` and lets quantmod re-add the default `addVo()`. A/B: the user's own call emits no warnings; the `save_html()` call emits `number of items to replace is not a multiple of replacement length`, which comes from `addVo`. Plain `quantmod::chartSeries(x, type = "candlesticks", TA = NULL)` emits none.

jooyoungseo:
Closes #97.

## Before / after, real `R CMD INSTALL` into a private library

```
                        BEFORE (main)                          AFTER
maidr_first    recorded 0; save_html: "No Base R plots     attach-time message naming the cause;
               detected. Please create a plot first        the same export error, now naming it too
               (e.g., barplot(), plot())."                 recorded still 0 — by design, see below

quantmod_first recorded 1; save_html OK                    unchanged

ns_only        CALL ERROR: no applicable method for `@`    drew ok; recorded 1; save_html OK
               applied to an object of class "name"        (candlestick, 20 points)
```

Escape hatch verified under `maidr_first`: `maidr::chartSeries(...)` records and exports correctly.

I reproduced the `@`-slot crash and its fix myself: unqualified `chartSeries(x, type = "candlesticks", TA = NULL)` gives `CALL ERROR` on `main` and `drew ok` on this branch.

## Route taken: (2), do NOT patch quantmod — and route (1) is measurably unsafe

I asked for route (1) to be ruled in or out on evidence rather than taste. It is out, and here is the measurement. quantmod builds `chob@passed.args <- as.list(match.call(expand.dots = TRUE)[-1])`. Forwarding through `...` corrupts that record:

```
direct  f(x, type="c", TA=NULL):     passed.args$TA class=NULL       deparse=NULL
via ... stub(x, type="c", TA=NULL):  passed.args$TA class=name       deparse=..3
direct  f(x, TA="addVo()"):          passed.args$TA class=character  deparse="addVo()"
via ... stub(x, TA="addVo()"):       passed.args$TA class=character  deparse="addVo()"
```

Route (1) installs a `...`-forwarding wrapper **inside quantmod's namespace**, so quantmod's own internal `chartSeries()` calls — `reChart`, `zoomChart` — would inherit that corruption, in a foreign package's internals where neither maidr's tests nor its users can see it. That is on top of CRAN's prohibition on modifying another package's namespace. I verified this table independently.

**The trade-off, and the thing to overrule me on if you disagree.** Under route (2), `library(maidr); library(quantmod); chartSeries(SPY)` still produces **no sonification**. The user gets a message and an accurate error instead of a silent failure and a false one, but they must change their script. Route (1) is the only way to make the naive script Just Work. Two concrete costs of this choice:

- **Noise.** The attach hook fires on every `library(quantmod)` while maidr is attached, including for people who never call `chartSeries()`. It is a `packageStartupMessage`, so both `suppressPackageStartupMessages()` and `options(maidr.startup_message = FALSE)` silence it — but it is shown to people who do not need it. If that is unacceptable, deleting the attach hook and keeping only the improved export error is a one-function change.
- **Discoverability is improved, not solved.** The user still has to read and act.

Verified for the chosen route: survives `detach()` → re-`library()`; `quantmod::chartSeries` is provably still quantmod's own function with no maidr code in its namespace body; `unloadNamespace("maidr")` leaves **0** hooks on both `attach` and `onLoad`.

## The `@`-slot crash

Not a subtle interaction — a plain `...`-forwarding defect, the same one the table above shows. `TA = ..3` is a `name`, so quantmod's `if (!is.null(TA))` is skipped (the *evaluated* `TA` is `NULL`) and `passed.args$TA` keeps the symbol, which then reaches `sapply(chob@passed.args$TA, function(x) eval(x@call))`. It fires **only** for an explicit `TA = NULL`; the default and `TA = "addVo()"` are fine because quantmod rebuilds the entry from the evaluated value.

`quantmod_first` never hit it because the generated wrapper's existing `retry_call_in_caller_frame()` rescued it silently; the hand-written stub in `base_r_wrapper_exports.R` had no retry. It now has the same one — established pattern, single-call fast path preserved, retry only on the exceptional error.

## Correcting the incidental note I filed on #97

I wrote that `save_html()`'s replay drops the explicit `TA = NULL` and lets quantmod re-add `addVo()`. **That is wrong**, and I have measured it three ways:

```
quantmod::chartSeries(TA=NULL) alone           warnings=0
save_html of a plain barplot                   warnings=0
save_html of chartSeries(TA=NULL)              warnings=1  [number of items to replace ...]
```

`TA = NULL` is present in the recorded args, survives `clean_maidr_args()`, and the replay itself emits nothing. The traceback puts the warning in `gridSVG::grid.export` → `primToDev.rect` → `devRect` → `svgRectString` — a length mismatch in gridSVG's own vectorised-rect export, triggered by chartSeries' single vectorised `rect()` call for the candle bodies. Not fixed: gridSVG is upstream and unmaintained since 2017, and fixing it means replacing the export pipeline.

**Worth knowing:** on `main` this warning is invisible only because the call crashes first. Fixing the crash makes it visible. That is a previously-hidden warning surfacing, not a new one — the measurements above separate the two.

## Verification

**Bite checks**, two of them, because there are two independent fixes:

| | result |
|---|---|
| fix in place | `FAIL 0 \| PASS 28` |
| `quantmod_masks_maidr()` forced FALSE | **`FAIL 14`** \| PASS 14 |
| `@`-slot retry reverted in the stub | **`FAIL 2`** \| PASS 26 |
| both restored | `FAIL 0 \| PASS 28` |

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4378`, against a `main` baseline of 4350 measured in a throwaway worktree. +28 is exactly the new file.

## What the tests do NOT cover — stated plainly

**The masking itself.** Which function a bare `chartSeries(...)` resolves to is fixed by search-path order at `library()` time, and a running testthat process cannot re-attach maidr behind quantmod. The tests `attach()` a stand-in `package:quantmod` frame ahead of `package:maidr`, which reproduces the **search-path state** the diagnostics key off and so pins every message this adds — but it does **not** prove that a real `library(quantmod)` masks the wrapper. That end is verified only out of process, by the `R CMD INSTALL` transcripts above. The same caveat is in a header comment in the test file.

## NEWS

The false claim is corrected in place rather than deleted, so the record shows it was wrong and when. Two new development-version bullets cover the masking diagnostics and the `@`-slot fix.

## Found here, not fixed

The gridSVG rect warning (above). The now-vestigial `.maidr_quantmod_onload_hook`: post-sealing, its `wrap_function()` call is a no-op for both the namespace and the locked `package:maidr`, but it still populates `.saved_graphics_fns`, so removing it risks a replay regression outside this issue's scope.

---


jooyoungseo:
Both observations are fair. Neither changed the code, and here is why for each.

**`call. = FALSE` — you're right that it is a behaviour change beyond the stated fix, and the PR body should have said so.** Recording it here rather than leaving it as an unexplained diff.

It is deliberate and it is load-bearing for this change specifically. The message is now two paragraphs:

```
No Base R plots detected. Please create a plot first (e.g., barplot(), plot()).
'quantmod' is attached ahead of 'maidr' on the search path, so a bare chartSeries()
call goes straight to quantmod and is not recorded. Attach 'quantmod' before 'maidr',
or call maidr::chartSeries() explicitly.
```

With `call. = TRUE` R prefixes `Error in show(plot) :` and wraps the whole thing around the call deparse, which pushes the actionable second paragraph behind noise the reader does not need — the failure is about search-path order, not about which function they called. It also matches the prevailing style: **25 of the package's 50 `stop()` calls already pass `call. = FALSE`**. If you would rather keep the prefix I will revert those three, but the message reads worse for it.

**File placement.** I looked before deciding not to move it. `ggplot2_system_init.R` already holds `.onLoad`, `.onUnload` and `.onAttach` — the package-level lifecycle hooks — despite its name. All three of the new helpers exist to serve those hooks, so they sit next to their only callers. The mismatch you spotted is real, but it is the *file name* that is wrong, not the placement: this file stopped being ggplot2-specific some time ago. Moving three functions out would leave the hooks separated from their helpers and not fix the misnomer. Renaming the file is the change that would actually resolve it, and that belongs on its own rather than buried in a quantmod fix.

**`.maidr_quantmod_onload_hook`** — agreed it wants tracking rather than a comment. I will file it once this lands, so the issue can point at the merged code.

---

PR #110, opened 2026-08-09
fix(ggplot2): highlight the bar a geom_col() is announcing

### LINKED ISSUE #94: geom_col() with an incomplete frame emits ragged series, so the highlight lands on the wrong bar
Verified by rendering and by replaying the bundled frontend's own regrouping against the exported rects. Surfaced while working #80 and deliberately left out of #87; filing it now that it is measured.

**Where:**
- `R/ggplot2_stacked_bar_layer_processor.R:159-170` — series length is `nrow(group_data)` after `split(original_data, original_data[[fill_col]])`
- `R/ggplot2_dodged_bar_layer_processor.R:232-245` — series length is `length(rows_by_fill[[fill_name]])`

The dodged processor's own comment, added in #87, already states the invariant this branch violates: *"raggedness is what the frontend cannot survive"*, *"`barValues` must be RECTANGULAR"*. #87 made that true for `stat = "count"` and left `stat = "identity"` alone.

## Reproduce

```r
# the caller's frame has no row for (q, B) -- a gap in the MIDDLE
df <- data.frame(x = c("p","p","q","r","r"), g = c("A","B","A","A","B"),
                 n = c(10, 20, 30, 50, 60), stringsAsFactors = FALSE)
ggplot(df, aes(x = x, y = n, fill = g)) + geom_col(position = "dodge")
```

```
dodged  series[1] n=3 z=A  (p,10) (q,30) (r,50)
        series[2] n=2 z=B  (p,20) (r,60)          -> RAGGED
stacked series[1] n=2 z=B  (p,20) (r,60)
        series[2] n=3 z=A  (p,10) (q,30) (r,50)   -> RAGGED
```

## What the reader actually gets

Running the bundled `mapToSvgElements` rect-branch loop, copied verbatim out of `inst/htmlwidgets/lib/maidr-3.75.1/maidr.js`, inside Chromium against the real rendered DOM:

`geom_col(position = "dodge")`:
```
series[0] z=A  x=q  y=30 -> geom_rect.rect.2.1.4  fill=rgb(0,191,196)   x=[718,776]
series[0] z=A  x=r  y=50 -> EMPTY (no highlight)
series[1] z=B  x=r  y=60 -> geom_rect.rect.2.1.3  fill=rgb(248,118,109) x=[530,647]
```
Announcing **"q, 30, group A"** highlights the teal **B** bar at **x=r**. Announcing **"r, 60, group B"** highlights the red **A** bar at **x=q**. Wrong category and wrong group, cross-mapped.

`geom_col()` stacked — `barValues[0].length = 2`, so the walk only ever visits 2 of the 3 columns:
```
series[0] z=B  x=r  y=60 -> geom_rect.rect.2.1.4  (which is A's segment)
series[1] z=A  x=r  y=50 -> EMPTY
never assigned: geom_rect.rect.2.1.5
```
An entire column is unreachable — never announced, never highlighted.

Control: the `stat = "count"` path from #87 on the same shape is rectangular `[3,3]` and maps perfectly, every A point to a red rect, every B point to a teal rect, the absent cell to `EMPTY`.

## Severity

On a `geom_col()` whose frame is not a complete grid — routine for pre-aggregated tidy data, which is the reason `geom_col()` exists — the highlight lands on a different bar than the one being announced, sometimes a different category in a different colour group. A sighted assistant and a blind user looking at "the same" bar are looking at different bars, with nothing indicating disagreement.

## What the frontend requires — and why "just zero-fill it" is not the whole answer

Reading the bundle, the contract is three separate requirements:

1. **Rectangular `data`.** The column loop is bounded by `barValues[0].length` and the node cursor is shared across all series, so raggedness either truncates the walk or runs the cursor off the end.
2. **A cell that maps to no rect must be exactly `0`.** That is the only sentinel the loop understands: `(sparse && barValues[s][col] === 0) || k >= n.length ? emptyElement() : n[k++]`. There is no `null`/`NA` path — `null` fails `Number(l.y)` to `NaN`, never equals `0`, and consumes a node, reintroducing the shift.
3. **`sparse` must be true**, i.e. rect count strictly less than total slots. Zero-filling satisfies that automatically.

So structurally the payload needs the full cartesian product with `0` in the absent cells — identical to #87. The objection #87 raised is **semantic, not structural**: for `geom_col()` a `0` would be *announced* as the value zero when the truth is "the caller supplied no value here".

The resolution is to keep `0` as the frontend's placeholder — a protocol token meaning "no rect", not a datum — and fix the announcement: carry a per-point marker so the reader hears "no data" rather than "0". If the schema cannot express that today, the honest interim is still to zero-fill, because a bar chart that highlights the wrong bar is strictly worse than one that says "0" for a row the caller omitted. What must not ship is the current ragged payload, which the repo's own comments already call unsurvivable.

jooyoungseo:
Closes #94.

## First: my issue text was wrong, and it changes the answer

I wrote that `0` is the only sentinel the regrouping loop understands, and that `null` "fails `Number(l.y)` to `NaN`, never equals `0`". Measured in Chromium:

```
Number(null)                0
Number(null) === 0          true
Number(undefined)           NaN
Number(undefined) === 0     false
```

`Number(null)` is `0`. `null` reaches the same `=== 0` test, consumes no node, and gets an empty highlight — exactly as a zero would. (`undefined` is the one that breaks, so *omitting* the key would not work.) And the announcement never reads `barValues` at all: the text channel reads the raw point, through a formatter with a dedicated missing branch (`missingText:r="missing"`, one occurrence in the bundle).

So the trade-off the issue framed — rectangular grid *or* honest announcement — is a false one. The schema can carry both.

**Absent cells therefore emit `NA` → JSON `null`, not `0`.** Measured in the real widget:

| payload | absent cell announces | highlight | page errors |
|---|---|---|---|
| ragged (before) | `x is r, n is 66, g is B` → **wrong bar** | wrong rect | AudioParam non-finite |
| `y: 0` | `x is q, n is 0, g is B` | correct (none) | none |
| **`y: null`** | **`x is q, n is missing, g is B`** | correct (none) | none |
| `y: 0` + `missing: true` | `n is 0` — extra field ignored | correct | none |

The extra-field option is dead weight. The package already emits `null` y for SMA lines, and `set_maidr_data_attr()` is already configured `na = "null"` for exactly this.

`stat = "count"` keeps `0`, unchanged from #87 — an absent cross-tab cell genuinely counted nothing.

## Before / after

Series lengths and values, verified independently of the authoring pass:

```
                     origin/main                      this branch
dodge gap MIDDLE     3,2  [10,30,50] [20,60]          3,3  [10,30,50] [20,null,60]
dodge gap START      3,2  [10,30,50] [40,60]          3,3  [10,30,50] [null,40,60]
stack gap MIDDLE     2,3  [20,60] [10,30,50]          3,3  [20,null,60] [10,30,50]
stack gap START      3    [10,30,50]                  3,3  [null,40,60] [10,30,50]
```

Replaying the bundled regrouping in Chromium against the real DOM, "cells correctly mapped / mis-mapped":

| | dodge start | dodge mid | dodge end | stack start | stack mid | stack end |
|---|---|---|---|---|---|---|
| before | 0 / 5 | 2 / 3 | 4 / 1 | 0 / 3, x non-monotonic | 3 / 2 | 4 / 1 |
| after | **6 / 0** | **6 / 0** | **6 / 0** | **6 / 0** | **6 / 0** | **6 / 0** |

Controls unchanged: `stat = "count"` dodged and stacked still 6/0, still `0`, dodged still `groupDirection: "forward"`.

## Two more bugs, found by testing the gap at the START

Both were outside the issue's measurements and are fixed here:

**A whole series could vanish.** `stack gap START` on `main` emits **one** series — fill B never appears at all. `stacking_order` was read off the first column via `min(x)`, so a fill level absent there was dropped from the payload entirely. It now reads the order off the **fullest** column, ties broken by smallest x, so complete grids order exactly as before.

**An existing test was passing on a coincidence.** The NA-facet panel in `test-ggplot2-facet-na-group.R` has rows `(x,u)` and `(y,v)`, giving `[1,1]` — rectangular by accident, but column 0 meant `x` in one series and `y` in the other, so each series was handed the other's rect. Now a proper 2×2 grid.

## Verification

**Bite check:** fix reverted, tests kept → **66 failures + 3 errors across 19 of 21 blocks**. The only two clean blocks are the guards asserting *unchanged* behaviour (dodged keeps the default reverse walk; `stat = "count"` still emits a real `0`) — those must pass both ways.

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4574` with `main` merged in, against a baseline of 4350 measured in a throwaway worktree.

One process note worth recording, since it bears on how much the rest of these numbers are worth: the authoring pass initially wrote its branch suite figure into its report *before* that run had finished. It turned out correct, and it self-corrected unprompted, but the number above is from my own run after merging `main`, not from that report.

roxygen: `man/` is byte-identical to `origin/main`. lintr: dodged 0 (main 0), stacked 5 (main 5, all pre-existing), both test files 0. No new test dependencies, no `withr`.

## Found here, not fixed

- **Duplicate `(x, fill)` rows.** Two rows in one cell draw overlapping rects and a grid has nowhere to put the second value. Both processors detect this and keep the old row-by-row path rather than dropping a row to force a grid. That frame stays ragged — a degenerate chart, not an incomplete one.
- **No single column holding every fill level**, so the stacking order cannot be placed geometrically. Stragglers are appended rather than dropped: the order may be imperfect, but a series absent from the payload can never be announced at all.
- **Audio and braille still read the coerced `0`** for a missing cell — `barValues` is the only numeric channel the frontend has. Text is where the distinction lands, and text is the channel this issue was about.

---


jooyoungseo:
Point 2 is right, and measuring it turned up a narrow **regression** rather than just an uncovered case. Fixed in `1ad3881`. Point 1 (the duplicated grid logic) I have deliberately not done here — reasoning at the bottom.

## What the measurement showed

`emitted` is the announced `x/y/z` triples; `-` is a missing component.

```
                       origin/main                          PR as reviewed
NA in x    (dodge)     3  p/10/A  -/30/A  q/20/B            4  p/10/A q/-/A p/-/B q/20/B
NA in fill (dodge)     2  p/10/A  q/20/B                    6  p/10/A q/-/A r/-/A p/-/B q/20/B r/-/B
NA in x    (stack)     2  p/10/A  -/30/A                    4  p/10/A q/-/A p/-/B q/20/B
```

For `NA in x`, `main` announces the value **30** under a broken category, and the PR as reviewed drops it entirely. That is worse, not merely different — a value the caller supplied stops being reachable. The `NA in fill` rows are the reverse, gaining phantom cells. Neither reading is right: ggplot2 draws that row as its own grey `NA` category, and no path here produces one.

## The fix

Both processors now take the **existing** ragged exit — the same one `anyDuplicated(cell_keys) != 0L` already takes — when either aesthetic carries a real `NA`:

```r
griddable <- anyDuplicated(cell_keys) == 0L &&
  !anyNA(x_values) && !anyNA(fill_values)
```

So this PR does not touch that case at all. After the change, against the same `main`:

```
                       origin/main          this branch
NA in x    (dodge)     3  p/10/A -/30/A q/20/B      identical
NA in fill (dodge)     2  p/10/A q/20/B             identical
NA in x    (stack)     2  p/10/A -/30/A             3  p/10/A -/30/A q/20/B
NA in fill (stack)     1  p/10/A                    2  p/10/A q/20/B
duplicate  (stack)     2  p/1/A  p/2/A              3  p/1/A p/2/A q/3/B
```

The three stacked rows still differ, and that is the *other* fix in this PR doing its job, not the grid path: `stacking_order` now reads off the fullest column, so fill B is no longer dropped from the payload entirely. Strictly more of the user's data announced, never less. Complete grids, gapped grids and `stat = "count"` are all byte-identical to the tables in the PR description.

## Verification

**Bite check:** the two `!anyNA(...)` clauses reverted, tests kept → **8 failures**, all in the new block, across both aesthetics × both positions. Restored → clean.

The new tests come in a pair, because the obvious one is vacuous on its own:

- *a real NA in x or fill keeps the row-by-row reading* — no point carries an `NA` value, and every announced value is one the caller supplied.
- *the NA guard is what holds those frames back, not their shape* — the same frames with the `NA` replaced by a real level are still incomplete, so they **must** reach the grid. Without this, the first test would pass just as well if the grid had stopped working altogether.

**Full suite** with `main` merged in (now including #108): `FAIL 0 | WARN 2 | SKIP 31 | PASS 4601`. The 2 warnings are the pre-existing `boxplot(notch = TRUE)` ones. lintr: dodged **0** (main 0), stacked 5 (main 5, all pre-existing), test file 0.

## Not done: factoring the shared grid logic

You are right that the two blocks are near-identical, and I would rather it were one helper. I am leaving it because the two call sites do not actually agree on their inputs — the dodged processor works from evaluated aesthetic *vectors* (`resolve_aes_values()`), the stacked one from *columns* of `original_data` keyed by name, and the stacked one has to interleave the result with `stacking_order` and `fill_groups`. A shared helper that took both shapes would be a parameter bag, and one that took only the vectors would push the column extraction into the caller and change the stacked processor's structure well beyond this fix. It is worth doing as its own change, where the refactor can be verified as a no-op against these tests rather than landing under a behaviour fix.

## Still open, filed separately

The real question these frames raise — *what should an `NA` category announce?* — is untouched by this PR and now filed as #112. Both readings available today are wrong; picking the right one needs a decision about the wire format, not a guard.

---

PR #113, opened 2026-08-09
fix(base-r): honour positional arguments and the axis grid R draws

### LINKED ISSUE #98: plot(matrix) announces the wrong axis grid, and hist(x, 20) announces the wrong bins
Two verified defects from the #60 leads with the same root shape: an argument-handling path that reimplements what R already resolved. Filed together because the durable fix is one change.

---

## A. `plot(matrix)` — axis grid built from a flattened matrix

**Where:** `R/base_r_point_layer_processor.R:123-127`. Line 123 is `if (is.null(y_data) && is.numeric(x_data)) {`, line 126 is `x_data <- seq_along(y_data)` — the reported line number is exact.

```r
m <- cbind(c(1, 2, 3, 4, 5), c(100, 200, 300, 400, 500))
plot(m)
```

```
EMITTED  x: min 0   max 10  tickStep 2      y: min 0  max 500 tickStep 100
DRAWN    par('usr') = [0.84, 5.16, 84, 516]
         axTicks(1): 1, 2, 3, 4, 5          axTicks(2): 100, 200, 300, 400, 500
         xy.coords -> x: 1..5   y: 100..500
```

The **data** is right — 5 points, column 1 against column 2 — because `extract_data()` uses `grDevices::xy.coords(x, NULL)`. Only `extract_axis_titles()` reimplements the fallback by hand, setting `y_data <- as.vector(m)` (all 10 numbers) and `x_data <- 1:10`, so `pretty(range(1:10))` gives an x grid twice as wide as the plot.

Worse with three columns (`cbind(1:5, seq(100,500,100), rep(-9,5))`, which R still plots as col1-vs-col2):
```
EMITTED  x[0, 16] step 5   y[-100, 500] step 100
DRAWN    x ticks 1..5      y ticks 100..500
```

Control: `plot(c(100,200,300,400,500))` emits `x[1,5] step 1`, `y[100,500] step 100` — exactly `axTicks()`. So this is matrix-specific. `grep -rn "seq_along(y_data)" R/` returns exactly one hit.

*Separate pre-existing gap, not part of this:* both cases emit `label: ""` where R draws `m[,1]`/`m[,2]`; the vector control also emits `""` where R draws `Index`/`v`.

**Severity:** grid navigation and any spoken axis range for a `plot(matrix)` scatter are wrong — every point reported at roughly half its true horizontal position, y floor announced as 0 when the lowest drawn tick is 100. The points sonify correctly, which makes it easy to trust.

---

## B. `hist(x, 20)` — positionally supplied `breaks` silently ignored

**Where:** `R/base_r_histogram_layer_processor.R:52-57`. Root cause upstream at `R/base_r_function_patching.R:766-768`: the wrapper is `function(...)` with no formals, so `list(...)` preserves only names the user typed.

```r
set.seed(1); x <- rnorm(60)
hist(x, 20)          # positional
hist(x, breaks = 20) # named
```

```
hist(x, 20)              emitted nbins= 9
   edges : -2.5 -2 -1.5 -1 -0.5 0 0.5 1 1.5 2
   counts: 1 1 4 7 16 10 14 4 3
hist(x, breaks = 20)     emitted nbins=22

truth hist(x,20)$breaks (n=22): -2.4 -2.2 ... 1.8 2
truth Sturges default   (n=9) : -2.5 -2 ... 1.5 2

POSITIONAL matches truth(breaks=20)?  FALSE
POSITIONAL matches Sturges default ?  TRUE     <- silently reverted
NAMED      matches truth(breaks=20)?  TRUE
```

The drawn SVG really has 22 bars while the payload describes 9. Resolved with a real CSS engine (Chromium `document.querySelectorAll`):

```
"rect[id^='graphics-plot-1-rect-1\\.1']" -> 22
```

and the layer carries 1 selector against 9 data points, so sonification walks 9 phantom bins over a 22-bar picture and the highlight flashes the whole histogram.

**Severity:** low frequency, high damage. `hist(x, 20)` is an ordinary idiom. Every announced number is fabricated relative to what a sighted colleague sees, with no warning — the payload looks internally consistent. The same defect silently drops positionally supplied `freq`, `right` and `include.lowest`.

---

## Fix sketch, covering both

Do the argument matching once, centrally. In the wrapper at `base_r_function_patching.R` (~line 766), run the captured call through `match.call(definition = ORIG, call = this_call, expand.dots = TRUE)` — or match `args_list` against `formals(ORIG)` — and store the **named** list. Every processor doing `args[["breaks"]]` / `args[["x"]]` then works for positional calls for free, and the hand-rolled "find the first unnamed argument" fallbacks in `resolve_xy_args()` (`R/base_r_arg_utils.R`) and in the histogram and heatmap `extract_data()` become redundant. Match against `graphics:::hist.default` for the generic/method split, and keep the current behaviour of not forcing NSE promises.

For **A** specifically, delete the hand-rolled fallback at `base_r_point_layer_processor.R:123-127` and reuse the resolution `extract_data()` already trusts: `grDevices::xy.coords(x_data, y_data)` inside a `tryCatch`, taking `coords$x` / `coords$y`. That collapses two code paths onto one source of truth and handles matrices, data frames, `ts` and list inputs. `xy.coords(..., setLab = TRUE)` also yields the `xlab`/`ylab` R would draw, which closes the empty-label gap noted above.

jooyoungseo:
Closes #98.

Both defects in that issue share one root cause — a path that re-derives what R already resolved — so this is one change with one central fix each.

**A.** Every recorded call is matched against the definition R dispatched to (`hist.default()`, not the `hist` generic) before the processors read it, so a positional argument arrives under the name R gave it.

**B.** `extract_axis_titles()` stops re-deriving the single-argument fallback by hand and asks `grDevices::xy.coords()`, the same resolution `extract_data()` already trusts.

## Before / after, measured against what R drew

`axTicks()` on the live device is the truth column.

```
                    origin/main                     this branch          drawn
2-col matrix        x [0,10] s2   y [0,500] s100    x [1,5] s1  y [100,500] s100   x 1..5  y 100..500
3-col matrix        x [0,16] s5   y [-100,500]      x [1,5] s1  y [100,500] s100   x 1..5  y 100..500
data frame          x   (none)    y   (none)        x [1,5] s1  y [100,500] s100   x 1..5  y 100..500
ts(start = 2001)    x [1,5] s1    y [10,50] s10     x [2001,2005] s1  y [10,50]    x 2001..2005
plain vector        x [1,5] s1    y [100,500]       unchanged                      x 1..5  y 100..500
```

`plot(matrix)` announced a grid **twice as wide** as the axis, so every point sat at roughly half its true horizontal position and the y floor was announced as 0 where the lowest drawn tick is 100. The data was always right, which is what made it easy to trust. The `ts` and data-frame rows were not in the issue; they fall out of using `xy.coords()` and are checked against `axTicks()` here rather than assumed.

```
                       emitted bins    drawn bars
hist(x, 20)   before         9             22
hist(x, 20)   after         22             22
hist(x, breaks = 20)        22             22      (unchanged, both sides)
```

### A positional `type=` had never reached the type detection at all

`detect_layer_type()` has read `args$type` for a long time, but `plot(x, y, "l")` never delivered one, so every positional type was read as the default points. Resolved against the exported SVG with a real CSS engine (Chromium `document.querySelectorAll`), "does the layer's selector match anything":

```
              main positional      main type =         this branch (both spellings)
plot type=p   point  5 nodes       point  5 nodes      point  5 nodes
plot type=l   point  0 nodes  ✗    line   1 node       line   1 node
plot type=b   point  5 nodes       line   NO SELECTOR ✗    point  5 nodes
plot type=o   point  5 nodes       line   1 node       line   1 node
plot type=c   point  0 nodes  ✗    line   NO SELECTOR ✗    line   NO SELECTOR
plot type=h/s/S  point 0 nodes ✗   line   NO SELECTOR ✗    line   NO SELECTOR
```

`type = "l"` was **broken on `main` in the positional spelling** — a point selector against a picture with no point glyphs, so the highlight was dead. That is fixed.

But naming the argument also routed `"b"` into a branch that was wrong for it, and there it *lost* a working highlight. R draws `"b"` with a gap at every symbol, so gridSVG exports the segments as `brokenline`, not the single `lines` polyline the line selector addresses. The `curve` branch a few lines below in the same function had **already reached exactly this conclusion** and listed `"b"` among the types that do not qualify; the `plot` branch could not act on it while a positional type never arrived. `"b"` now reads as the symbols it draws, in **both** spellings — so this branch is strictly better than `main` on the named form too, not just at parity.

`"c"`, `"h"`, `"s"` and `"S"` are deliberately left alone: they export under names the line selector cannot address either, but unlike `"b"` they draw no symbols to fall back to, so there is nothing to point at until a trace type exists for them (`"s"`/`"S"` is what #49 is for). Their reading is unchanged from a named `type =` today.

### `options(maidr.auto_show = FALSE)` disables recording entirely

Not a change here, and not something I would have guessed: with that option set, no Base R call is recorded at all, so `save_html()` reports "No Base R plots detected". It cost me two wrong measurements before I noticed. Worth knowing; not touched, because whether that is intended is a separate question.

### `plot(formula)` emits zero data points

Unchanged by this branch — `plot(mpg ~ wt, data = mtcars)` announces a `point` layer with **0** points on both sides. Out of scope here, and the replay path is separately covered below.

## What keeps the central fix safe

`match.call()` reorders arguments into formal order. This does not: the recorded list keeps the caller's order and only **gains names**, because `apply_barplot_sorting()` and friends still read the height out of slot 1 while replay uses `do.call()`, which honours names regardless of position.

**The dispatch argument is deliberately left unnamed.** S3 dispatch happens on the first argument of the *generic*, and methods rename it — `plot.formula()` calls it `formula`, the generic calls it `x`. Naming it breaks replay in one direction or the other. There is a test that asserts the name stays empty, and another that actually replays a recorded `plot(y ~ x)` and `boxplot(y ~ x)` end to end.

The NSE branch of the wrapper is untouched: it holds unevaluated expressions, and resolving the dispatched method would have to force them.

## Verification

**Bite checks**, one per fix, tests kept:

| reverted | result |
|---|---|
| `match_recorded_args()` call removed from the wrapper | **10 failures** |
| the `xy.coords()` fallback restored to the hand-rolled one | **9 failures** |
| `"b"` mapped back to `line` | **6 failures** |
| all three in place | **0** |

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4504` with `main` merged in (including #108). The 2 warnings are the pre-existing `boxplot(notch = TRUE)` ones.

**Selector resolution** is from Chromium against the real exported SVG, not from reading ids — the table above is the output of that sweep. The regression sweep also covered `barplot`, `barplot(matrix, beside = TRUE)`, `boxplot`, `heatmap`, `lines()`/`points()` overlays and `plot(x, y, xlim = )`: all byte-identical between the two sides.

roxygen: `roxygenise()` produces no diff; `man/` gains the three new internal helper pages and `NAMESPACE` is unchanged. lintr: `base_r_adapter.R` 1 (main 1, pre-existing), `base_r_arg_utils.R` 0, `base_r_point_layer_processor.R` 0, test file 0.

`base_r_function_patching.R` reports **one new lintr warning** — "no visible global function definition for `match_recorded_args`". It is a per-file analysis artefact: lintr cannot see across files, `match_recorded_args()` lives in `base_r_arg_utils.R`, and the other call site in the same file is not flagged only because it sits inside `eval(substitute(...))`. `codetools::checkUsagePackage("maidr")` — the same machinery `R CMD check` runs, but with the package loaded — reports **0** notes, and the `create_barplot_wrapper()` path is exercised by a passing test.

## Not done

- `xy.coords(setLab = TRUE)` would also supply the `xlab`/`ylab` R draws, closing the empty-label gap the issue notes in passing. Left out: it changes the announced labels for every scatter plot in the package, which wants its own before/after rather than riding along here.
- The duplicate "find the first unnamed argument" fallbacks the issue predicts will become redundant are still in place. They are now belt-and-braces rather than the only resolution, and removing them is a cleanup that should not share a commit with a behaviour fix.

---


jooyoungseo:
Two corrections to the description, both mine, plus the re-measurement after merging `main`.

## The roxygen check in the description was worthless

I wrote "`roxygenise()` produces no diff". I had been invoking it as `roxygen2::roxygenise(load_code = "source")`, which on this package **aborts**:

```
! R6 class <BaseRAdapter> lacks source references.
```

It leaves `man/` untouched and exits, so "no diff" meant "it never ran", not "the docs are in sync". Plain `roxygen2::roxygenise()` works.

Re-checked properly. It regenerates three pages that drift with **no source edit at all** — `BaseRAdapter.Rd`, `maidr-package.Rd`, `dot-maidr_chartseries_ta_warned.Rd`, the long-standing set — and nothing else. Those are reverted rather than carried, per the established convention. The three helper pages this PR adds are correct and complete; no page is missing and none is stale. `NAMESPACE` unchanged.

## `main` moved twice under this branch

#108 and then #104 (pie) + #110 landed while this was open. `main` is merged in, and `R/base_r_adapter.R` needed a real merge because #104 touched the same `detect_layer_type()` switch — resolved as both changes, its `coord_polar` work alongside the `type` branch here.

**Re-measured after that merge:**

| | |
|---|---|
| full suite, this branch | `FAIL 0 \| WARN 2 \| SKIP 31 \| PASS 4924` |
| `main` baseline at `bcdbeea`, throwaway worktree | `FAIL 0 \| WARN 2 \| SKIP 31 \| PASS 4832` |

The 2 warnings are the pre-existing `boxplot(notch = TRUE)` ones. The earlier figure of 4504 in the description was against the pre-#104 tree; it is superseded, not contradicted.

**Re-ran the regression sweep against the merged tree**, including #104's new `pie()` path, since naming the recorded arguments touches every Base R call:

```
                     main bcdbeea          this branch
pie(c(A=3,B=5,C=2))  pie / 6 points        pie / 6 points
pie(c(3,5,2), lbls)  pie / 6 points        pie / 6 points
```

Unchanged, as are `barplot`, `barplot(matrix, beside = TRUE)`, `boxplot`, `heatmap`, `lines()`/`points()` overlays, `plot(x, y, xlim = )` and `mosaicplot`'s static fallback. The only rows that move are the ones this PR is about.

`plot(x, y, "b")` also re-checked post-merge: `point / 10 points`, identical to `main`, so the carve-out survives #104.

---

Issue #99, opened 2026-08-08
heatmap() of an unnamed matrix announces positional labels instead of the row/column indices it draws


jooyoungseo:
Verified by rendering and by reading the drawn labels out of the exported SVG with a real browser. From the #60 Medium leads.

**Correction to the lead as filed:** #60 said the labels "no longer match", implying a regression. That framing is wrong — see below.

**Where:** `R/base_r_heatmap_layer_processor.R:90-95`, the `is.null(row_names)` / `is.null(col_names)` fallback.

## Reproduce

```r
m <- matrix(c(10,11,12,13, 90,91,92,93, 50,51,52,53, 20,21,22,23, 70,71,72,73),
            nrow = 5, ncol = 4, byrow = TRUE)   # deliberately UNNAMED
heatmap(m)
```

## Measured

```
EMITTED x (cols, left->right): 1,2,3,4
EMITTED y (rows, top->bottom): 5,4,3,2,1
EMITTED points grid (top->bottom):
   90 91 92 93
   70 71 72 73
   50 51 52 53
   20 21 22 23
   10 11 12 13

heatmap() rowInd: 1,4,3,5,2   colInd: 1,2,3,4
SVG right-axis labels TOP->BOTTOM (sorted by measured getBoundingClientRect): 2,5,3,4,1
```

The **values** are correct — `rowInd = 1,4,3,5,2` reversed for top-down gives rows 2,5,3,4,1, exactly what was emitted. The **labels** are pure positions `5,4,3,2,1` instead of the original row indices `2,5,3,4,1` that `heatmap()` prints. Only position 3 coincidentally agrees.

With the columns also reordered, both axes disagree:
```
EMITTED x: 1,2,3,4      SVG bottom-axis labels LEFT->RIGHT: 1,4,3,2
EMITTED y: 5,4,3,2,1    SVG right-axis labels TOP->BOTTOM:  4,3,5,2,1
```

Control with `rownames(m) <- paste0("R",1:5); colnames(m) <- paste0("C",1:4)`: emitted `y: R2,R5,R3,R4,R1`, `x: C1,C2,C3,C4` — correct, because the reordered matrix carries its dimnames along. So this is exclusively the unnamed fallback.

## Why the "regression" framing is wrong

[...]

PR #115, opened 2026-08-09
fix(base-r): announce the axis labels a heatmap() caller supplied

### LINKED ISSUE #106: heatmap()'s labRow= and labCol= arguments are ignored
Found while fixing #99, and left out of that change because it is a different input.

maidr reads a heatmap's axis labels from `dimnames()`, and — after #99 — from the dendrogram ordering when there are no dimnames. It never reads the caller's own `labRow=` / `labCol=` vectors, so an explicit

```r
heatmap(m, labRow = c("alpha", "beta", "gamma", "delta", "epsilon"))
```

draws those strings on the axis and announces something else.

`stats::heatmap()` resolves labels in this order:

```r
labRow <- labRow[rowInd] %||% rownames(x) %||% (1L:nr)[rowInd]
```

so an explicit `labRow` wins over dimnames, and it is subscripted by `rowInd` like everything else. #99 implemented the second and third arms of that expression; the first is still missing.

**Not yet measured.** Unlike the other issues filed today, this one has not been reproduced end to end — it was inferred from reading `stats::heatmap`'s source next to the processor while fixing #99. Anyone picking it up should confirm the rendered SVG against the emitted payload first, the same way #99 was verified: Chromium, `svg text` sorted by measured `getBoundingClientRect()`, plot region derived from the `image-rect` cells.

## Fix sketch

Same shape as #99. The recorded call already carries `labRow`/`labCol` when the user passed them, so the fallback chain becomes: caller's labels subscripted by the ordering → dimnames → `ordering$rowInd` → `seq_len()`. Note this interacts with the positional-argument work in #98: if `labRow` was supplied positionally it may not be visible under its name today, so check that first.

jooyoungseo:
Closes #106.

The issue says it was inferred from reading `stats::heatmap`'s source and **not measured**. It is now, end to end, and it reproduces exactly as described.

## Reproduced

`stats::heatmap` resolves each axis in one expression, read out of the installed function rather than from memory:

```r
labRow <- labRow[rowInd] %||% rownames(x) %||% (1L:nr)[rowInd]
labCol <- labCol[colInd] %||% colnames(x) %||% (1L:nc)[colInd]
```

The caller's labels come first and beat dimnames, and they carry the same subscript as the data. #99 implemented the second and third arms; the first was missing.

Drawn strings are the `<text>` nodes in the real export, not what the source says it would draw:

```
                                    DRAWN (rows)                    ANNOUNCED (rows)
heatmap(m, labRow=, labCol=)        alpha gamma delta epsilon beta  2 5 4 3 1        ✗
heatmap(m + dimnames, labRow=UPPER) ALPHA GAMMA DELTA EPSILON BETA  beta ... alpha   ✗
heatmap(m, Rowv=NA, Colv=NA, ...)   alpha beta gamma delta epsilon  5 4 3 2 1        ✗
```

So an unnamed matrix announced bare indices beside the caller's words, and a matrix that *also* carried dimnames announced the dimnames while the axis showed the caller's strings. All three are correct after the change (announced order is the reverse of drawn throughout — see below).

## The announced order is deliberately the reverse of the drawn one

`heatmap()` puts reordered row 1 at the bottom of the image while the payload's row 0 is the top, so the y labels come out reversed relative to the glyph order. That convention predates this change and #99's tests pin it; the new tests `rev()` before comparing against the drawn labels rather than quietly matching a different order.

## Deliberately mirrored, not rejected

A short or `NA`-carrying `labRow` gives `NA` for the positions it cannot fill, because `labRow[rowInd]` does. `axis()` draws that as the glyphs "NA", so mirroring keeps the announcement equal to the picture rather than substituting a guess. Unit-tested.

## On the #98 interaction the issue flags

Checked: `labRow`/`labCol` are the 17th and 18th formals of `heatmap()`, so nobody supplies them positionally, and the recorded call already carries them under their names when passed as `labRow =`. This does not depend on #113, and #113 does not change it.

## Verification

**Bite check:** the new arm removed, tests kept → **5 failures**. Restored → clean. The `heatmap_caller_labels()` unit block passes both ways by design, since it tests the helper rather than its call site.

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4858`, against a `main` baseline of **4832** measured in a throwaway worktree at `bcdbeea`. +26 is exactly the new file. The 2 warnings are the pre-existing `boxplot(notch = TRUE)` ones.

The unchanged cases are pinned too, so the new first arm cannot displace #99's work: a call with no `labRow` keeps the ordering-derived indices, a named matrix keeps its dimnames, and `image()` — which has no such argument and never reorders — is untouched.

lintr 0 on both files (main 0). `tools::checkRd()` clean on the new page.

## A process note on roxygen

`roxygen2::roxygenise(load_code = "source")` — which I had been using — **aborts** on this package with "R6 class `<BaseRAdapter>` lacks source references" and silently leaves `man/` untouched, so a "no diff" result from it means nothing. The plain `roxygen2::roxygenise()` works. That is why the new `man/heatmap_caller_labels.Rd` is here and why I re-checked #113's `man/` the same way. roxygen2 is 7.3.2, matching `RoxygenNote`; the three perennially-stale pages (`BaseRAdapter.Rd`, `maidr-package.Rd`, `dot-maidr_chartseries_ta_warned.Rd`) that drift with no source edit are reverted rather than carried.

Also worth knowing, since it cost me two wrong measurements: `%` in a roxygen block must be escaped as `\%`, or the `.Rd` silently comes out wrong.

---


jooyoungseo:
Acted on the edge case, and the fix turned out to be a closer match to `heatmap()` than what I had.

## The observation was right

When `compute_heatmap_ordering()` came back `NULL`, `heatmap_caller_labels(labels, NULL)` returned the caller's vector **as written**. A mismatched-length `labRow` would then produce a label array whose length disagreed with `points`.

## The fix, and why it is not just a length guard

`heatmap()` applies a subscript on *both* paths — when there is no clustering, `rowInd` is simply `1L:nr`. My `NULL` branch was skipping the subscript entirely, which is the part that was wrong; the length mismatch was the symptom.

The caller now passes the identity index instead of `NULL`:

```r
row_order <- if (is.null(ordering)) seq_len(nrow(heat_matrix)) else ordering$rowInd
```

and the helper's `NULL` branch is gone, so there is one path rather than two. Subscripting is what holds the result to one label per row in both directions: too few labels pad with `NA` (what `axis()` draws, already covered), too many now get the surplus dropped — which is what `heatmap()` does too.

Two assertions added for the direction that was untested:

```r
expect_length(heatmap_caller_labels(c("a", "b", "c", "d"), seq_len(2)), 2)
expect_length(heatmap_caller_labels(c("a", "b"), seq_len(4)), 4)
```

## Re-verified after the change and after merging `main` (#114)

**Bite check:** the arm removed, tests kept → **5 failures**, unchanged. Restored → clean.

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4890`. The baseline moved with #114, so the delta is what matters and it is still exactly this file. lintr 0 on both files; `man/heatmap_caller_labels.Rd` regenerated for the changed `@param`.

## On the other two points

The reordering asymmetry you traced — dimnames read off the already-permuted matrix, `labRow` subscripted live — is exactly the reason only the first and third arms of `heatmap()`'s expression carry a subscript, and it is why the code looks lopsided. Good catch confirming it rather than flagging it as a bug.

You noted you could not run R here. For the record, the `lintr 0` and `checkRd()` results were re-run after this change, not carried over.

---

PR #116, opened 2026-08-09
docs: stop roxygen scattering R6 method titles into \keyword{} entries

### LINKED ISSUE #109: Add @description to R6 method docblocks to stop roxygen scattering sentences into \keyword{}
Follow-up to #104, which fixed this for two files. The same root cause affects fifteen more.

## The bug

An R6 method docblock that opens with **untagged title text** instead of `@description` makes roxygen2 7.3.2 do two wrong things:

1. It emits `Undocumented R6 method` and scatters the title sentence's **individual words** into the *class-level* block as `\keyword{}` entries. `man/BaseRAdapter.Rd` had accumulated `\keyword{Check}`, `\keyword{a}`, `\keyword{can}`, `\keyword{handle}`, `\keyword{the}`, `\keyword{this}` — words lifted straight out of "Check if this adapter can handle a plot object".
2. A title following an already-tagged block gets glued onto the **previous** `@return`, so `can_handle`'s Returns section ends with the next method's description.

`R CMD check` NOTEs non-standard `\keyword` entries. Adding `@description` makes both symptoms vanish — verified in a minimal throwaway package.

## Remaining files

`#104` fixed `R/ggplot2_adapter.R` and `R/base_r_adapter.R` only, to keep that PR's diff scoped. Still polluted:

- `BaseRLineLayerProcessor`
- `BaseRPlotOrchestrator`
- `BaseRProcessorFactory`
- `Ggplot2BoxplotLayerProcessor`
- `Ggplot2LineLayerProcessor`
- `Ggplot2PlotOrchestrator`
- `Ggplot2PointLayerProcessor`
- `Ggplot2ProcessorFactory`
- `PlotSystemRegistry`
- `ProcessorFactory`
- `dot-base_r_function_classes`
- `get_device_state`
- `group_device_calls`
- `process_faceted_plot_data`
- `process_patchwork_plot_data`

The work is mechanical — add `@description` to each R6 method block in the corresponding source file, then regenerate.

## Regeneration note

`devtools` does not install in a minimal container (`textshaping`/`ragg` need harfbuzz headers). Use:

```r
roxygen2::roxygenise(".", load_code = roxygen2::load_pkgload)
```

with **roxygen2 pinned to 7.3.2** to match `DESCRIPTION`'s `RoxygenNote` — the current CRAN roxygen2 rewrites all 246 `.Rd` files. Never hand-edit `man/` or `NAMESPACE`; check with `git diff --stat man/` that only the intended files moved.

jooyoungseo:
Closes #109.

Documentation only. Full suite lands on **exactly** the `main` baseline, which is the point: `FAIL 0 | WARN 2 | SKIP 31 | PASS 4832` on both, measured in a throwaway worktree at `bcdbeea`.

## The issue's diagnosis was right for ten of the fifteen files, and wrong for five

The issue attributes all fifteen to R6 method blocks opening with an untagged title. That holds for the ten R6 classes. It does not explain `get_device_state`, `group_device_calls`, `process_faceted_plot_data`, `process_patchwork_plot_data` or `dot-base_r_function_classes`, which are plain functions and objects with no methods at all.

Those five have a **second, separate cause**: a file-level roxygen block with no `@name` and no terminator.

```r
#' Base R State Tracking
#'
#' This module tracks the plotting state for each graphics device, ...
#'
#' @keywords internal
                          <- nothing here
#' Get or Initialize Device State
```

roxygen cannot attach the header to an object, so it folds into the block below and `get_device_state`'s own title and description come back out as `\keyword{Get}`, `\keyword{Initialize}`, `\keyword{Retrieves}`, `\keyword{the}`, `\keyword{a}` … Adding a `NULL` terminator fixes it, verified one file at a time before touching the rest.

That is the pattern `globals.R`, `svg_utils.R`, `rdp_utils.R`, `shiny.R`, `fallback_rendering.R` and `base_r_arg_utils.R` already use.

## Verified, not assumed

Both mechanisms were confirmed by isolating one file each and regenerating before doing the sweep:

| | before | after |
|---|---|---|
| `get_device_state.Rd` — add `NULL` only | 14 bogus keywords | 1 (`internal`) |
| `Ggplot2PointLayerProcessor.Rd` — add `@description` only | 4 bogus | 1 |

Whole-package audit after the sweep — every `\keyword{}` entry in `man/` that is not `internal`, `datasets` or `package`:

```
before:  15 files, 176 bogus entries
after:   none
```

`tools::checkRd()` is clean on all 17 changed pages. `NAMESPACE` and `DESCRIPTION` are untouched. lintr is byte-identical to `main` on all 17 source files.

## A second symptom, also gone

The issue's point 2: a title following an already-tagged block gets glued onto the **previous** `@return`. Visible in the diff — `Ggplot2PointLayerProcessor$process()`'s Returns section ended with the *next* method's entire description, and that text has moved to where it belongs:

```diff
 \subsection{Returns}{
 List with data and selectors
-Extract axis information from the plot
-
-Returns per-axis objects with label and optional grid navigation fields
-(min, max, tickStep). Grid fields are only included when they can be
-successfully extracted from the built plot scales.
 }
```

## One thing beyond the issue, and why

Tagging the method blocks alone leaves each class page reading:

```
\description{
Point Layer Processor

Point Layer Processor
}
\details{
Processes scatter plot layers (geom_point) to ...
}
```

— the title twice, with the real description demoted to details. That is the shape `man/BaseRAdapter.Rd` has carried since #104, so I would have been *introducing* it into fifteen more pages. An explicit `@description` on each class block collapses it back to one clean description with no `\details{}`. One line per file, and the two pages #104 already fixed get it too, so all seventeen now render the same way.

## Stated plainly

The five newly-terminated files now emit roxygen's `Block must have a @name` notice, joining the six that already do. It is roxygen's own message, not `R CMD check`'s, and it is what the pattern costs — the alternative is five `.Rd` pages documenting a module header that describes no object. `R CMD check`'s actual complaint, the non-standard `\keyword` entries, is what this removes.

roxygen2 is 7.3.2, matching `RoxygenNote`. Worth recording since it wasted time elsewhere today: `roxygenise(load_code = "source")` **aborts** on this package with "R6 class `<BaseRAdapter>` lacks source references" and leaves `man/` untouched, so a clean `git status` after it means nothing. Plain `roxygen2::roxygenise()` is what the issue prescribes and what was used here.

---


jooyoungseo:
`main` moved under this branch (#114, then #115), so it is merged in and the suite re-measured.

| | |
|---|---|
| this branch | `FAIL 0 \| WARN 2 \| SKIP 31 \| PASS 4862` |
| `main` baseline at `26001ce` (= `bcdbeea` + #114's 30 tests) | 4862 |

Still exactly the baseline, which is the assertion this PR rests on: it changes documentation and nothing else. The 4832 in the description was against `bcdbeea`; both numbers describe the same zero delta, just from different starting points.

The whole-package keyword audit was re-run after the merge — no non-standard `\keyword{}` entry remains anywhere in `man/`.

---

PR #117, opened 2026-08-09
fix(base-r): scope the original-function lookup to the namespaces it names

### LINKED ISSUE #111: `.maidr_quantmod_onload_hook` is vestigial for wrapping but still load-bearing for replay
Follow-up promised in #108. Not a user-visible defect — a cleanup whose safety I could not establish inside that PR's scope, so it is recorded rather than done.

## What the hook does today

`.maidr_quantmod_onload_hook` fires on quantmod's `onLoad` and calls `wrap_function()` for quantmod's graphics entry points. Post-sealing, that call is a **no-op in both directions**:

- maidr's namespace is sealed by the time the hook runs, so the wrapper cannot be rebound there.
- `package:maidr` is locked, so the attached copy cannot be rebound either.

So the hook no longer wraps anything, and #108 stopped relying on it for the masking diagnostics — those key off search-path order instead.

## Why it was not deleted

`wrap_function()` has a second effect the hook still reaches: it populates `.saved_graphics_fns`. That table is what `save_html()`'s replay resolves the original function through. Deleting the hook removes entries from it, and I did not measure whether any replay path still reads a quantmod entry from there.

Concretely, before this can be removed:

1. Enumerate which keys the hook contributes to `.saved_graphics_fns` that no other call site writes.
2. For each, check whether `save_html()`'s replay resolves through `.saved_graphics_fns` or through `getExportedValue()`/`::`.
3. Confirm with a real `R CMD INSTALL` into a private library — the sealing behaviour is not reproducible under `pkgload::load_all()`, which is what made this hard to settle inside #108.

If step 2 shows every key has another writer, the hook and its registration can go. If not, the hook should be narrowed to *only* the `.saved_graphics_fns` population, with the dead `wrap_function()` rebinding removed and a comment saying why the rest stays.

## Priority

Low. Nothing is broken; this is dead code that reads as live and will mislead the next person to touch the quantmod path.

jooyoungseo:
Closes #111.

**That issue asked the wrong question, and I filed it.** It asks whether `.maidr_quantmod_onload_hook` can be deleted as vestigial. Investigating it turned up a live defect in a different function, and once that is fixed the hook is neither vestigial nor a problem. So this PR fixes the defect and leaves the hook alone.

## What is actually wrong

`find_original_function()` probed `graphics`, `stats` and `grDevices` with `get(name, envir = asNamespace(...))` — R's default `inherits = TRUE`. **A namespace environment's parent chain ends at the search path.** So the lookup does not stop at `graphics`; it walks out into whatever is attached — including `package:maidr`, which exports wrappers under the very names being looked up.

`chartSeries()` hits this. maidr's quantmod hook fires while quantmod is *loaded* but not yet *attached*, so at that instant `package:maidr` is ahead of quantmod on the path and the probe finds **maidr's own recording stub**. That value is then stored as the "original", and `base_r_wrapper_exports.R`'s correct write is guarded by `if (is.null(...))`, so it never lands.

## Measured, on a real `R CMD INSTALL` into a private library

```
                                    main                      this branch
.saved_graphics_fns[["chartSeries"]]
  after library(quantmod)           MAIDR'S OWN STUB          quantmod's own
  after chartSeries()               MAIDR'S OWN STUB          quantmod's own
  after save_html()                 MAIDR'S OWN STUB          quantmod's own
get_original_function("chartSeries") MAIDR'S OWN STUB         quantmod's own
```

`replay_plot_call()` does `do.call(orig_fn, args)`, so on `main` the replay calls **maidr's own wrapper**, which forwards through `...`. That is precisely the corruption #108 documented: `TA = NULL` arrives as a symbol and quantmod dies on `chob@passed.args$TA`. Counting real `quantmod::chartSeries` entries during one `save_html()`:

```
main         2   <- first call fails on the corrupted TA, retry rescues it
this branch  1
```

So the only reason nothing is visibly broken today is the retry added in #108 for exactly this failure. **The exported payload is byte-identical either way** — verified across three builds, with the session-unique plot ids normalised:

```
main (hook present)            2cd90d071b9dd1c42129fa1042c735bd
hook registration deleted      2cd90d071b9dd1c42129fa1042c735bd
this branch (hook kept)        2cd90d071b9dd1c42129fa1042c735bd
```

## The fix

The probes are scoped with `inherits = FALSE`, and **`base` joins the chain** because `plot()` lives there rather than in `graphics` (since R 4.0) — an inherited lookup was what found it before. Verified that all 35 wrapped names still resolve from `{graphics, stats, grDevices, base, quantmod}` with the narrowed lookup; the wrap list is a fixed literal in `base_r_function_classification.R`, so that set is closed.

## So: can the hook go?

**Not needed either way, and left alone.** With the lookup fixed the hook does something correct — it pre-populates the right original at attach time — and deleting it is also harmless, because the stub writes the same value on first call (measured: the "hook registration deleted" build above). Since both are now equivalent in behaviour, removing working code does not belong in the same change as a correctness fix. If the maintainers want it gone for the CRAN-cleanliness reason in #111, that is a clean one-hunk follow-up with this measurement already in hand.

What #111 got right: the hook's `wrap_function()` call *is* a no-op for rebinding post-sealing, and its only live effect *was* the `.saved_graphics_fns` write. What it got wrong is that this write was the bug rather than the load-bearing part.

## Verification

**Bite checks**, tests kept:

| reverted | result |
|---|---|
| `inherits = FALSE` back to the default | **1 failure** |
| `base` removed from the chain | **1 failure** |
| both in place | **0** |

One failure each, not more, because the in-process test session cannot reproduce the attach-order race that makes `chartSeries` resolve wrongly — that is only reachable out of process, in the `R CMD INSTALL` transcripts above. The test that does fire is the synthetic one, which is the invariant itself: a name bound only in the global environment must **not** be found by a probe into `graphics`.

**Full suite:** `FAIL 0 | WARN 2 | SKIP 31 | PASS 4936`, against a baseline of **4862** at this branch's base. The new file accounts for **72** of the +74; the remaining **2 I could not account for** — the two files that touch this machinery (`test-function-patching.R`, 52 both sides) are unchanged, and skips are 31 on both. Recording it rather than rounding it away.

lintr unchanged from `main` (1 pre-existing in the source file, 0 in the test). roxygen produces no `man/` diff — `find_original_function()` already had a page and its signature is unchanged.

---


jooyoungseo:
Resolved the +2 the description left unexplained, and it is not a discrepancy at all — my arithmetic was comparing two things that are not comparable.

Per-file pass counts, both trees measured in paired runs on the same machine:

```
branch total 5056   main total 4982

test-original-function-lookup.R    branch=74   main=0   delta=+74
(no other file differs)
```

**The whole +74 is the new file.** The "72" in the description came from a *filtered* run (`test_local(filter = "original-function-lookup")`), and two of the sweep tests' assertions only exist in a full-suite run: they iterate `names(.saved_graphics_fns)`, and that set is larger once other test files have driven more wrapping. Nothing else in the suite moves, which is the property that actually matters here.

That is the same implicit dependency you flagged as a non-blocking note — the sweeps' coverage is tied to what the session has wrapped. Your read was right, and it turns out to be observable: the sweep genuinely covers two more names in a full run than in a filtered one. I have left the tests iterating that set rather than hard-coding a list, because covering *whatever is wrapped* is the invariant I want, and a hard-coded list would silently stop covering a name added later.

Also applied your comment nit — the eight-line `inherits = FALSE` rationale now sits above the loop rather than in its body. Same text, one level up.

**Re-measured after that and after merging `main` (#113, #115, #116):**

| | |
|---|---|
| this branch | `FAIL 0 \| WARN 2 \| SKIP 31 \| PASS 5056` |
| `main` baseline at `2990b3f` | `FAIL 0 \| WARN 2 \| SKIP 31 \| PASS 4982` |

lintr is 2 on both sides of the same file — identical, and both pre-existing (the `match_recorded_args` one arrived with #113 and is the cross-file analysis artefact documented there).

---

PR #118, opened 2026-08-09
fix(pie): keep the sign of a negative slice, and name the axes Base R leaves unnamed

jooyoungseo:
Two accessibility defects, both found by rendering r-maidr's own output in a browser rather than by reading the payload.

## 1. A negative pie slice was announced as positive

A wedge's magnitude is its own extent, `ymax - ymin`. That is unsigned, and ggplot2 stacks a negative datum **below** the baseline:

```
v = c(60, -40, 30)   →   ymin   ymax
  A                        30     90     value  60
  B                       -40      0     value -40
  C                         0     30     value  30
```

So `ymax - ymin` for B is `0 - (-40)` = **40**. The emitted payload was `[{A,60},{B,40},{C,30}]` — the author wrote `-40` and MAIDR announced `40`, computing its share against a total of 130.

xability/maidr#772 already made the renderer treat a negative slice as a gap, so it cannot corrupt every other slice's percentage. **That defence never fired here**: laundering the sign in the producer left it nothing to catch.

Verified end to end in Chromium.

Before:
```
Fruit is B, Units is 40, Percentage is 30.8%
```

After:
```
Fruit is A, Units is 60, Percentage is 66.7%
Fruit is B, Units is missing, Percentage is missing
Fruit is C, Units is 30, Percentage is 33.3%
```

The two real shares now sum to 100% against 90, rather than being diluted by a value that should never have counted.

The sign is restored from which side of the baseline the segment sits on, testing `ymax <= 0 && ymin < 0` rather than `ymin < 0` alone — so a zero slice (`ymin == ymax == 0`) is not wrongly negated, and a segment straddling zero is read as positive rather than having its sign guessed.

Whether a producer should *reject* a negative outright is a separate question, open at xability/maidr#771. This change pre-empts none of it — no answer to that question is served by destroying the sign first.

## 2. Every Base R chart announced its points with the nouns missing

Heard by a screen reader, before:

```
 is Apples,  is 30, Percentage is 30.0%
 is A,  is 3
```

The processors emitted `label: ""` whenever the author passed no `xlab`/`ylab` — the common case — and an empty string is not `NULL`, so the renderer's own default never applied. (The renderer-side floor for that is xability/maidr#773; this is the producer-side half, which makes the labels *specific* rather than merely present.)

Each processor now says what its own call actually tells it, and **leaves the label absent rather than empty** where it cannot do better, so the renderer's default applies rather than being defeated.

**Two premises I gave the implementation turned out to be wrong, and the evidence was followed instead:**

- `hist()` and `boxplot.formula()` *do* carry defaults — `hist.default` sets `xlab = xname`, `plot.histogram` sets `ylab`, `boxplot.formula` derives both from the formula. What is missing is not the title but the **record** of it: maidr's wrappers record evaluated argument values, so R's internally-computed defaults never reach a processor. Where R computes a title from data maidr still holds, R's own title is reproduced rather than a noun invented.
- A histogram's y is **"Frequency"/"Density"**, not "Count". `hist(freq = FALSE)` and non-equidistant breaks plot densities, and the extractor emits densities there — so "Count" would have been the wrong noun in exactly the way this change exists to avoid.

## Testing

`testthat::test_local()` → **FAIL 0 | WARN 2 | SKIP 98 | PASS 4357** (baseline on a clean tree: 4282). The 2 warnings are the pre-existing `boxplot(notch = TRUE)` ones.

New `tests/testthat/test-base-r-axis-labels.R` — 24 tests, 63 assertions — goes through a full render and reads the `maidr-data` payload rather than asserting on an intermediate. Four new tests in the pie processor file cover the sign: negative kept, all-positive unchanged, zero not negated, and the negative reaching the emitted HTML rather than only `extract_data`.

An intermediate state (code changed, tests not yet updated) failed 21 assertions of the old `label == ""` contract across the pie, point, line, smooth, heatmap and unknown files; each was rewritten to state the new contract rather than deleted.

`man/` regenerated with roxygen2 7.3.2 via `roxygenise(load_code = load_pkgload)`; `NAMESPACE` unchanged, all new functions internal, no `.Rd` hand-edited.

## Note

The browser verification used a bundle built from maidr `main`, since the npm release carrying pie has not shipped yet. Everything asserted about announcements above was observed in Chromium against that bundle.

---


jooyoungseo:
Thanks — I checked the roxygen finding and it splits in two: **the observation is right, the attribution is not.**

## The observation holds

Confirmed exactly as described. `DESCRIPTION` has no `Roxygen: list(markdown = TRUE)`, no file uses `@md`, and the regenerated `.Rd` carries the syntax literally:

```
man/build_axis_config.Rd:24:  to [build_axes()], that drops the axis key entirely.
man/build_axes.Rd:18:    empty, so it serializes as the JSON object `{}` rather than as `[]`.
```

Someone running `?build_axes` sees backticks. That is a real defect.

## It is not a regression, and it is much larger than this PR

The finding calls it "a real formatting regression relative to the rest of the codebase's convention". Checked against `main`, unmodified:

| On `main`, in roxygen blocks | Count |
|---|---|
| Lines using backticks | 130 |
| Lines using `[fn()]` link syntax | 21 |
| **Total** | **149**, across **17 files** |
| Lines using `\code{}` | 71 |
| Lines using `\link{}` | 7 |

The markdown-ish style is not what this PR broke away from — it is already the **dominant** style in the package, by roughly 2:1 over Rd markup. `[show()]`, `[maidr_on()]` and `[maidr_off()]` in `R/base_r_wrapper_exports.R` are on `main` today, and the shipped `man/` on `main` already renders them literally (`man/adjust_chartseries_bracket.Rd`, `man/Ggplot2CandlestickProcessor.Rd` and a dozen others). `?show`'s "See also" is already showing literal brackets to users, and has been.

So converting only my four new blocks to `\code{}`/`\link{}` would fix 4 of 149 instances while making those blocks the odd ones out in files whose neighbouring comments all use backticks. That trades a uniform defect for a non-uniform one and leaves `?show` — a genuinely user-facing topic, unlike everything in this PR, which is all `@keywords internal` — still broken.

## What I am doing instead

The correct fix is the other option the finding offers: `Roxygen: list(markdown = TRUE)` in `DESCRIPTION`, which fixes all 149 lines at once and matches what the existing comments were evidently written expecting. That is a repo-wide change needing its own `R CMD check` — enabling markdown reinterprets every existing block, so anything containing `*`, `_` or `[` needs checking for mangling, and the `.Rd` diff will span most of `man/`.

Filing it as its own issue and PR rather than expanding a pie-sign fix into a package-wide documentation change. **Not deferring it** — it is next in the queue, and I will link the PR here.

Everything else in the review matches what I found: the `below <- ymax <= 0 & ymin < 0` boundary analysis, the `attach_axis_format()` call-site sweep, and the `NULL$x` degradation after the guards were dropped are all as stated. CI is green apart from `macos-15-intel (release)`, still running.

---

jooyoungseo:
As promised — the roxygen markdown finding is now #119, fixed in #120.

Enabling `Roxygen: list(markdown = TRUE)` turned out to need four escape fixes, all found by rendering every page with `Rd2txt` before and after and diffing the text rather than reading the diff. One of them was worse than cosmetic: `\%` gets escaped again into `\\%`, which truncates the line and in `create_maidr_iframe.Rd` swallowed the rest of an `\itemize` block, leaving malformed Rd. `R CMD check` on the result is item-for-item identical to `main`, and `testthat::test_local()` gives FAIL 0 | PASS 4361.

---

PR #120, opened 2026-08-09
docs: interpret the markdown the roxygen comments were already written in

### LINKED ISSUE #119: Roxygen blocks use markdown syntax the package never enabled, so `?help` shows literal backticks
Most of this package's roxygen comments are written in markdown — backticks for code, `[fn()]` for links — but `DESCRIPTION` carries no `Roxygen: list(markdown = TRUE)` and no file uses `@md`. roxygen2 therefore copies the syntax through verbatim, and the shipped `man/` renders it as literal characters.

Surfaced by review on #118, which correctly spotted it in that PR's new blocks. It is not specific to that PR — the same syntax is on `main` today, in quantity.

## Scale

Counted on `main`, unmodified:

| In roxygen blocks under `R/` | Lines |
|---|---|
| Backticks | 130 |
| `[fn()]` link syntax | 21 |
| **Total** | **149**, across **17 files** |
| `\code{}` | 71 |
| `\link{}` | 7 |

The markdown style is the dominant one, by roughly 2:1. This is not drift in one corner.

## What a user sees

Already in the shipped `man/`, before any of #118's changes:

```
man/base_r_wrapper_exports.Rd   [show()] for displaying accessible plots, [maidr_on()],
man/adjust_chartseries_bracket.Rd   ... via base R `title()` with `par(adj=1)`
man/Ggplot2CandlestickProcessor.Rd  ... expands into TWO ggplot layers
```

`?show` — an exported, user-facing topic — already shows `[show()]` and `[maidr_off()]` as literal bracketed text in its See also section, with no link. Most of the rest are `@keywords internal`, so the visible damage is concentrated in the exported topics, but the inconsistency is everywhere.

## Fix

Add to `DESCRIPTION`:

```
Roxygen: list(markdown = TRUE)
```

then regenerate `man/`. This matches what the existing comments were evidently written expecting, and fixes all 149 lines at once. The alternative — rewriting them all to `\code{}`/`\link{}` — is a much larger diff that moves the package away from the style it already mostly uses.

## Why it needs care rather than a one-liner

Enabling markdown reinterprets **every** existing block, not just the ones that meant to be markdown:

- `*` and `_` become emphasis. Several blocks use `*` as a literal bullet (`R/base_r_arg_utils.R:72-82`) or inside prose, and `_`-containing identifiers appear unbacktricked in places.
- `[` becomes a link. Any bare `[` in prose (subscripting is discussed in several comments — `labRow[rowInd]`, `args[[2]]`) needs checking; some are already inside backticks and will be safe, some are not.
- `%` and `\%` escaping interacts with markdown differently — `R/base_r_heatmap_layer_processor.R:372` has `\%||\%`.

So the change is: enable it, regenerate, then read the `.Rd` diff for mangling rather than assuming it is mechanical. `R CMD check` must be clean afterwards.

## Scope note

Deliberately not folded into #118 — that PR is a pie sign fix plus Base R axis labels, and this would turn it into a package-wide documentation change touching most of `man/`. Tracked here so it gets its own diff and its own check.

jooyoungseo:
Closes #119. Raised by review on #118, which spotted it in that PR's new blocks; it turned out to be package-wide and predate it.

## The problem

Most roxygen comments here are written in markdown — backticks for code, `[fn()]` for links — but `DESCRIPTION` never enabled it and no file uses `@md`. roxygen2 copies the syntax through verbatim.

Counted on `main`:

| In roxygen blocks under `R/` | Lines |
|---|---|
| Backticks | 130 |
| `[fn()]` links | 21 |
| **Total** | **149**, across **17 files** |
| `\code{}` | 71 |
| `\link{}` | 7 |

Markdown is the dominant style by ~2:1. `?show` has been listing `[maidr_on()]` as literal bracketed text with no link, and a dozen internal topics show raw backticks.

So this enables it rather than rewriting 149 lines to Rd markup — that is the style the comments were evidently written for.

## Four escapes had to change

Enabling markdown reinterprets **every** block, not only the ones that meant to be markdown. I found these by rendering all 261 pages with `Rd2txt` before and after and diffing the text, rather than by reading the diff.

**`\%` is escaped again into `\\%`**, which Rd renders as a literal backslash and then starts a comment that eats the rest of the line:

```
before:  width: Width of the iframe (default: "100%")
after:   width: Width of the iframe (default: "100\
```

In `create_maidr_iframe.Rd` the truncation swallowed the rest of the `\itemize` block and left **malformed Rd** — `unknown macro '\item'`, `unexpected section header '\value'`. In markdown mode roxygen2 escapes the percent itself, so the three sites are now plain `%`.

**A bare bracket becomes a link.** The chartSeries header example `"[2024-01-12/2024-01-15]"` compiled to a link with no target:

```
* checking Rd cross-references ... WARNING
Missing link or links in documentation object 'adjust_chartseries_bracket.Rd':
  '2024-01-12/2024-01-15'
```

Now a code span, which is what it always was semantically.

**`[extract_data()]`** pointed at an R6 method, which has no topic to link to. Backticks, like every other method reference in the package. (From #104, not from #118.)

## Verification

Not assumed — each of these was checked:

| Check | Result |
|---|---|
| All `.Rd` parse without warnings | **261 / 261** |
| Every generated `\link[=target]` resolves to an `\alias` | **all 17** |
| `R CMD check` | 3 WARNINGs, 1 NOTE |
| `R CMD check` on `main`, same flags | 3 WARNINGs, 1 NOTE — **item for item identical** |
| `testthat::test_local()` | **FAIL 0 \| WARN 2 \| SKIP 98 \| PASS 4361** |

The baseline check was run by `git archive origin/main` into a clean tree and building it the same way, so the comparison is like for like. The remaining warnings are this container's missing `en_US.UTF-8` locale (pre-existing currency symbols in `format_utils.R`) and the vignette directory, an artefact of `--no-build-vignettes`. The 2 test warnings are the pre-existing `boxplot(notch = TRUE)` ones.

Every emphasis markdown introduced was checked to come from intentional prose (`*all*`, `*below*`, `*generic*`, `*internal*` and the bold headings in `maidr-package.Rd`) — no identifier containing `_` or `*` was mangled.

## One page renders better than before

`build_axes.Rd` on `main`:

```
empty, so it serializes as the JSON object  rather than as [].
```

The `{}` was being eaten — unescaped braces are Rd grouping. With markdown they are inside a code span and survive:

```
empty, so it serializes as the JSON object '{}' rather than as [].
```

## Scope

Comment text, `DESCRIPTION`, and regenerated `man/` only. No executable code changed — the four `R/` edits are all inside `#'` blocks. `NAMESPACE` unchanged.

---

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,5 +54,6 @@ Suggests:
     zoo,
     xts
 Config/testthat/edition: 3
+Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2
 VignetteBuilder: knitr
--- a/R/base_r_heatmap_layer_processor.R
+++ b/R/base_r_heatmap_layer_processor.R
@@ -369,7 +369,7 @@ heatmap_applies_revc <- function(args) {
 #'
 #' `heatmap()` gives an explicit `labRow=`/`labCol=` priority over the
 #' matrix's own dimnames, and subscripts it by the same ordering it applies to
-#' the data: `labRow[rowInd] \%||\% rownames(x) \%||\% (1L:nr)[rowInd]`. Reading
+#' the data: `labRow[rowInd] %||% rownames(x) %||% (1L:nr)[rowInd]`. Reading
 #' the labels off the reordered matrix therefore announced the dimnames -- or,
 #' for an unnamed matrix, bare indices -- while the axis showed the caller's
 #' strings.
--- a/R/base_r_pie_layer_processor.R
+++ b/R/base_r_pie_layer_processor.R
@@ -119,7 +119,7 @@ BaseRPieLayerProcessor <- R6::R6Class(
     #' @description Generate one selector per wedge, index-aligned to the data
     #' @param layer_info Layer information
     #' @param gt Grob tree to search
-    #' @param extracted_data Points from [extract_data()], used for the count
+    #' @param extracted_data Points from `extract_data()`, used for the count
     #' @return List of CSS selector strings, one per slice
     generate_selectors = function(layer_info, gt = NULL, extracted_data = NULL) {
       n_slices <- length(extracted_data)
--- a/R/format_utils.R
+++ b/R/format_utils.R
@@ -293,7 +293,7 @@ prefix_to_currency_code <- function(prefix) {
 #'
 #' Converts R strftime format strings to JavaScript Intl.DateTimeFormat options.
 #'
-#' @param format R date format string (e.g., "\%Y-\%m-\%d")
+#' @param format R date format string (e.g., "%Y-%m-%d")
 #' @return List of Intl.DateTimeFormat options
 #' @keywords internal
 r_date_format_to_intl_options <- function(format) {
--- a/R/svg_utils.R
+++ b/R/svg_utils.R
@@ -652,7 +652,7 @@ extract_rect_index_from_id <- function(grob_id) {
 #' Reposition chartSeries date-range bracket header to prevent clipping
 #'
 #' `quantmod::chartSeries()` renders a bracketed date-range header
-#' (e.g. "[2024-01-12/2024-01-15]") via base R `title()` with `par(adj=1)`.
+#' (e.g. `"[2024-01-12/2024-01-15]"`) via base R `title()` with `par(adj=1)`.
 #' For short timeseries the text width exceeds the available right margin
 #' and the closing bracket is clipped at the SVG viewBox edge. This is
 #' upstream quantmod issue #129 (open since 2016, no fix). See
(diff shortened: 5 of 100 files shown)


jooyoungseo:
Good catch on `man/Ggplot2PieLayerProcessor.Rd`, and it checks out — `man/` was already stale relative to `R/` before this PR:

```
$ git show origin/main:R/ggplot2_pie_layer_processor.R  | grep -c negative   →  3
$ git show origin/main:man/Ggplot2PieLayerProcessor.Rd  | grep -c negative   →  0
```

So #118 landed the sign-fix docstring in the source without a `man/` regeneration, and running `roxygenise` here swept it in. Recording it plainly rather than leaving it as unexplained content in the diff: **that file's ~16-line paragraph about negative slices is #118's documentation catching up, not markdown doing anything.** It would have appeared in whatever PR next regenerated the docs.

Worth noting for the future: nothing in CI compares `man/` against `R/`, so this drift is silent. Not filing it as an issue on the strength of one instance, but if it recurs a `roxygen2::roxygenise()`-and-check-clean step in CI is the fix.

The `\itemize` indentation change in `maidr-package.Rd` is roxygen2's own formatting and does not reach the rendered page — I diffed `Rd2txt` output for all 97 changed pages, and that file's rendered text is unchanged apart from the intended markup.

On the unverifiable claims: fair, and the numbers are `261/261` Rd files parsing clean, and `R CMD check` giving 3 WARNINGs + 1 NOTE both here and on a `git archive origin/main` tree built with identical flags — same four items, so nothing new. Also confirmed the `\%` fix is not cosmetic: before it, `create_maidr_iframe.Rd` failed to parse at all (`unknown macro '\item'`, `unexpected section header '\value'`), which `R CMD check` would have caught on its own.

---

Issue #121, opened 2026-08-10
ggplot2: a factor y on geom_line() is announced as a level code, not a level name


jooyoungseo:
## What happens

A `geom_line()` whose `y` aesthetic is a factor emits the factor's **integer level code** as `y` and carries no `label`, so a screen reader announces the number ggplot2 uses internally rather than the name drawn on the axis.

```r
library(ggplot2)
stages <- factor(c("Awake", "REM", "N1", "N2", "N3", "N2"),
                 levels = c("N3", "N2", "N1", "REM", "Awake"), ordered = TRUE)
df <- data.frame(t = 1:6, stage = stages)

p <- ggplot(df, aes(t, stage, group = 1)) + geom_line()
maidr:::Ggplot2LineLayerProcessor$new(list(index = 1))$process(
  p, list(title = ""), built = ggplot2::ggplot_build(p)
)$data[[1]][[1]]
#> $x  "1"
#> $y  5
```

The reader hears **"5"**. The axis says **"Awake"**.

## What it should do

The same layer drawn with `geom_step()` already does the right thing, because #49 gave `Ggplot2StepLayerProcessor` a level-name lookup:

| | `geom_line()` | `geom_step()` |
|---|---|---|
| point keys | `x, y` | `x, y, label` |
| first point | `{"x":"1","y":5}` | `{"x":"1","y":5,"label":"Awake"}` |

`y` stays numeric on purpose — it drives sonification, braille, and the min/max range — and the level *name* rides alongside as `label`. `geom_line()` should carry the same pairing.

## Scope

`Ggplot2StepLayerProcessor` builds the lookup keyed on the *built* `y` value from the full unfiltered columns, so it stays correct regardless of which rows survive NA-dropping, panel filtering or group splitting. Moving that into `Ggplot2LineLayerProcessor` would let the step class inherit it rather than own it, and would cover `geom_path()` and `tidyquant::geom_ma()` at the same time, since they share the class.

Worth checking whether the smooth and Base R line processors have the same gap before deciding how wide the fix should be.

## Not a crash

An earlier claim in #49's description said this path failed serialisation with `No method asJSON S3 class: mapped_discrete`. [That does not reproduce](https://github.com/xability/r-maidr/pull/49#issuecomment-5243765145) — `mapped_discrete` also inherits `numeric`, so `jsonlite` encodes it fine, and `save_html()` succeeds. The defect is the missing name, not an error.

Found on `ea17360`.

[...]

PR #122, opened 2026-08-10
fix(ggplot2): carry the level name for a factor y on a line

jooyoungseo:
Addresses the producer half of #121. Filed [xability/maidr#785](https://github.com/xability/maidr/issues/785) for the consumer half.

## Read this first: what this does *not* do

I originally wrote this PR claiming a factor-`y` line chart would announce "Awake" instead of "5". **It does not.** I checked it in Chromium against the bundled `maidr.js` rather than stopping at the payload, and the announcement is unchanged.

The `label` reaches the wire correctly — `{"x":"1","y":5,"label":"Awake"}` is in the emitted HTML — but **maidr's JS line trace ignores a per-point `label`**. Only its step trace reads one:

| trace | payload point | announced |
|---|---|---|
| `step` | `{"x":"1","y":5,"label":"Awake"}` | `Hour is 1, Sleep stage is Awake` |
| `line` | `{"x":"1","y":5,"label":"Awake"}` | `Hour is 1, Sleep stage is **5**` |

So this is the half r-maidr owns, and the reader-facing fix waits on xability/maidr#785. I would rather merge a correct half with an honest NEWS entry than a PR whose changelog claims a win that does not happen.

## The defect on this side

`ggplot_build()` replaces a factor level with its numeric position, so `geom_line()` and `geom_path()` emitted the code and no name at all.

| | before | after |
|---|---|---|
| `geom_line()` | `{"x":"1","y":5}` | `{"x":"1","y":5,"label":"Awake"}` |
| `geom_path()` | `{"x":"1","y":5}` | `{"x":"1","y":5,"label":"Awake"}` |
| `geom_step()` | already correct | unchanged |

## The shape: a move, plus a rewritten lookup

`geom_step()` already carried a level lookup (#49), and it was never step-specific — it belongs to whatever draws a discrete `y`. So it moves up to `Ggplot2LineLayerProcessor` and the step class inherits it:

- `geom_line()`, `geom_path()` and `tidyquant::geom_ma()` gain it by sharing the class — verified in the factory, all three route through `"line"`.
- `Ggplot2StepLayerProcessor` **loses its `extract_data()` override entirely** rather than growing a second copy. Its own tests pass untouched, which is the check that the inheritance is faithful.
- Blast radius is bounded: `Ggplot2LineLayerProcessor` has exactly one subclass, the step processor. The smooth and Base R line processors are separate classes.

**The lookup itself was rewritten twice, because review caught it and then the fix caught itself.** Both are worth recording since the first two versions look obviously right:

1. **Original (step's):** pair the built `y` column against the caller's column *by row index*. Wrong for `geom_line()` — `GeomLine$setup_data()` sorts the built data by `(PANEL, group, x)`, which is the documented difference from `geom_path()`, while the caller's column keeps its own order:

   ```
   t = c(3, 1, 2), stage = c("N1", "N3", "N2")
   built y = 1, 2, 3   ->   labelled N1, N3, N2      (should be N3, N2, N1)
   ```

2. **Second attempt:** read `levels(original_y)`, order-independent. Still wrong — a discrete scale defaults to `drop = TRUE`, so a factor declaring five levels of which two are drawn is coded `1..2`, not by position in `levels()`:

   ```
   declared: N3,N2,N1,REM,Awake    drawn: N3, Awake    built y = 2, 1
   levels() naming  ->  code 2 = "N2"      the axis says "Awake"
   ```

3. **Shipped:** read the **panel's own y labels** — the strings ggplot2 draws on the axis. Order-independent, drop-aware, and it makes the announcement agree with what a sighted reader sees, which is the actual contract rather than an implementation detail being reconstructed.

Two robustness notes: it gates on the built column's `mapped_discrete` class rather than a scale accessor, since those have moved between ggplot2 versions; and it takes `panel_id`, so a facet with `scales = "free_y"` reads its own scale instead of panel 1's.

`get_original_y_column()` is **deleted** — nothing reaches it now, and it was the sole source of the row-alignment hazard. A duplicate `get_layer()` left on the step processor is deleted too; my earlier edit asserted the block existed but not that the removal had happened, so it silently no-opped.

`y` stays numeric — sonification, braille and the min/max range read it. A continuous `y` is untouched, so no `label` is emitted and the number is announced, which is right there. The `mapped_discrete` → `numeric` class strip is guarded on `is.numeric()`, TRUE for `mapped_discrete` (it inherits `numeric`) and FALSE for a stringified `y`, so it cannot produce `NA`.

## Correcting the record from #49

#49's description claimed this path failed serialisation with `No method asJSON S3 class: mapped_discrete`. **It does not** — `mapped_discrete` inherits `numeric`, so `jsonlite` encodes it and `save_html()` succeeds. I generalised a crash I hit while building the step processor without testing `geom_line()`. Retracted [in that thread](https://github.com/xability/r-maidr/pull/49#issuecomment-5243765145) and restated accurately in #121. Stripping the class is still worth doing, but it is tidiness, not a crash fix.

## Testing

Six regression tests, **each checked against a mutant** rather than assumed to bite:

| mutation | result |
|---|---|
| neuter the label attach | name test **errors** (`pt$label` length 0), grouped test **fails** (`NULL` vs `"A"`) |
| disable the class strip | bare-number test **fails** (`"mapped_discrete" "numeric"` vs `"numeric"`) |
| restore the row-index pairing | unsorted test **fails** (`1=N1 2=N3 3=N2` vs `1=N3 2=N2 3=N1`) |
| restore the `levels()` naming | dropped-level test **fails** |

They cover: level names in order; `y` left a bare number and serialisable; a continuous `y` emitting no label; a grouped factor `y` keeping both `z` (the series) and `label` (the level); an unsorted `geom_line()`; and a factor with unused levels. The step class keeps its own dropped-level test, replacing the one that asserted the now-deleted helper's mechanics rather than any behaviour.

`testthat::test_local()` — **0 failures, 0 errors, 1 skip** (pre-existing: `test-ggplot2-orchestrator.R:89`).

`man/` regenerated with roxygen2 7.3.2 and `Roxygen: list(markdown = TRUE)`, matching `DESCRIPTION`.

## Found while fixing this, filed separately

The same positional pairing is used for **x**, and it is broken on `main` today: an unsorted `geom_line()` announces every point with another point's x. Filed as #123 with the verified scope — non-faceted single and multi-series are affected; faceted plots and `geom_path()` are not, and serve as controls. Deliberately not widened into this PR.


jooyoungseo:
Both findings were right. Fixed in `5138987`, and the row-order one turned out to be hiding a second bug.

## 1. Row misalignment — confirmed, and it went deeper

Ran your repro rather than taking it on reading:

```
t = c(3, 1, 2), stage = c("N1", "N3", "N2")
built:     x = 1,2,3   y = 1,2,3
extracted: y=1 label=N1 | y=2 label=N3 | y=3 label=N2     ← wrong, exactly as predicted
```

Rewriting the lookup to read `levels(original_y)` fixed that — and was **still wrong**, which the same repro caught. A discrete scale defaults to `drop = TRUE`, so a factor declaring five levels of which two are drawn is coded `1..2`, not by position in `levels()`:

```
levels declared: N3,N2,N1,REM,Awake
built y:         2,1
axis draws:      N3, Awake
levels() naming: code 2 → "N2"     ← wrong; the axis says "Awake"
```

The lookup now reads **the panel's own y labels** — the strings ggplot2 actually draws on the axis. Order-independent, drop-aware, and it makes the announcement agree with what a sighted reader sees, which is the real contract here rather than an implementation detail I was reconstructing.

Two robustness notes on the new version: it gates on the built column's `mapped_discrete` class rather than a scale accessor, since those have moved between ggplot2 versions; and it takes `panel_id`, so a facet with `scales = "free_y"` reads its own scale instead of panel 1's.

`get_original_y_column()` is **deleted** — nothing reaches it now, and it was the sole source of the row-alignment hazard. Its unit test asserted the mechanism rather than the behaviour, so it is replaced by one pinning the dropped-level case on the step class.

Three new tests, each checked against a mutant: the unsorted-x case and the dropped-level case both fail on the two previous implementations.

## 2. Duplicate `get_layer()` — confirmed, my mistake

You were right that it was copied, not moved. My edit asserted the block *existed* but never asserted the removal had *happened*, so the replace silently no-opped and left the copy behind while adding another to the parent. Deleted, along with the stray blank lines; `man/Ggplot2StepLayerProcessor.Rd` regenerated so it no longer lists the method.

## On the pre-existing x bug you predicted

You wrote that `get_original_x_column()` "likely has the identical latent issue". It does, and it is worse than the label one — it is on `main` today, independent of this PR:

```
# on merged main (ea17360), no changes from this branch
t = c(3, 1, 2)  ->  extracted x = 3, 1, 2 against built x = 1, 2, 3
```

Every point of an unsorted `geom_line()` is announced with **another point's x**. Filing that separately rather than widening this PR, since it needs its own reproduction and its own regression tests.

`testthat::test_local()`: 0 failures, 0 errors, 1 pre-existing skip.

---

PR #124, opened 2026-08-10
fix(ggplot2): give an unsorted line each point's own x

### LINKED ISSUE #123: ggplot2: an unsorted geom_line() announces every point with another point's x
## What happens

`geom_line()` whose source rows are not already sorted by `x` announces the **wrong x for every point**. Not a shifted label — the coordinate itself is another row's.

```r
library(ggplot2)
df <- data.frame(
  t = c(3, 1, 2),
  stage = factor(c("N1", "N3", "N2"),
                 levels = c("N3", "N2", "N1", "REM", "Awake"), ordered = TRUE)
)
p <- ggplot(df, aes(t, stage, group = 1)) + geom_line()

ggplot2::ggplot_build(p)$data[[1]][c("x", "y")]
#>   x y
#> 1 1 1     <- built data is x-sorted
#> 2 2 2
#> 3 3 3

maidr:::Ggplot2LineLayerProcessor$new(list(index = 1))$extract_data(p)
#>   x = "3", y = 1     <- x is row 1 of the CALLER's data
#>   x = "1", y = 2
#>   x = "2", y = 3
```

The built row that is `x = 1, y = 1` is announced as `x = 3`. A reader is told the series runs 3 → 1 → 2 when it is drawn 1 → 2 → 3, so the shape they reconstruct is not the shape on screen.

Reproduced on `ea17360` with no branch changes applied.

## Verified scope

Every path below was run, not inferred:

| path | built x | emitted x | verdict |
|---|---|---|---|
| numeric, sorted (control) | `1,2,3` | `1,2,3` | ✅ |
| numeric, unsorted | `1,2,3` | **`3,1,2`** | ❌ |
| Date, unsorted | `19723,19724,19725` | **`01-03, 01-01, 01-02`** | ❌ |
| discrete, unsorted | `1,2,3` | **`c,a,b`** | ❌ |
| multi-series, unsorted | `1,2,3` per group | **`3,1,2`** on both series | ❌ |
| faceted, unsorted | `1,2,3` per panel | `1,2,3` | ✅ |
| `geom_path()`, unsorted | `3,1,2` | `3,1,2` | ✅ |

`y` is correct throughout — this is confined to x recovery.

**Faceted plots are NOT affected**, correcting a guess in the first draft of this issue. The `!is.null(panel_id)` branch recovers x from `sort(unique(plot$data[[x_col]]))` indexed by the built position and then through the panel's break/label mapping — both keyed by *value*, so ordering cannot desynchronise them. It is the non-faceted branch that reaches `get_original_x_column()` and indexes it positionally.

`geom_path()` escapes because `GeomPath$setup_data()` does not reorder, so pairing by row happens to be right. `geom_line()` sorts by `(PANEL, group, x)` — that sort is the documented difference between the two geoms.

## Where it is

`extract_single_line_data()` pairs the caller's column with the built row index:

```r
x_val <- if (!is.null(orig_x)) {
  self$format_x_value(orig_x[i])        # caller's row i
} else {
  self$format_x_value(layer_data$x[i])  # built row i
}
```

`extract_multiline_data()` does the same, which is why both series come out identically scrambled.

`get_original_x_column()` exists for a real reason — built x for a Date is a numeric day count (`19723`), so something must turn it back into `2024-01-03`. The recovery is needed; reading it out of the caller's column *by row number* is what is wrong.

## Fix direction

Recover from the **built** value rather than the caller's rows, per scale type:

- discrete x → the panel's own x labels keyed by position, exactly what #122 does for a discrete y
- Date / POSIXct x → convert the built numeric through the scale's transformation
- plain numeric x → the built value already *is* the value; no recovery needed

All three are keyed by value or by the axis, so none can desynchronise. That would let `get_original_x_column()`'s positional indexing be deleted the same way `get_original_y_column()` was in #122.

Bounded to two methods, non-faceted only, with `geom_path()` and the faceted branch as controls that must not change.

## Severity

Higher than the label gap in #121. A wrong name is recoverable from context; a wrong coordinate silently misstates the data, and nothing in the payload signals it.

Predicted by the reviewer on #122 while looking at the y-label equivalent; kept out of that PR because it needs its own reproduction and regression tests.

jooyoungseo:
Closes #123.

## The defect

`GeomLine$setup_data()` sorts the built data by `(PANEL, group, x)` — that sort is the documented difference between `geom_line()` and `geom_path()` — while the caller's column keeps its own order. x was recovered by pairing the two **by row number**, so every point of an unsorted `geom_line()` carried another point's coordinate:

```r
df <- data.frame(t = c(3, 1, 2), v = c(30, 10, 20))
p  <- ggplot(df, aes(t, v)) + geom_line()

ggplot_build(p)$data[[1]]$x     #> 1 2 3     <- as drawn
# announced                     #> 3, 1, 2
```

A reader is told the series runs 3 → 1 → 2 when it is drawn 1 → 2 → 3. This is worse than the wrong *label* in #121: a wrong coordinate cannot be recovered from context, and nothing in the payload signals it.

## Verified scope

Each row was run, not inferred:

| path | built x | before | after |
|---|---|---|---|
| numeric, sorted | `1,2,3` | `1,2,3` | `1,2,3` |
| numeric, unsorted | `1,2,3` | **`3,1,2`** | `1,2,3` |
| Date, unsorted | days-since-epoch | **`01-03, 01-01, 01-02`** | `01-01, 01-02, 01-03` |
| discrete, unsorted | `1,2,3` | **`c,a,b`** | `a,b,c` |
| multi-series, unsorted | `1,2,3` per group | **`3,1,2`** on both | `1,2,3` on both |
| faceted, unsorted | `1,2,3` per panel | `1,2,3` | unchanged |
| `geom_path()`, unsorted | `3,1,2` | `3,1,2` | unchanged |

## The fix

x is recovered from the **built value itself**, which is order-proof by construction, per scale type:

- **discrete** — the panel's own x labels indexed by the level code, the same source `build_level_lookup()` uses for a discrete y (#122)
- **Date / POSIXct / log** — the scale transformation's inverse, via the existing `get_x_transformation()` helper, which already tolerates both the `get_transformation()` and `$trans` spellings across ggplot2 versions
- **plain numeric** — the built value already *is* the value

Both extractors take the recovered vector rather than deriving it themselves, because it has to be computed **before** the label mapping in `extract_data()` overwrites `layer_data$x`. A first attempt that recovered inside the extractors read the rewritten column and broke Date recovery outright (`NA, NA, NA`) — worth recording, because it looked correct until run.

## Two things deliberately left alone

**The faceted branch.** It recovers x by *value* — through `sort(unique(...))` and the panel's break/label mapping — so it never had the defect. It is also what honours a caller's own `labels =` formatter. An earlier revision applied the new recovery there too and regressed the existing free-scale test from `"<1>"` to `"1"`: the raw datum is not what that axis shows. Scoped to non-faceted.

**`geom_path()`.** It connects observations in the order they appear, and `GeomPath$setup_data()` does not reorder, so `3, 1, 2` is correct there and must stay.

Both are controls in the test file rather than assumptions in a comment.

## Testing

Five regression tests. Four fail under the old pairing:

| test | under the old pairing |
|---|---|
| unsorted numeric | **fails** |
| unsorted discrete | **fails** |
| unsorted Date | **fails** |
| unsorted multi-series | **fails** (both series) |
| `geom_path()` order preserved | passes either way — the control |

The `geom_path()` test passing under the mutant is the point: it proves the fix did not simply sort everything.

`testthat::test_local()` — **0 failures, 0 errors, 1 skip** (pre-existing: `test-ggplot2-orchestrator.R:89`).

`man/` regenerated with roxygen2 7.3.2 and `Roxygen: list(markdown = TRUE)`.

## Related

Found while fixing #122 — the y-label equivalent — and predicted there by the reviewer. #121 remains open: its producer half shipped in #122, and the reader-facing half waits on xability/maidr#785.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DZocjnGTiRvLDCbcvqYqa3)_


jooyoungseo:
Adversarially probed this before waiting on review, since my own "obviously correct" version was wrong twice while building it. Seven untested combinations, all correct:

| case | announced |
|---|---|
| NA in y, unsorted | `2,3,4` — the surviving rows, still aligned after the drop |
| duplicate x values | `1,1,2,2` |
| `scale_x_log10()`, unsorted | `1,10,100` |
| `scale_x_reverse()`, unsorted | `3,2,1` — matches the drawn order under the reversed scale |
| POSIXct, unsorted | `01:00, 02:00, 03:00` |
| single point | `5` |
| expression x (`aes(t + 0, v)`), unsorted | `1,2,3` |

The NA case is the one I most expected to break: the recovered vector is aligned to the built frame and the NA-drop happens afterwards against the same index, so it holds.

## One behaviour change beyond the bug, disclosed

That last row is not just a fix. On `main`, a **non-faceted continuous x** was announced differently depending on how the aesthetic was written:

| `main` | announced |
|---|---|
| `aes(t, v)` — raw column | `1, 2, 3` (the datum; a `labels =` formatter was already ignored) |
| `aes(t + 0, v)` — expression | `1.0, 2.0, 3.0`, or `<1>, <2>, <3>` with a custom formatter |

The difference is incidental: with a raw column, `get_original_x_column()` found something and bypassed the label mapping; with an expression it found nothing and fell through to it. Same plot, same data, two different readings.

This PR makes non-faceted continuous x uniformly announce **the datum**. I think that is the right resolution rather than the reverse, because on a continuous axis labels exist only at tick breaks — there is no label for a point that does not sit on one, so the value is the only per-point answer available. The `<1>` reading worked only because that data happened to land exactly on breaks.

The faceted path makes the opposite choice and is deliberately untouched — its existing test pins `"<1>", "<2>", "<3>"`, and an earlier revision of this PR that applied the new recovery there regressed it.

Flagging it rather than letting it ride: it is a real change in what a reader hears for expression-x continuous plots, it is not required to fix #123, and if the project would rather preserve `main`'s expression-x behaviour exactly I can narrow the recovery to the ordering fix alone.

---

jooyoungseo:
Both points taken, fixed in `2e9a91e`. The second one caught a wrong assumption of mine.

## `get_original_x_column()` deleted

You are right, and the framing is the useful part: it is not merely dead, it is *dead code containing the defect this PR closes*. Confirmed unreachable — `grep -rn get_original_x_column R/ tests/` returns only its own definition. Deleted, both man pages regenerated, and the method index no longer lists it in either.

My PR body said its positional indexing "can go", which was a statement of intent I then did not act on. Thanks for not letting that slide.

## The unsorted-faceted test — and what writing it exposed

Added, and it immediately failed. **The code was right; my expectation was wrong.**

I wrote it asserting `"<1>", "<2>", "<3>"`, believing the faceted branch always honours a `labels =` formatter. It does not. It honours one only when it *cannot* resolve a raw column — the neighbouring free-scale test gets `"<1>"` because its x is the expression `x + 0`. With a plain column, the faceted branch announces the datum, same as the non-faceted path now does.

So the test asserts `"1", "2", "3"`, and carries that explanation, because the distinction is not visible from the call site.

That is the third time in this pair of PRs that a label-versus-datum assumption of mine was wrong and only execution caught it. It also sharpens the disclosure in my earlier comment: the datum-vs-formatter split is not faceted-vs-not, it is **raw column vs expression**, on both paths. `main` already announced the datum for a raw column everywhere; the expression case was the odd one out, and this PR makes non-faceted consistent with the dominant behaviour.

## On the shared-helper suggestion

Agreed that `recover_x_values()`'s discrete branch and `build_level_lookup()` are near-duplicates — both index a panel's `get_labels()` by integer code behind the same usability guard. I have left them separate here deliberately: extracting the helper touches the y path that shipped ninety minutes ago in #122, and I would rather that consolidation be its own change with its own review than a drive-by inside a bug fix. Worth doing; not worth widening this PR for.

`testthat::test_local()`: 0 failures, 0 errors, 1 pre-existing skip.

---

PR #129, opened 2026-08-10
style: replace literal non-ASCII characters in R sources with ASCII

### LINKED ISSUE #126: Literal non-ASCII characters in R/ trip R CMD check's portability NOTE
## What happens

31 literal non-ASCII characters — 20 em dashes and 11 rightwards arrows — sit in comments and roxygen blocks on 26 lines across 11 files in `R/`. `R CMD check` reports:

```
* checking R files for non-ASCII characters ... NOTE
Found the following files with non-ASCII characters:
Portable packages must use only ASCII characters in their R files,
using \uxxxx escapes for other characters.
```

Reproduce with the same call the check uses:

```r
files <- list.files("R", pattern = "[.][Rr]$", full.names = TRUE)
Filter(function(f) length(capture.output(tools::showNonASCIIfile(f))) > 0, files)
#> 11 files
```

Lines carrying at least one, per file:

| file | lines |
|---|---|
| `R/base_r_function_patching.R` | 5 |
| `R/ggplot2_violin_layer_processor.R` | 5 |
| `R/svg_utils.R` | 4 |
| `R/ggplot2_print_method.R` | 3 |
| `R/ggplot2_system_init.R` | 3 |
| `R/base_r_wrapper_exports.R` | 1 |
| `R/ggplot2_heatmap_layer_processor.R` | 1 |
| `R/ggplot2_line_layer_processor.R` | 1 |
| `R/ggplot2_patchwork_utils.R` | 1 |
| `R/ggplot2_plot_orchestrator.R` | 1 |
| `R/html_dependencies.R` | 1 |

## Why CI is green

`.github/workflows/R-CMD-check.yaml` uses `r-lib/actions/check-r-package@v2`, which defaults to failing on warnings and above. A NOTE therefore passes, so this has accumulated invisibly. It is a CRAN blocker rather than a runtime one — nothing misbehaves today.

## What it should do

Every occurrence is in a comment or a roxygen block, so the fix is textual and carries no behaviour change: an em dash becomes a plain ` - ` and an arrow becomes `->`. Prose reads fine either way, and ASCII keeps the sources portable without anyone having to remember an escape.

`DESCRIPTION` already declares `Encoding: UTF-8`; that governs the *encoding*, not the portability rule, which asks for ASCII in R sources regardless.

## Not the same as the lintr failure

`lintr::lint_package()` aborting with `subscript out of bounds` on `R/format_utils.R` is unrelated and is **not** a repository defect. That file's currency symbols are already written the portable way — as lowercase backslash-u escapes, which is exactly what the NOTE above asks for — and the abort only happens in a session whose `LC_CTYPE` is `C`:

```r
Sys.getlocale("LC_CTYPE")               #> "C"
lintr::get_source_expressions("R/format_utils.R")
#> Error: subscript out of bounds

# same file, same lintr, UTF-8 locale
#> ok, 108 expressions
```

Under a UTF-8 locale `lint_package()` runs to completion (429 pre-existing style findings, none of them errors). Nothing to change for it.

Found on `da7d4c3`.

jooyoungseo:
Closes #126.

## The NOTE

`R CMD check` asks for ASCII in R sources, and 31 literal characters — 20 em dashes and 11 rightwards arrows, on 26 lines across 11 files — were tripping it:

```
* checking R files for non-ASCII characters ... NOTE
Portable packages must use only ASCII characters in their R files,
using \uxxxx escapes for other characters.
```

It has been invisible because `.github/workflows/R-CMD-check.yaml` uses `check-r-package@v2`, which fails on warnings and above, so a NOTE passes. This is a CRAN blocker rather than a runtime one — nothing misbehaves today.

## Changes

Every occurrence is in a comment or a roxygen block, so this is textual and carries no behaviour change:

- an em dash used as a parenthetical break becomes a plain hyphen
- an arrow becomes `->` (`Q3→ymax` reads `Q3->ymax`, `value axis = Y → sort by svg_y` reads `-> sort by svg_y`)
- one roxygen sentence in `merge_line_layers()` that used a *pair* of em dashes as parentheses is reworded with commas, because the second one fell at end of line and a trailing ` -` read as a dangling operator

`man/` is regenerated for the five topics whose roxygen text changed. It was in sync with `R/` before this change and is in sync after.

Escapes would satisfy the rule too, but plain ASCII reads the same in prose and leaves nothing to remember.

## Verification

```r
files <- list.files(c("R", "man"), pattern = "[.](R|r|Rd)$", full.names = TRUE)
Filter(function(f) length(capture.output(tools::showNonASCIIfile(f))) > 0, files)
#> 0 files
```

Full test suite: **FAIL 0 | WARN 18 | SKIP 2 | PASS 5598**. The 18 warnings are pre-existing ggplot2 aesthetic warnings from the candlestick/patchwork fixtures, unrelated to this change.

## Not the lintr failure

`lintr::lint_package()` aborting with `subscript out of bounds` on `R/format_utils.R` is a separate thing and is **not** a repository defect, so nothing here addresses it. That file's currency symbols are already written the portable way, and the abort only happens in a session whose `LC_CTYPE` is `C`; under a UTF-8 locale `lint_package()` runs to completion. Detail is in #126.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DZocjnGTiRvLDCbcvqYqa3)_

### DIFF
--- a/R/base_r_function_patching.R
+++ b/R/base_r_function_patching.R
@@ -320,8 +320,8 @@ snapshot_call_env <- function(args, caller_env) {
 #' Evaluate an expression, muffling the retry's promise-restart warning
 #'
 #' When a call fails part-way through forcing an argument, that argument's
-#' promise is left interrupted. Forcing it again — which both the retry and
-#' the argument recording do — makes R warn "restarting interrupted promise
+#' promise is left interrupted. Forcing it again - which both the retry and
+#' the argument recording do - makes R warn "restarting interrupted promise
 #' evaluation". It is an artifact of retrying, not anything the user's call
 #' did, so it is muffled; every other warning passes through untouched.
 #'
@@ -471,7 +471,7 @@ initialize_base_r_patching <- function(include_low = TRUE, include_layout = TRUE
   # a later call (e.g. via a packageEvent hook for quantmod).
   lapply(fns_to_wrap, wrap_function)
 
-  # Special handling for S3 generics (lines, points) — only the first
+  # Special handling for S3 generics (lines, points) - only the first
   # call actually installs them (gated on .saved_graphics_fns).
   if (is.null(.maidr_patching_env$.saved_graphics_fns[["lines"]]) ||
       is.null(.maidr_patching_env$.saved_graphics_fns[["points"]])) {
@@ -513,7 +513,7 @@ wrap_function <- function(function_name) {
   tryCatch(
     assign(function_name, wrapper, envir = ns),
     error = function(e) {
-      # Namespace is sealed — wrapper was already installed during .onLoad
+      # Namespace is sealed - wrapper was already installed during .onLoad
       NULL
     }
   )
@@ -1178,7 +1178,7 @@ apply_barplot_sorting <- function(args) {
 #' @return NULL (invisible)
 #' @keywords internal
 restore_original_functions <- function() {
-  # Deactivate patching — wrappers will pass through to originals
+  # Deactivate patching - wrappers will pass through to originals
 
   .maidr_patching_env$.patching_active <- FALSE
 
--- a/R/base_r_wrapper_exports.R
+++ b/R/base_r_wrapper_exports.R
@@ -130,7 +130,7 @@ chartSeries <- function(...) {
   # expressions, so an explicit `TA = NULL` reaches quantmod as a `name`
   # and dies in `sapply(chob@passed.args$TA, function(x) eval(x@call))`.
   # Retrying with the call rebuilt in the caller's frame gives quantmod
-  # the literal arguments it expects — the same fallback the generated
+  # the literal arguments it expects - the same fallback the generated
   # wrappers use (see retry_call_in_caller_frame()).
   call_failed <- FALSE
   result <- tryCatch(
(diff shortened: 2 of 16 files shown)


jooyoungseo:
On the scoping question — confirmed empirically, and the scoping in this PR is right.

The check is `tools:::.check_package_ASCII_code`, and it builds its file list from exactly one directory:

```r
code_dir <- file.path(dir, "R")
if (dir.exists(code_dir)) {
    R_files <- list_files_with_type(code_dir, "code", full.names = FALSE,
                                    OS_subdirs = c("unix", "windows"))
    for (f in R_files) {
        text <- readLines(file.path(code_dir, f), warn = FALSE)
        if (.Call(C_check_nonASCII, text, !respect_quotes))
            wrong_things <- c(wrong_things, f)
    }
}
```

So it reads `R/` and its `unix` / `windows` OS subdirectories, and nothing else. `tests/` and `inst/` are outside it, which is why the em dashes still in `test-svg-utils.R` and `inst/examples/ggplot2/step.R` do not contribute to this NOTE and are deliberately left alone here.

Worth adding that `man/` is not covered by this check either. It is included in the PR's verification only because five `.Rd` files are generated from roxygen blocks this change touches, so they had to be regenerated to stay in sync — not because the NOTE would have flagged them.

---

PR #140, opened 2026-08-11
fix(format): resolve a currency prefix outside a UTF-8 locale

### LINKED ISSUE #139: A euro axis is announced as dollars when the session locale is not UTF-8
## What happens

The same chart, the same code, emits a different currency depending on the session's locale:

```r
euro <- rawToChar(as.raw(c(0xe2, 0x82, 0xac)))   # a euro sign
Encoding(euro) <- "UTF-8"                        # marked, as readr/jsonlite give it

p <- ggplot(df, aes(cat, val)) + geom_col() +
  scale_y_continuous(labels = scales::label_dollar(prefix = euro))

maidr:::extract_axis_format(ggplot2::ggplot_build(p), "y")
```

| locale | emitted |
|---|---|
| `C.UTF-8` | `$currency: "EUR"` ✅ |
| `C` | `$currency: "USD"` ❌ |

A euro-denominated axis is read out to a screen-reader user **as dollars**. No error, no warning, and the payload is internally consistent — nothing looks wrong from the inside. Every non-dollar currency in the map is affected: GBP, JPY, INR, KRW, RUB, ILS, PHP, THB, PLN.

## Why

`R/format_utils.R` writes the map keys as escapes, which is correct CRAN practice and what #129 introduced:

```r
currency_map <- list("$" = "USD", "€" = "EUR", "£" = "GBP", ...)
```

In a **non-UTF-8 locale** those keys land with `Encoding()` of `"unknown"` — the right bytes, no declared encoding. An incoming prefix that *is* marked `"UTF-8"` then fails to match, because `match()` translates to the native encoding first and the euro has no representation in `C`:

```
locale: C
key       : unknown | e2 82 ac
probe_utf : UTF-8   | e2 82 ac
probe_utf %in% names(m): FALSE      # identical bytes, no match
```

In a UTF-8 locale the escapes yield `"UTF-8"`-marked keys and both sides agree, which is why CI has never seen this.

`detect_scales_format_type()` has the same comparison against `currency_symbols`, so it is one bug in two places.

## Where the marked string comes from

A user's own `prefix = "€"` literal is read in the session's own encoding, so it matches and works. The mismatch needs a prefix that carries a `UTF-8` mark while the session is not UTF-8 — which is the normal output of `readr`, `jsonlite`, `read.csv(encoding = "UTF-8")`, or anything that sets `Encoding()`. Labels driven from a loaded data file are the realistic path.

A `C` locale is not exotic: it is the default in many Docker images, including the one this was found in, and in some CI and RStudio Server setups.

## Fix

Declare the encoding on both sides before comparing, rather than leaving it to the locale. The escapes always produce UTF-8 bytes regardless of the running locale, so saying so is accurate, not a coercion:

```r
as_utf8 <- function(x) { Encoding(x) <- "UTF-8"; x }
match(as_utf8(prefix), as_utf8(names(currency_map)))
```

Verified across both locales and all three input encodings — UTF-8-marked, unmarked-but-UTF-8-bytes, and plain ASCII:

```
=== C ===              === C.UTF-8 ===
  utf8    -> EUR         utf8    -> EUR
  native  -> EUR         native  -> EUR
  dollar  -> USD         dollar  -> USD
```

`enc2utf8()` alone does **not** work: it treats an `"unknown"` string as native and mistranslates the keys, leaving the C-locale case still failing.

## Scope

- [ ] `prefix_to_currency_code()` — the map lookup
- [ ] `detect_scales_format_type()` — the `currency_symbols` membership test
- [ ] Tests that run the comparison under a non-UTF-8 locale, or that construct a `UTF-8`-marked prefix directly, since the existing tests pass in both today

## Not the same thing

`pkgload::load_all()` emits ten `unable to translate '<U+20AC>' to native encoding` warnings in a `C` locale. Those are `parse()` reporting it cannot render the escapes for source references, and they are cosmetic — the bytes it stores are correct either way. They are how this was noticed, not the defect.

Found on `089df23`.

jooyoungseo:
## What was wrong

A euro-denominated axis was announced to a screen-reader user **as dollars**. Same chart, same code — the answer decided by the session locale:

```r
euro <- rawToChar(as.raw(c(0xe2, 0x82, 0xac))); Encoding(euro) <- "UTF-8"
p <- ggplot(df, aes(cat, val)) + geom_col() +
  scale_y_continuous(labels = scales::label_dollar(prefix = euro))
maidr:::extract_axis_format(ggplot2::ggplot_build(p), "y")
```

| locale | emitted |
|---|---|
| `C.UTF-8` | `$currency: "EUR"` |
| `C` | `$currency: "USD"` |

Every non-dollar currency was affected — GBP, JPY, INR, KRW, RUB, ILS, PHP, THB, PLN — with no error, no warning, and a payload that is internally consistent. Nothing looks wrong from the inside.

## Why — and why my first diagnosis in #139 was only half of it

I opened #139 thinking this was an encoding *mark* mismatch, fixable by declaring both sides UTF-8 before comparing. That fix did not work, and printing the parsed function showed why:

```
`<U+20AA>` = "ILS", `<U+20B1>` = "PHP", `z<U+0142>` = "PLN"
```

**The key was already destroyed at parse time.** A `\uXXXX` escape behaves differently depending on where it sits:

- in an ordinary string it keeps its bytes and its UTF-8 mark in every locale;
- in a **tag** — `list("€" = "EUR")` — it becomes a symbol, symbols must be representable in the native encoding, and outside a UTF-8 session R substitutes the placeholder *text* `<U+20AC>`.

So no comparison could have recovered it. The map had to stop being a named list. It is now two parallel vectors matched on index, which is why `CURRENCY_SYMBOLS` — a plain character vector all along — was never affected and `detect_scales_format_type()` half-worked.

`as_utf8()` covers the genuine second half, which #139 did describe correctly: a prefix from `readr`, `jsonlite`, or `read.csv(encoding = "UTF-8")` carries a `"UTF-8"` mark, and native-marked keys will not match it even on identical bytes. `enc2utf8()` is **not** a substitute — it reads an `"unknown"` string as native and mistranslates these keys, leaving the C-locale case still failing. I tried that first.

## Verification

All 15 mapped prefixes, both locales, and both input encodings (UTF-8-marked and unmarked):

```
=== C ===        locale: C         all correct: TRUE
=== UTF-8 ===    locale: C.UTF-8   all correct: TRUE
```

## Tests

`tests/testthat/test-format-utils.R`, 9 cases. The awkward part is that **a UTF-8 session cannot see this bug at all**, so an ordinary test would have passed before and after and proved nothing. Two devices instead:

1. **A subprocess forced into `LC_ALL=C`**, which resolves all eleven symbols and checks currency *detection* separately. Skips rather than fails if a subprocess cannot be started.
2. **A static case that reads `R/` through the parser** and rejects any `\uXXXX` escape in tag position. This bites in *any* locale, and it is the one that matters going forward: a future `list("€" = ...)` written on a UTF-8 machine would pass every behavioural test here while being broken for exactly the readers this repo exists for.

The static case uses `getParseData()` rather than a regex, because `EQ_SUB` is the token R uses for a tag — so the comment above `prefixes` that *discusses* the mistake is a `COMMENT` token and cannot be mistaken for the mistake. It caught that false positive on its first run, which is how I found out I needed the parser.

**Verified they bite:** against the code this replaces, **4 of the 9 fail** — both in-session behavioural cases (`USD` where `EUR` was expected), the forced-locale case, and the static case.

| Gate | Result |
|---|---|
| full suite | **4649 passed**, 0 failed, 0 errors (from 4640) |
| skipped | 97, unchanged — all optional packages (`patchwork`, `quantmod`, `shiny`) |
| non-ASCII literals in `R/format_utils.R` | 0, unchanged — the escapes are kept, only their position moved |

## Related

Closes #139.

Found while chasing the ten `unable to translate '<U+20AC>' to native encoding` warnings that `pkgload::load_all()` emits in a `C` locale. Those warnings are cosmetic — `parse()` reporting it cannot render the escapes for source references — but they were the thread that led here.

---

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,7 +52,8 @@ Suggests:
     quantmod,
     TTR,
     zoo,
-    xts
+    xts,
+    pkgload
 Config/testthat/edition: 3
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2
--- a/R/format_utils.R
+++ b/R/format_utils.R
@@ -146,6 +146,44 @@ extract_from_scales_closure <- function(label_func) {
   build_format_config(format_type, prefix, suffix, accuracy, digits)
 }
 
+#' Declare a String's Encoding as UTF-8
+#'
+#' The currency symbols in this file are written as `\uXXXX` escapes, which
+#' always produce UTF-8 bytes. What varies is whether R *marks* them: in a
+#' UTF-8 session it does, and in a `C` session it leaves them `"unknown"`.
+#' A prefix arriving from `readr`, `jsonlite` or `read.csv(encoding =)` is
+#' marked `"UTF-8"` either way, so in a non-UTF-8 session the two sides
+#' disagree and `match()` -- which translates to the native encoding first,
+#' where a euro has no representation -- misses on identical bytes. Every
+#' non-dollar currency then fell through to the `USD` default, and a euro
+#' axis was announced as dollars (#139).
+#'
+#' Declaring the encoding states what the bytes already are; it does not
+#' convert them. `enc2utf8()` is not a substitute -- it reads an `"unknown"`
+#' string as native and mistranslates these keys.
+#'
+#' @param x A character vector.
+#' @return `x`, marked as UTF-8.
+#' @keywords internal
+#' @noRd
+as_utf8 <- function(x) {
+  Encoding(x) <- "UTF-8"
+  x
+}
+
+#' Currency Symbols Recognised as a Prefix
+#'
+#' Shared by [detect_scales_format_type()] and [prefix_to_currency_code()] so
+#' the test that decides something is a currency cannot drift from the map
+#' that names which one.
+#'
+#' @keywords internal
+#' @noRd
+CURRENCY_SYMBOLS <- c(
+  "$", "\u20ac", "\u00a3", "\u00a5", "\u20b9",
+  "\u20a9", "\u20bd", "\u20aa", "\u20b1", "\u0e3f"
+)
+
 #' Detect Format Type from scales Closure Parameters
 #'
 #' @param prefix Prefix string from closure
@@ -166,10 +204,10 @@ detect_scales_format_type <- function(prefix, suffix, digits, scale, accuracy) {
     return("percent")
   }
 
-  # Currency: common currency symbols as prefix
-  currency_symbols <- c("$", "\u20ac", "\u00a3", "\u00a5", "\u20b9",
-                        "\u20a9", "\u20bd", "\u20aa", "\u20b1", "\u0e3f")
-  if (!is.null(prefix) && (prefix %in% currency_symbols ||
+  # Currency: common currency symbols as prefix.  Both sides are declared
+  # UTF-8 so the comparison does not depend on the session locale -- see
+  # `as_utf8()`.
+  if (!is.null(prefix) && (as_utf8(prefix) %in% as_utf8(CURRENCY_SYMBOLS) ||
       grepl("^[A-Z]{0,2}\\$", prefix))) {
     return("currency")
   }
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both actionable notes taken in `03eb409`. The third I'm leaving, with reasoning.

**The literal `₿`** — right, and the irony is the point: a literal non-ASCII character in the one test file whose subject is what literals do. Escaped to `"₿"`, and I took the prose euros in its comments with it, so the file that enforces the rule now keeps it. Zero literal non-ASCII bytes in the new file.

**`pkgload` undeclared** — this is the better of the two, and worth being precise about the consequence, because it is not "the test fails". It **skips**. The fallback is `library(maidr)`, which cannot load a package that is not installed; the subprocess exits non-zero, `in_locale()` returns `NULL`, and `skip_if()` quietly retires the one case that proves the bug exists. On a plain `devtools::test()` — the common local path — the C-locale case would have stopped running and nothing would have said so.

A test that silently stops running is exactly what the static case was added to prevent, so it would have been poor to leave the behavioural half open to it. Added to `Suggests`.

**`as_utf8()` on the constants is redundant — keeping it.** You're right on the mechanism: an escape in an ordinary string is already marked `"UTF-8"` at parse time in every locale, so `CURRENCY_SYMBOLS` and `prefixes` do not need re-marking. I'm applying it to both sides because the property the code needs is *"this comparison does not depend on the locale"*, and a one-sided call states a weaker thing — that one side happens not to need it, which a reader then has to re-derive from where each escape sits. That distinction is the whole bug, so I'd rather it be visible at the comparison than correct-by-coincidence.

Thanks for independently confirming the tag-position mechanism rather than taking the PR description's word for it — and for checking `.lintr`'s `object_name_linter` against the `CURRENCY_SYMBOLS` naming, which I had matched against `html_dependencies.R` by eye without checking the config.

| Gate | Result |
|---|---|
| full suite | **4649 passed**, 0 failed, 0 errors, 97 skipped |
| `test-format-utils.R` | 9 passed, 0 skipped — the C-locale cases really ran |
| literal non-ASCII in the new test file | 0 |

---

PR #141, opened 2026-08-11
feat(base-r): read a normalized barplot as stacked_normalized_bar

jooyoungseo:
## Description

The base R adapter classified every stacked barplot as `stacked_bar`, so a 100% stacked bar was announced as a plain one and its segments framed as magnitudes rather than as shares of their category. This is the base R half that #138 explicitly left out.

## The part worth reviewing: there is no flag to read

I went looking for the base R equivalent of ggplot2's `position = "fill"` and there isn't one. `barplot.default` takes no normalisation argument at all:

```
height, width, space, names.arg, legend.text, beside, horiz, density, angle,
col, border, main, sub, xlab, ylab, xlim, ylim, xpd, log, axes, axisnames,
cex.axis, cex.names, inside, plot, axis.lty, offset, add, ann, args.legend, ...
```

The idiomatic 100% stacked bar in base R is written by normalising the matrix first — `barplot(prop.table(m, 2))` — and by the time the call is recorded, that normalisation is indistinguishable from data that arrived proportional.

So the classification here reads **what the chart draws** rather than what the author declared. When every column sums to 1, every bar is drawn to a common full height and each segment is that category's share; a chart like that is a 100% stacked bar whatever the numbers were beforehand. That is a weaker footing than the ggplot2 side has, and it is why the tests pin the **refusals** as carefully as the claims.

**Not claimed:**
- Columns summing to **100** — a count matrix can total 100 by coincidence, and nothing in the drawing distinguishes that from percentages. Claiming it would be a guess rather than a reading.
- A **single-row** matrix — one series stacked against nothing is not a stack, and reading it as "100% stacked" would announce a composition with no parts.
- `beside = TRUE` — the dodge check runs first and must.
- A bare **vector** — only a matrix stacks.

Tolerance rather than equality on the sum, because `prop.table()` divides and the columns land near 1 rather than exactly on it. An equality test would reject the very charts this exists to catch; there is a case for that too.

## Changes Made

- `is_normalized_barplot()` on the base R adapter, checked before the plain stacked branch.
- `stacked_normalized_bar` registered in the base R processor factory and `get_supported_types()`, routing to the existing `BaseRStackedBarLayerProcessor`.
- That processor **hardcoded** `type = "stacked_bar"` in its result — unlike the ggplot2 side, where the orchestrator carries the detected type through — so it now reads its own `layer_info$type`.

**No value changes were needed**, which is the other difference from #138. There the `count` column and the pre-rescale `y` both had to be bypassed; here the values already are the drawn shares, because the author normalised before calling `barplot()`. Only the emitted type differs.

## Testing

New file `tests/testthat/test-base-r-normalized-bar.R`, **9 assertions, all passing**.

An earlier draft of this file guarded every case with `skip_if_not(exists("BaseRAdapter"))`, and every one of them silently **skipped** — the internal classes are not visible the way the file was being run. A test that skips proves nothing, so the guards are gone and the file reaches the classes through `maidr:::`; a missing symbol now fails loudly instead of reporting green.

Full suite: `test_dir("tests/testthat")` reports **9 failures, all in the pie tests** — the same pre-existing set #138 documented, unchanged in count and identity by this branch.

## Notes

The branch was reset from `main` after #138 merged, so the push is a force-with-lease over remote history that is already in `main` as the squashed `271ac3d`.

---

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,5 +55,5 @@ Suggests:
     xts
 Config/testthat/edition: 3
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.2
 VignetteBuilder: knitr
+Config/roxygen2/version: 8.1.0
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,5 +55,5 @@ Suggests:
     xts
 Config/testthat/edition: 3
 Roxygen: list(markdown = TRUE)
+RoxygenNote: 7.3.2
 VignetteBuilder: knitr
-Config/roxygen2/version: 8.1.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,17 @@
   follow the type: a filled layer reports the drawn share rather than
   `stat_count()`'s untouched `count` column or the pre-rescale `y` from the
   user's own data frame.
+* 100% stacked bar support for Base R. `barplot()` takes no normalisation
+  argument at all — the idiomatic 100% stacked bar is written by normalising
+  the matrix first, as `barplot(prop.table(m, 2))` — so the classification
+  reads the drawn geometry instead: a stacked barplot whose every column sums
+  to 1 is emitted as `stacked_normalized_bar`. Deliberately narrow, and it
+  claims nothing it cannot read. Columns summing to 100 stay a plain stacked
+  bar, because a matrix of counts can total 100 by coincidence and the drawing
+  does not distinguish that from percentages; a single-row matrix stays one
+  too, since a series stacked against nothing is not a stack. The values need
+  no adjustment here, unlike the 'ggplot2' case above: they already are the
+  drawn shares.
 
 * Added step plot support for 'ggplot2' (`geom_step()`) and Base R
   (`plot(type = "s")`, `plot(type = "S")`, and the `lines()` equivalents).
--- a/R/base_r_adapter.R
+++ b/R/base_r_adapter.R
@@ -130,6 +130,8 @@ BaseRAdapter <- R6::R6Class(
         "barplot" = {
           if (self$is_dodged_barplot(args)) {
             "dodged_bar"
+          } else if (self$is_normalized_barplot(args)) {
+            "stacked_normalized_bar"
           } else if (self$is_stacked_barplot(args)) {
             "stacked_bar"
           } else {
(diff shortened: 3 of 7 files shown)


jooyoungseo:
Both gaps addressed in `6981c91`. The second one did not go the way you suggested, and the reason is worth flagging.

### 1. `NEWS.md`

Added, as a sibling bullet to the ggplot2 entry from #138. You were right that the changelog otherwise advertises 100% stacked bar support for ggplot2 only. The entry states the difference explicitly — no flag to read, so the geometry is read instead, and what it refuses to claim.

### 2. `man/BaseRAdapter.Rd` — regenerated by hand, deliberately

I ran `roxygen2::roxygenise()` first. The result was not usable:

```
47 files changed, 6500 insertions(+), 5578 deletions(-)
```

The repo's `man/` was generated by an **older roxygen2** than the one available here, and the two render differently — method headings (`\subsection{Method \code{f()}}` vs `\subsection{\code{Class$f()}}`), cross-references (`\link[maidr:LayerProcessor]{maidr::LayerProcessor}` vs the unqualified form), and `\preformatted` blocks all change shape. Every `.Rd` in the package gets rewritten.

Worse, it is not confined to documentation. `NAMESPACE` loses entries:

```diff
-importFrom(grid,grid.draw)
-importFrom(grid,grid.newpage)
```

That is a change to what the package imports, not to what it documents, and shipping it inside a feature branch is how a package acquires a runtime break that no test in the diff would explain.

So I reverted all of it and wrote the `is_normalized_barplot()` section by hand, matching the format of the `is_stacked_barplot()` block directly above it — the same output the repo's own roxygen2 would produce. The diff is **38 insertions, one file**, and `tools::checkRd()` passes on the result.

Hand-editing a generated file is not something to do lightly, and I would rather it were regenerated properly. But that regeneration is a toolchain change for whoever holds the pinned roxygen2 version, and it should land on its own where the NAMESPACE diff is the point rather than a side effect. Happy to file it as an issue.

### On the minor notes

- The duplicate `is_stacked_barplot()` call is intentional — it keeps `is_normalized_barplot()` answerable on its own rather than only correct when called from one particular branch of `detect_layer_type`. It is a matrix type check and a `beside` read, so the cost is nil.
- `args[[1]]` positional access matches `is_dodged_barplot`/`is_stacked_barplot` exactly; changing it here alone would leave the three siblings inconsistent. Worth doing to all three at once, not to one.

### Verification re-run

You were right to flag the "9 assertions" as self-reported. Re-run after these changes:

```
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 9 ] Done!
```

Also re-installed the package with docs (`R CMD INSTALL` without `--no-docs`) to confirm the new Rd section builds.

---

jooyoungseo:
Good catch, and it was a miss on my part rather than a judgement call — fixed in `4bafe8a`.

`RoxygenNote: 7.3.2` → `Config/roxygen2/version: 8.1.0` was residue from the `roxygenise()` run I described in the previous comment. I reverted `man/` and `NAMESPACE` after that run and did not think to check `DESCRIPTION`, so the one piece of version churn this branch exists to avoid slipped in through the back door. Reverted; `DESCRIPTION` is now out of the diff entirely.

Which version the package pins is exactly the kind of decision that should be made deliberately, as you say — that is part of what #143 is for, alongside the `NAMESPACE` `grid` imports that the same regeneration drops.

One correction to the summary above: `man/BaseRAdapter.Rd` is **not** regenerated. The section is hand-written to match the format of the `is_stacked_barplot()` block directly above it, for the reasons in [my earlier comment](https://github.com/xability/r-maidr/pull/141#issuecomment-5247955204) — a real `roxygenise()` here rewrites 47 files and 6500 lines and drops NAMESPACE entries. `tools::checkRd()` passes on the result, and the diff is 38 insertions in one file.

Current diff, after the revert:

```
 NEWS.md                                     | 11 ++
 R/base_r_adapter.R                          | 46 +++++++++
 R/base_r_processor_factory.R                |  4 ++
 R/base_r_stacked_bar_layer_processor.R      | 10 ++-
 man/BaseRAdapter.Rd                         | 38 ++++++++
 tests/testthat/test-base-r-normalized-bar.R | 89 ++++++++++++++++++
```

---

Issue #135, opened 2026-08-10
feat: support geom_errorbar, geom_ribbon and geom_pointrange


jooyoungseo:
## Description

None of ggplot2's uncertainty geoms have a layer processor: `geom_errorbar()`, `geom_errorbarh()`, `geom_linerange()`, `geom_pointrange()`, `geom_crossbar()`, and `geom_ribbon()` used as a confidence band. They all fall through to `Ggplot2UnknownLayerProcessor`, so the interval is dropped.

## Why this is the most consequential gap in the R binding

Confidence intervals are the point of most published statistical figures — whether two group means differ is answered by whether their intervals overlap. That comparison is currently unavailable to a MAIDR user reading an R chart.

It also affects a layer that r-maidr *does* support: `geom_smooth()` draws a confidence ribbon by default (`se = TRUE`), and `Ggplot2SmoothLayerProcessor` reads only the fitted line. So the band is being silently discarded on charts that otherwise work today.

## Scope

- [ ] `R/ggplot2_errorbar_layer_processor.R` reading `ymin` / `ymax` (and `xmin` / `xmax`) from the computed layer data
- [ ] Register the geoms in the factory and `get_supported_types()`
- [ ] Attach the ribbon to `geom_smooth()` output rather than dropping it
- [ ] Handle asymmetric and one-sided intervals

## Blocked on

Core interval support: xability/maidr#789.

## Difficulty

Medium.

jooyoungseo:
Partially done in #144 (merged as `49a5d87`) — narrowing rather than closing.

**Landed:** `geom_errorbar()`, `geom_errorbarh()`, `geom_linerange()`, `geom_pointrange()` and `geom_crossbar()` now emit `error_bar` layers, in both orientations, with the estimate aesthetic optional throughout.

**Still open, and what this issue should now track:**

- **`geom_ribbon()`** — a continuous band rather than discrete samples. `error_bar` navigates sample-by-sample, so a ribbon over a dense x needs either a different trace type or a deliberate decision about resampling. Not a variation on what landed.
- **`geom_smooth()`'s confidence band** — structurally different again: the band belongs to an *existing* `smooth` layer rather than forming one of its own, so it means extending that layer's payload rather than adding a processor. Emitting it as a separate `error_bar` layer would give the reader two layers for one drawn curve.

[...]

Issue #143, opened 2026-08-11
chore(docs): man/ was generated by an older roxygen2, so any regeneration rewrites the package


jooyoungseo:
## Description

`man/` is out of step with current roxygen2. Running `roxygen2::roxygenise()` on an unmodified checkout produces:

```
47 files changed, 6500 insertions(+), 5578 deletions(-)
```

None of that is content. The two versions render the same roxygen source differently:

- method headings — `\subsection{Method \code{f()}}` becomes `\subsection{\code{Class$f()}}`
- cross-references — `\link[maidr:LayerProcessor]{maidr::LayerProcessor}` becomes the unqualified form
- `\preformatted` blocks move from inline to multi-line

## Why this needs its own change

It is not confined to documentation. `NAMESPACE` **loses** entries:

```diff
-importFrom(grid,grid.draw)
-importFrom(grid,grid.newpage)
```

That is a change to what the package imports rather than to what it documents, and it is exactly the kind of thing that turns into a runtime failure with no test in the diff to explain it.

The practical effect is that a feature branch adding one documented method cannot regenerate docs: it either ships a 47-file rewrite plus a NAMESPACE change unrelated to its subject, or it hand-writes the one section. Neither is good, and the first is worse.

## What this means today

`man/BaseRAdapter.Rd` in #141 carries a hand-written `is_normalized_barplot()` section, matched to the format of the blocks around it and validated with `tools::checkRd()`. That is a stopgap, and the same choice will face the next person who documents a method.

## Scope

- [ ] Establish which roxygen2 version the current `man/` was generated with, and whether to pin it or move forward
- [ ] Regenerate the whole of `man/` in one commit whose diff is understood to be formatting
- [ ] Confirm the `grid` imports are genuinely unused before letting NAMESPACE drop them — and restore the roxygen tags if they are not
- [ ] Consider a CI check that fails when `man/` drifts from its source, so this cannot silently reopen

PR #144, opened 2026-08-11
feat(ggplot2): read the uncertainty geoms as error bar layers

jooyoungseo:
Addresses the ggplot2 half of #135. The R binding for the `error_bar` trace type, which landed in core as xability/maidr#789 / xability/maidr#819.

`geom_errorbar()`, `geom_errorbarh()`, `geom_linerange()`, `geom_pointrange()` and `geom_crossbar()` all fell through to `Ggplot2UnknownLayerProcessor`, so the interval was dropped — and the interval is usually the finding, since whether two group means differ is answered by whether theirs overlap.

## The trap: both pairs of bounds are present, and only one is data

This is what the processor exists for, and I found it by probing rather than assuming. A vertical `geom_errorbar()` computes:

```
  x   y  ymin ymax  xmin  xmax  flipped_aes
  1  4.2  3.8  4.6  0.55  1.45  FALSE
```

`ymin`/`ymax` are the interval. `xmin`/`xmax` are the **cap width** — how wide the little crossbars are drawn. That's a styling parameter, not data at all. Reading the wrong pair produces a perfectly navigable chart describing the styling, which is the kind of wrong that never gets reported as a bug.

Which pair is right follows the layer's orientation, and ggplot2 records that **two different ways**:

| Layer | How orientation is recorded |
|---|---|
| `geom_errorbar(orientation = "y")` | `flipped_aes = TRUE` column in the built data |
| `geom_errorbarh()` | **no `flipped_aes` column at all** — horizontal by construction |

So reading only the flag would classify every `geom_errorbarh()` as vertical and emit its cap *heights* as the interval. Both paths are handled, and both are pinned.

## Categories are announced by name

ggplot2 maps a discrete axis onto integer positions before computing the layer, so the built data carries `1, 2, 3` where the chart draws `control, low, high`. Worse, those positions follow the **scale's** order, not the data's — in the test fixture `low` is drawn third and `high` second — so emitting them wouldn't even read as row numbers. Labels come from the panel's scale, guarded so the lookup doesn't fire on a continuous axis (whose break labels are not an index into anything and would rename every point).

## Emitted shape

`x` is the category and `y` the magnitude in **both** orientations, with the bounds in `yMin`/`yMax` and `orientation` saying which is on screen where. That's what `ErrorBarTrace` consumes — it reads the magnitude as `y`/`yMin`/`yMax` with no orientation branch — so emitting screen-aligned keys would leave a horizontal chart with no interval at all. Pinned by a test asserting the vertical and horizontal payloads agree on every number and differ only in `orientation`.

## Verification actually run

| Step | Result |
|---|---|
| `test_file(test-ggplot2-errorbar-layer-processor.R)` | **25 passing**, 0 skipped |
| Full `test_dir("tests/testthat")` | 4679 passing |

**On the pre-existing failures:** the full suite reports 5 failures / 4 errors in `test-fabricated-selector-guard.R` and `test-ggplot2-pie-layer-processor.R`. I checked whether my adapter change caused them by stashing this branch and re-running those two files on the baseline: **identical 5 FAIL / 4 ERROR without my changes**. They pre-date this work and are untouched by it.

No test skips — the 25 are real assertions, verified as running rather than silently skipping.

## Out of scope, deliberately

`geom_ribbon()` and attaching the confidence band to `geom_smooth()` output are the remaining half of #135. A ribbon is a continuous band rather than discrete samples, and folding it into `geom_smooth()` means modifying an existing layer's payload rather than adding a processor — different enough in shape that it belongs in its own change. #135 should stay open for it.

---


jooyoungseo:
Good catch on the "unrelated-looking diffs" — that was a real mistake on my part, now fixed by rebasing (`78c92cf` → `13e2dfb`).

**What happened:** I branched off my previous working branch rather than off `origin/main`, so this PR carried three earlier base-R commits (the normalized-barplot work) that had *already* been squash-merged as #141. Squashing rewrites the history, so my branch still held the pre-squash originals and the merge base predated them — which made GitHub render `base_r_adapter.R`, `base_r_processor_factory.R`, `base_r_stacked_bar_layer_processor.R`, `man/BaseRAdapter.Rd`, `NEWS.md` and `test-base-r-normalized-bar.R` as changes belonging to this PR.

**They weren't.** I verified the content was byte-identical to main before touching anything:

```
git diff --stat origin/main...HEAD     # three-dot: showed 10 files, 709 insertions
git diff --stat origin/main HEAD       # two-dot:   showed 4 files, 512 insertions
```

The two-dot diff is the real answer — everything outside the four error-bar files already matched `main` exactly. So this was a history artifact, not duplicated or divergent code, and there was never a risk of re-applying or reverting the base-R work.

Rebased onto `origin/main`, so the PR now contains exactly what it claims:

```
R/ggplot2_adapter.R                                     |  14 +
R/ggplot2_errorbar_layer_processor.R                    | 300 +
R/ggplot2_processor_factory.R                           |   2 +
tests/testthat/test-ggplot2-errorbar-layer-processor.R  | 196 +
```

Re-ran the suite on the rebased state: **25 passing, 0 failed, 0 errored, 0 skipped.**

Worth flagging for the review still in progress: the earlier checklist item about `base_r_*` changes and roxygen/`man/` sync no longer applies — `man/BaseRAdapter.Rd` is not part of this PR, and nothing here adds an exported symbol, so `NAMESPACE` and `man/` are untouched by design.

---

jooyoungseo:
Four findings, all real. Three fixed in 387ffca; the selectors one is filed as #145 with the evidence for why it isn't a one-liner.

## Finding 4 was a bug, and worse than you flagged it

You marked this "Low, needs verification" and couldn't run R. I ran it, and it's **not low** — it also isn't limited to `geom_linerange()`:

```
errorbar  no y -> cols: x, ymin, ymax | has y: FALSE
linerange no y -> cols: x, ymin, ymax | has y: FALSE
```

`geom_errorbar(aes(x, ymin, ymax))` over a `geom_col()` is *the* standard way to draw a bar chart with error bars. It builds with no `y` column, so my `all(c(value_col, category_col) %in% ...)` guard returned `list()` and **the entire layer vanished** — no interval, no estimate, no error. The most idiomatic usage was the one that silently produced nothing.

(`geom_pointrange()` and `geom_crossbar()` genuinely require `y` and error at build time, so those two were never affected.)

Fixed: when the estimate is absent the layer reports the **centre of the drawn span**. That's a property of the bar the chart drew, not a claim about an unobserved estimate — and it's what keeps the bounds, which are the real data here, reachable at all. A layer that *does* carry `y` always uses the drawn value. Both are asserted on the asymmetric sample where the centre (5.3) and the mean (5.1) differ, so a symmetric fixture can't make either look right by accident.

I'd flag that fallback as the one judgment call in this PR: dropping the layer is clearly wrong, and every alternative either invents a different number or leaves the trace with an unmeasured value row. Happy to change it if you'd rather the layer carried bounds only.

## Findings 2, 3 and the `get_available_processors()` nit — done

`man/Ggplot2ErrorbarLayerProcessor.Rd`, a NEWS.md bullet, and the missing entry in `get_available_processors()`.

Worth knowing about the man page: running `roxygenise()` here rewrites `DESCRIPTION`, `NAMESPACE` and **every** `man/*.Rd` — the toolchain drift already filed as #143. So I generated, kept only the new file, and reverted the rest; `git diff DESCRIPTION NAMESPACE` is empty. The `.Rd` is genuinely roxygen-generated, not hand-written.

## Finding 1 (selectors) — real, filed as #145 rather than rushed

You're right that this is a real accessibility gap and that I didn't mention it in the "out of scope" section, because it was an oversight rather than a decision.

I looked at fixing it here and stopped, for a reason worth recording. The inherited `generate_selectors()` matches `geom_point.points`, and these geoms' grob names aren't even consistent with each other:

| Geom | Panel child grob |
|---|---|
| `geom_linerange()` | `geom_linerange.segments.39` |
| `geom_pointrange()` | `geom_pointrange.gTree.78` |
| `geom_crossbar()` | `geom_crossbar.gTree.116` |
| `geom_errorbar()` | **`GRID.polyline.1`** — no geom prefix at all |

The name-matching pattern every sibling uses doesn't reach `geom_errorbar`, the most common of the five. Getting it right means per-geom resolution plus verification against a real gridSVG export — and an unverified selector is *worse* than an absent one, since it highlights the wrong element instead of nothing. #145 has the full scope.

The layer still sonifies, announces and brailles correctly meanwhile; what's missing is the visual highlight sync.

## Also addressed: uneven orientation coverage

Pinned the flipped built-data shape for `geom_pointrange`, `geom_crossbar` and `geom_linerange` rather than assuming it matches `geom_errorbar`'s — which matters precisely because two of them don't inherit `GeomErrorbar`.

**Verification:** 42 passing in this file (was 25), 0 failed / 0 errored / 0 skipped. Full suite 4696 passing (was 4679), with the same pre-existing 5 failures / 4 errors in `test-fabricated-selector-guard.R` and `test-ggplot2-pie-layer-processor.R` that I confirmed earlier are present on a clean `main`.

---

jooyoungseo:
You flagged selectors three times, so I stopped treating it as scoping and went and found out *why* it resists a fix. The answer changes the recommendation, and it's worth having on record before anyone tries the obvious patch.

**`geom_errorbar` renders its entire layer as a single `polyline` grob:**

```
layer_grob class: polyline | name: GRID.polyline.1
  id.lengths: 0
  total vertices: 24        # 3 samples x 8 vertices (bar + two caps)
```

versus a sibling that highlights today:

```
boxplot layer_grob name: geom_boxplot.gTree.27  class: gTree
```

So there is **no per-sample element to address**. Grob naming was the symptom; this is the cause. A `generate_selectors()` override — even a correctly-named one — resolves to *one element covering all three samples*, and MAIDR's `ErrorBarTrace.mapToSvgElements` requires one per point:

```ts
if (flat.length !== this.points.length) {
  return null;
}
```

One element for three points fails that check, the trace returns `null`, and you get the same absent highlight — except the R side now *looks* correct, so the next person to investigate finds a `selectors` key and assumes the fault is elsewhere. **That is strictly worse than emitting nothing**, which is why I'd rather ship the honest absence than the plausible-looking key your suggested patch would produce.

The real fix is one of two things, both design decisions rather than patches: split the polyline into per-sample grobs before export, or teach the JS trace a range-selector concept so one element can stand for a layer. Recorded in [#145](https://github.com/xability/r-maidr/issues/145) with the probe output.

Worth noting `geom_linerange`/`geom_pointrange`/`geom_crossbar` *are* tractable (named `segments`/`gTree` grobs) — but shipping highlighting for four of five geoms and silently not the most common one is its own inconsistency, and deserves a deliberate call rather than being decided by what happened to be easy.

I'm not claiming this makes the gap unimportant. Audio, text and braille all work; the visual highlight sync doesn't, and #145 is open with a concrete scope rather than a vague follow-up.

Everything else you re-checked is confirmed fixed. Thanks for staying on the selectors point — the deferral is better-founded for it, and the trap it uncovered would have cost someone real time.

---

Issue #145, opened 2026-08-11
error_bar layers emit no selectors, so nothing highlights during navigation


jooyoungseo:
## Description

`Ggplot2ErrorbarLayerProcessor$process()` (added in #144) returns `data`, `axes`, `type` and `orientation`, but no `selectors`. Every sibling ggplot2 processor emits them, and `Ggplot2PlotOrchestrator$combine_layer_results()` passes `result$selectors` straight into the emitted layer.

The consequence is narrow but real: an error bar layer sonifies, announces and brailles correctly, but **nothing highlights on the SVG** as the reader navigates. The frontend degrades rather than crashing, so it fails quietly.

## Why it wasn't fixed in #144

The inherited `generate_selectors()` matches `geom_point.points` grobs, which none of these geoms produce — and their grob names are not consistent with each other. Probed on ggplot2 3.4.4:

| Geom | Panel child grob |
|---|---|
| `geom_linerange()` | `geom_linerange.segments.39` |
| `geom_pointrange()` | `geom_pointrange.gTree.78` |
| `geom_crossbar()` | `geom_crossbar.gTree.116` |
| `geom_errorbar()` | **`GRID.polyline.1`** — no geom prefix at all |

So the name-matching approach every other processor uses does not reach `geom_errorbar`, which is the most common of the five. A correct implementation needs per-geom handling plus a way to address the polyline, and the resulting selector strings have to be verified against the actual gridSVG export — a wrong selector is worse than an absent one, because it highlights the wrong element rather than nothing.

That is enough work, and enough verification surface, to belong in its own change rather than being appended to the extraction PR.

## Scope

- Resolve the drawn grob for each of the five geoms, including the unnamed polyline `geom_errorbar()` produces.
- Emit one selector per sample, matching the order the data is emitted in, so the JS trace's `mapToSvgElements` resolves (it requires `flat.length === points.length` and returns `null` otherwise).
- Assert `result$selectors` in `test-ggplot2-errorbar-layer-processor.R`; no test currently reads that key, which is why the gap was invisible.
- Verify against a real SVG export rather than against grob names alone.

Found in review of #144.

jooyoungseo:
Probed further while re-reviewing #144, and the obstacle is worse than grob naming — naming is a symptom, not the cause.

`geom_errorbar` renders its **entire layer as one `polyline` grob**:

```
layer_grob class: polyline | name: GRID.polyline.1
  id.lengths: 0
[...]

PR #147, opened 2026-08-11
feat(ggplot2): read geom_area as the area chart it is

### LINKED ISSUE #134: feat: support geom_area and stacked area plots
## Description

`geom_area()`, `geom_ribbon()` used as an area, and `stat_align()` stacked areas have no layer processor. There is no `"area"` entry in `Ggplot2ProcessorFactory`, so these layers fall through to `Ggplot2UnknownLayerProcessor`.

## Why the stacked case needs care

In a stacked area chart the rendered band top is a cumulative total and the band height is the series' own value. Reading the layer as a line would announce one of those two numbers with no way for the user to tell which — so this wants the dedicated area type rather than a fallback to `line`.

## Scope

- [ ] `R/ggplot2_area_layer_processor.R`
- [ ] Register `"area"` (and the stacked variant) in the factory and `get_supported_types()`
- [ ] Read `ymin` / `ymax` from ggplot2's computed layer data so the band is carried, not just the top edge
- [ ] Announce series value and cumulative total separately when stacked

## Blocked on

Core area trace support: xability/maidr#788.

## Difficulty

Medium.

jooyoungseo:
Closes #134.

`geom_area()` fell through to `Ggplot2UnknownLayerProcessor`, so an area chart carried **no data at all**. It now emits `area`, `stacked_area` or `stacked_normalized_area` — all three of which landed upstream in xability/maidr#788, so this is no longer blocked.

## Why the stacked variants rather than a line fallback

A stacked area draws **two** numbers at each point that a line layer would conflate: the band's *height* is the series' own value, and its *top edge* is the running total. A reader told one of them has nothing to say which they heard. The area trace announces the value and reports the total beside it.

## Two traps in ggplot2's built data, both silent

**`y` is the cumulative top, not the value.** MAIDR's area trace sums the series itself to reach the running total, so emitting `y` would hand it a cumulative number to accumulate *again* — totals growing with the number of series rather than with the data. The value is `ymax - ymin`.

**Most of the rows are not data.** `geom_area()` defaults to `stat = "align"`, which inserts interpolation vertices and closes each polygon on the baseline. A four-year, two-series chart builds **24 rows for 8 observations**:

```
       x      y  ymin   ymax group align_padding
 1999.997  0.000 0.000  0.000     1          TRUE
 2000.000  5.000 2.000  5.000     1         FALSE
 2000.003  5.009 2.003  5.009     1         FALSE   <- not a data point
 2000.997  7.991 2.997  7.991     1         FALSE   <- not a data point
```

Read whole, the chart announces twelve points per series — including a reading of `5.009` at **"year 2000.003"**, a value the data does not hold at an x the chart does not have. `align_padding` does *not* identify them: it marks only the two closing vertices, so it cannot be the filter.

The rows that are data are those whose `x` the layer was given, and a test asserts that filter produces byte-identical output to `stat = "identity"` — because this is a claim about `StatAlign`, which is comparatively new.

## Detection

Sits **before** the line branch, since `GeomArea` inherits `GeomRibbon` inherits `GeomPath`, and stays a `class(...)[1]` comparison for the same reason `GeomStep` does.

It claims only an identity-ish stat. The same geom is how a filled density curve is drawn, and `geom_area(stat = "density")` is a *smooth* whose rows are a computed curve rather than observations. **The existing smooth test caught this** — I had the branch too broad, and `test-ggplot2-smooth-layer-processor.R:736` failed with `"area"` where it wanted `"smooth"`. Its own comment states the rule. Now pinned from both sides.

## A small refactor

`get_own_layer()` moves from `Ggplot2ErrorbarLayerProcessor` to `LayerProcessor`. The area processor inherits the *line* processor, which did not have it — and adding a second copy is how the two would drift. The man-page diff is symmetric: 28 lines out of the errorbar page, 27 into the base.

## Verification actually run

| Step | Result |
|---|---|
| `testthat::test_file` on the new suite | **29 passed** |
| `testthat::test_dir` (whole package) | **1 failure, pre-existing** |
| `tools::checkRd` on every touched page | clean |

The one failure is `test-fabricated-selector-guard.R:146` ("pie emits no selector"). I checked it against a clean checkout with my changes stashed and it fails identically there, so it is not from this branch.

**On `man/`**: `roxygenise()` in this environment rewrites DESCRIPTION, NAMESPACE and every man page — the drift filed as #143. Rather than ship 700 lines of unrelated churn, I reverted all of it and hand-patched `LayerProcessor.Rd` with the moved method, so the diff shows the change and nothing else. Reverting that page instead would have left `Ggplot2ErrorbarLayerProcessor.Rd` linking to a method its page no longer documents.

## Not in scope

`geom_ribbon` is listed on #134 as "used as an area", but its built data is a band around a line (`PositionIdentity`, bounds from the aesthetics) rather than a stacked band, and it is already in scope on #135 with the other uncertainty geoms. Better read there.

---


jooyoungseo:
All four looked at. Three fixed in 4b710eb; the selector one I checked and it deserves its own change, with evidence for why.

## 1 — selectors. Real, and **not** the same situation as the error bar

You asked whether this is a known limitation like #144's. I checked the grobs rather than assuming, and it is the opposite case.

I filed #145 arguing error bar selectors should stay absent, because `geom_errorbar` renders **the whole layer as one polyline grob** — a selector resolves to 1 element for N points, `ErrorBarTrace` discards a selector list whose length disagrees with the data, so a "fix" would be worse than the absence.

`geom_area` renders one grob per **series**:

```
 - geom_area.gTree.10 class: gTree,grob,gDesc
     * geom_ribbon.gTree.4 gTree
     * geom_ribbon.gTree.8 gTree
```

Exactly the granularity `AreaTrace` wants — it extends `LineTrace`, whose multiline highlight needs `selectors.length === data.length`, and that is satisfiable here. So this one is worth doing.

It is not in this PR because the existing machinery does not transfer: `curve_selectors()` matches grid's auto-generated `GRID.polyline.N` names, and `layer_polyline_grobs()` deliberately *skips* geom-named gTrees like `geom_smooth.gTree.N` to avoid miscounting — an area layer is precisely the case that helper excludes by design. It needs its own grob walk, plus the faceted and patchwork paths, with gridSVG naming being the fiddliest part of this package.

**#148** has the grob dump, the sketch, and the note that the smooth processor is the closer precedent than the line one.

## 2 — duplicated `get_layer_built_data`. Fixed, and you were right that it was the same argument

Byte-for-byte identical, panel-scoping included. Hoisted to `LayerProcessor` and both copies removed — the same treatment `get_own_layer()` got in this branch, for the reason this branch already gave. Two copies of a lookup is how they come to disagree about which rows a faceted layer owns.

## 3 — `get_available_processors()`. Fixed

One line. A list that omits a processor the factory can build is wrong about the package, even if only introspection reads it.

## 4 — the discrete-axis edge case. Not minor, and fixed

You flagged this as low confidence and did not request a change without a repro. Having looked at it, the failure mode is real: `seq_along(unique(values))` gives `1..k` for the *k values present in the data*, while ggplot2 assigns positions from the **scale's** levels. A factor carrying an unused level shifts every position, the membership test then matches the wrong rows, and the chart silently describes interpolation vertices as data — which is the bug this processor exists to prevent.

Rather than reconstruct the positions more carefully, it no longer reconstructs them at all. `source_x_values()` answers NULL for a discrete axis, and the caller uses a rule that needs no reconstruction: whatever positions the scale chose, **they are whole numbers, and `StatAlign`'s inserted vertices are not**. Exact, and shorter than the guess it replaces.

`test_that("a discrete x keeps its categories and drops the padding")` uses a factor with an unused `Q4` level — the shape that would have drifted.

## Not evaluated → now evaluated

You noted you could not run R. For the record: `testthat::test_file` **33 passed**; `test_dir` whole-package has **1 failure, `test-fabricated-selector-guard.R:146`, which I verified fails identically on a clean checkout with this branch stashed**; `tools::checkRd` clean on all three touched pages.

---

jooyoungseo:
You were right, and it is a better catch than the round it replaced. Fixed in `ad28910`.

## The gap was exactly where you put it

`source_x_values()` had four ways to return NULL, and only one of them meant "discrete":

| reason | axis is | whole-number rule is |
|---|---|---|
| `values` is not numeric | discrete | exact |
| `built$plot` is NULL | unknown | a guess |
| data absent or empty | unknown | a guess |
| `mapped_column()` resolved no column | **continuous, commonly** | **wrong** |

The last row is the one that bites, and your example is the reproduction: `aes(x = year / 2)` resolves to the label `"year / 2"`, matches no column, and falls straight into the discrete branch. `2000 / 2` is 1000 and whole; `2001 / 2` is 1000.5 and is not. So the filter kept alternate rows and dropped the rest — half the chart, with nothing in the output to say anything had gone missing. Worse than the drift the previous round fixed, because that one at least announced itself as a wrong count.

And it did contradict the docstring two methods up, which still promised "a layer whose x values cannot be recovered keeps every row" — the invariant survived the rewrite in prose and not in code.

## The fix

`source_x_values()` now answers the three cases separately rather than folding two of them together:

```r
list(kind = "numeric", values = unique(values))  # filter by membership
list(kind = "discrete")                          # the whole-number rule
NULL                                             # could not read it -- keep every row
```

I took the sentinel option rather than the test-only option because the test would have pinned the wrong behaviour. The whole-number rule is not a lenient fallback that happens to be imperfect; it is a rule about a different kind of axis, and there is no reading of "keep every row" under which dropping the fractional half of a continuous axis is the intended outcome.

`drop_alignment_vertices()` returns early on NULL now, so the promise in its own documentation is the code path.

## Test

`test_that("a continuous x whose values cannot be read keeps every row")` builds exactly your `aes(x = year / 2)` case and asserts three things: that `source_x_values()` really does answer NULL for it, that the row count survives `drop_alignment_vertices()` unchanged, and — the part that makes the test non-vacuous — that the layer genuinely contains non-whole-numbered rows, so the old code would have dropped them. It fails on the previous commit.

## Selectors

Unchanged from my last reply and still deliberate: #148 has the grob dump showing `geom_area` renders one ribbon gTree per series, which is the granularity `AreaTrace` needs, so it is worth building — but `curve_selectors()`/`layer_polyline_grobs()` do not transfer (that helper skips geom-named gTrees by design, which is precisely what an area layer is), and it needs the faceted and patchwork paths too. Separate change, filed with the evidence.

## Test results

`test_file` on the area suite: **36 passed, 0 failed**. `tools::checkRd` clean on the amended page.

Whole-package `test_dir`: **9 failures, all pre-existing** — I ran it again with this branch stashed and got the identical set (4 in `test-fabricated-selector-guard.R`, 14 across `test-ggplot2-pie-layer-processor.R`). Correcting my previous comment, which named only the `fabricated-selector-guard` one: the pie failures were there too and I under-reported them. They are `subscript out of bounds` on gridSVG selector lookups and unrelated to this branch — pass count goes 4738 → 4741 with these changes, nothing moves the other way.

---

PR #149, opened 2026-08-11
fix(ggplot2): read the layer index from the field production writes

### LINKED ISSUE #148: feat(ggplot2): area layers emit no selectors, so nothing highlights during navigation
## What

`Ggplot2AreaLayerProcessor` (#147) returns `data` / `title` / `axes` / `type` and no `selectors`, so an area chart's bands do not highlight as a reader navigates. Text, audio and braille all work.

## Unlike #145, this one is worth fixing

I filed #145 saying error bar selectors should stay absent, because `geom_errorbar` renders the **whole layer as one polyline grob** — a selector would resolve to 1 element for N points, and `ErrorBarTrace` discards a selector list whose length disagrees with the data, so the "fix" would be worse than the absence.

`geom_area` is the opposite shape. Inspecting the panel grobs of a two-series stacked area:

```
 - geom_area.gTree.10 class: gTree,grob,gDesc
     * geom_ribbon.gTree.4 gTree
     * geom_ribbon.gTree.8 gTree
```

**One ribbon gTree per series** — exactly the granularity the consumer wants. `AreaTrace` extends `LineTrace`, whose multiline highlight needs `selectors.length === data.length`, and that is satisfiable here.

## Why it is not in #147

The existing selector machinery does not transfer. `Ggplot2LineLayerProcessor$curve_selectors()` matches on grid's auto-generated `GRID.polyline.N` names and counts curves inside a single polyline grob — see `layer_polyline_grobs()`, which explicitly *skips* geom-named gTrees like `geom_smooth.gTree.N` to avoid miscounting. An area layer is the case that helper excludes by design.

So this needs its own grob walk: find the layer's `geom_area.gTree`, take its `geom_ribbon.gTree` children in order, and build a selector per series from the gridSVG ids. That is a self-contained piece of work with its own failure modes — gridSVG naming is the fiddliest part of this package, judging by the volume of comment the line processor devotes to it — and it does not belong bolted onto the change that made the layer readable in the first place.

## Sketch

- Walk the panel grob for the layer's own `geom_area.gTree` (the `geom_grob_prefix()` helper already scopes a processor to its own tree, and the smooth processor is the closer precedent than the line one).
- One selector per `geom_ribbon.gTree` child, in draw order, which is series order.
- Guard on the count: emit nothing rather than a list the consumer will discard, which is what `curve_selectors()` already does by returning NULL on a mismatch.
- Worth checking the faceted path too — `generate_selectors()` takes a `grob_id` for facets and a `panel_ctx` for patchwork, and both need the same treatment.

Raised by review on #147.

jooyoungseo:
Closes #148, and fixes a bug found while building it that is more serious than the issue.

## The bug first, because it is the important half

**`geom_area()` and every error bar geom emit no data at all.** Both are merged and both are inert.

```
area/stack -> type: area   series: 0   selectors: 0
errorbar   -> type: error_bar   points: 0
```

`layer_info` is constructed in six places — the orchestrator, the facet utils, the patchwork utils — and every one of them names the field **`index`**. `get_layer_index()` reads that. But `get_layer_built_data()` and `get_own_layer()`, the two helpers hoisted to `LayerProcessor` in #144 and #147, read **`layer_info$layer_index`**, which nothing sets. Both return `NULL` for every processor the orchestrator builds.

The consequences ran deeper than an empty array:

| helper | returns | so |
|---|---|---|
| `get_layer_built_data()` | `NULL` | no rows → no series |
| `get_own_layer()` | `NULL` | `resolve_area_type()` sees no position → a filled chart never reports as normalized |

After the fix, through the same path:

```
area/stack -> type: stacked_area              series: 2  selectors: 2  values: 3, 5, 4, 7
area/fill  -> type: stacked_normalized_area   series: 2  selectors: 2  values: 0.6, 0.625, …
errorbar   -> type: error_bar                 points: 3
```

### Why 4,700 tests missed it

Every unit test constructs its own `layer_info`, and the ones covering these helpers supplied `layer_index` — so the lookup succeeded in the tests and failed in the product. The bug lives exactly in the gap between what the tests build and what the orchestrator builds, and no unit test can close that gap however thorough it is.

So `tests/testthat/test-ggplot2-layer-index-key.R` renders a chart through `save_html()` and reads the emitted payload. Four of its tests fail on the parent commit. One of them is the unit-level statement of the same thing, so a future failure says *which key disagreed* rather than only that a chart came out empty.

I also corrected a test I added in #147 that constructed `layer_index` alone: after this change it passed for the wrong reason — no layer resolved, rather than an unreadable axis — which is a vacuous test, not a passing one.

## Now #148: area selectors

`geom_area` draws each series as its own ribbon holding a filled polygon:

```
 - geom_area.gTree.10
     * geom_ribbon.gTree.4  ->  GRID.polygon.1, GRID.polyline.2
     * geom_ribbon.gTree.8  ->  GRID.polygon.5, GRID.polyline.6
```

One polygon per series, which is the granularity `AreaTrace` needs — it extends the line trace, whose multi-series highlight requires `selectors.length === data.length`. Verified against the rendered SVG rather than assumed: gridSVG emits `GRID.polygon.1.1` and `GRID.polygon.5.1`, and the generated selectors resolve to them.

The **polygon** rather than the outline polyline each ribbon also draws: the band is the mark the reader is being pointed at, and outlining it would highlight a hairline around the shape instead of the shape.

The existing machinery could not be reused, as #148 predicted: `curve_selectors()` counts curves inside one auto-named polyline grob, and `layer_polyline_grobs()` deliberately skips geom-named gTrees to avoid miscounting — an area layer being exactly the shape that rule excludes.

A count that disagrees with the data **withdraws the selectors entirely**. A short or mispaired list is worse than none: the consumer cannot distinguish it from a correct one, and a highlight on the neighbouring band tells the reader the wrong thing about every value it announces.

## One shared method rather than a third copy

`find_layer_grob_tree()` moves from `Ggplot2SmoothLayerProcessor` to `LayerProcessor`. Same argument that moved `get_own_layer()` and `get_layer_built_data()` there, made twice in #147's review. What differs between callers is only *which* layer they want, so that is an argument — the smooth processor passes its resolved index because it may describe a layer other than its own, and everything else takes the default.

## Verification

| Step | Result |
|---|---|
| `test_file` on the area suite | **47 passed** |
| `test_file` on the new key suite | **13 passed**, and **8 failures** on the parent commit |
| `test_dir` whole package | **9 failures, all pre-existing** — identical set with this branch stashed |
| `tools::checkRd` on all four touched pages | clean |

Package-wide passes go **4,741 → 4,765**; nothing moves the other way. The 9 pre-existing failures are 4 in `test-fabricated-selector-guard.R` and 14 assertions across `test-ggplot2-pie-layer-processor.R`, all `subscript out of bounds` on gridSVG selector lookups and unrelated to this branch.

**On `man/`**: `roxygenise()` here still rewrites DESCRIPTION, NAMESPACE and all 48 pages — the drift filed as #143. Rather than ship 6,700 lines of churn, I generated the pages, spliced in only the four subsections that changed, and reverted the rest, so the diff shows the change and nothing else.

---

https://claude.ai/code/session_015TFhhzxcMetSHV7z9NCrJ8

---


jooyoungseo:
CI is green on all four platforms now — `withr` was the whole of it.

Worth recording how it hid, because the shape is reusable. The four `R-CMD-check` jobs failed while their own testthat run reported `[ FAIL 0 | WARN 21 | SKIP 3 | PASS 5735 ]`, and the tail of every job log showed nothing but those warnings and a successful artifact upload. The failing step ran *before* the ones I was reading:

```
  '::' or ':::' import not declared from: 'withr'
...
Status: 1 WARNING
Error: R CMD check found WARNINGs
```

`error-on: "warning"` makes that fatal. It was invisible from three directions at once: the tests passed (`withr` is installed in CI transitively), the local check I ran used `--no-manual` without `--as-cran`, and the job log's *tail* is the always()-guarded reporting steps, not the check. I had been chasing the hand-patched `man/` pages, which were fine.

Took the `index < 1` observation in `054cda3`. You are right that it is unreachable — every production site assigns the index from `seq_along()` — so this is consistency, not a fix. The argument for doing it anyway is that `find_layer_grob_tree()`, added in this same branch, checks `target < 1L`, so the branch shipped three guards that validate an index and one that means something different by it. That difference is what the next hand-built `layer_info` finds.

Local suite: `FAIL 9`, **identical with and without the change**, all nine in the pie processor tests and all from grid version drift in this container rather than the package — CI reports `FAIL 0`.

---

PR #150, opened 2026-08-12
fix(ggplot2): highlight the interval an error bar layer draws

### LINKED ISSUE #145: error_bar layers emit no selectors, so nothing highlights during navigation

jooyoungseo:
## Description

Closes #145.

`Ggplot2ErrorbarLayerProcessor$process()` returned `data`, `axes`, `type` and `orientation` and no `selectors`, so all five uncertainty geoms sonified, announced and brailled correctly and **highlighted nothing**. `ErrorBarTrace.mapToSvgElements` requires the flattened match to be exactly as long as the layer's data and discards anything else, so the frontend degraded quietly rather than erroring.

## The part worth reviewing: the five geoms do not draw alike

The issue predicted this and it turned out to be the whole of the work. Probed against real `gridSVG::grid.export()` output on ggplot2 3.4.4:

| geom | drawn as | elements per sample |
|---|---|---|
| `geom_linerange()` | `geom_linerange.segments.N` | 1 |
| `geom_pointrange()` | a gTree holding that same `segments` grob **and** a `points` grob | 1 |
| `geom_crossbar()` | a gTree holding a `polygon` (the box) **and** a `segments` (the middle line) | 1 |
| `geom_errorbar()` / `geom_errorbarh()` | **`GRID.polyline.N`** — no geom prefix at all | **3** |

Both of those last two columns are why the inherited `generate_selectors()` could not be reused. It matches `geom_point.points` by name, which reaches none of these; and a name-prefix search reaches `geom_errorbar()` least of all, which is the most common of the five.

**The three-per-sample fact is the one that would have bitten silently.** `geom_errorbar()`'s eight points per sample are `cap, NA, whisker, NA, cap`, and grid breaks a polyline at a missing point, so the export carries three `<polyline>` elements for one bar:

```
<g id="GRID.polyline.1.1">
  <polyline id="GRID.polyline.1.1.1a" points="90.9,232.32 121.09,232.32"/>   <- cap
  <polyline id="GRID.polyline.1.1.1b" points="106,232.32 106,154.11"/>       <- whisker
  <polyline id="GRID.polyline.1.1.1c" points="90.9,154.11 121.09,154.11"/>   <- cap
  ...
```

Addressing the group's children directly resolves to 9 elements for 3 samples, the trace discards the lot, and the layer highlights nothing — the same outcome as emitting no selectors at all, reached by a route that looks like it worked.

## Changes Made

`generate_selectors()` on the errorbar processor, plus five helpers it decomposes into:

- **`find_interval_grob()`** — the layer's own tree by name where ggplot2 gives one, then the child that spans the interval: the box for a crossbar, the whisker for a pointrange. The middle line and the estimate are also one per sample, so this decides which mark lights up, not whether anything does.
- **`find_unnamed_interval_grob()`** — position for the two geoms with no name. ggplot2 lays a panel out as the grill, a leading `zeroGrob`, **one child per layer in layer order**, a trailing `zeroGrob` and the border, and a layer that draws nothing still takes its slot as a `zeroGrob`. Deliberately narrow: it is gated on the geom being `GeomErrorbar`/`GeomErrorbarh` and on the resolved grob being a `polyline`, because a positional hit on a neighbouring layer would highlight another layer's marks — worse than the missing highlight this fixes, since a reader can hear nothing but cannot hear wrongness.
- **`interval_grob_shape()` / `grob_point_groups()` / `drawn_run_count()`** — the sample count and the stride, computed from the grob's `id` and its missing points rather than assumed. A sample count that disagrees with the emitted data returns no selector, and so does a stride other than the two checked against an export.

One incidental finding: the leading blank is found by **class**, not by an absent name. A `zeroGrob` is named, and its name is the four characters `"NULL"`, so `is.null(g$name)` matches nothing and the first draft of the positional lookup silently found no anchor.

`man/Ggplot2ErrorbarLayerProcessor.Rd` is regenerated **for this one class only**. Running `roxygen2::roxygenise()` on the package rewrites 53 files including `DESCRIPTION` and `NAMESPACE` — the installed roxygen2 is 8.1.0 against the pinned `RoxygenNote: 7.3.2`, which is exactly #143. I generated into a scratch copy, checked the diff for this file was additive (248 insertions, 2 deletions — both of them correct: `Ggplot2PointLayerProcessor$generate_selectors` leaves the inherited list because it is now overridden), copied that file alone, and confirmed `tools::checkRd()` is clean. The wholesale rewrite stays #143's.

## Testing

**11 new cases in `tests/testthat/test-ggplot2-errorbar-layer-processor.R`** covering the selector shape per geom, the unnamed-grob resolution, the stride, which child a pointrange and a crossbar address, an error bar as the second layer beside a `geom_col()`, an empty layer beside it, and the two refusals (no gtable, no points).

**A new `tests/testthat/test-errorbar-selectors.R`** renders through `save_html()` and asks the exported SVG. This is the part the issue asked for, and it is the part that matters: the processor's own tests assert strings, and a string that resolves to nothing passes every one of them.

- every geom's selector finds **exactly** 3 elements for 3 samples;
- an error bar's elements are **vertical** and a horizontal one's are **horizontal**, which separates the whisker from the caps geometrically rather than by trusting the stride;
- the elements arrive in the order the data was emitted — the fixture's intervals are 0.8, 2.6 and 0.3 wide, so the drawn lengths pin the pairing and a shuffle fails;
- an error bar drawn over a `geom_col()` addresses polylines, not the bars.

**Verified they bite.** With `selectors` removed from the emitted payload again: **21 failures** in the processor file and **6** in the export file. The export helper reports a missing selector as a missing selector rather than as an XPath error, so the failure stays readable.

**Verified against a real CSS engine.** `selectr` is not a dependency, so the export tests translate the two selector shapes to XPath by hand. Rather than trust that, I ran the emitted selectors through `querySelectorAll` in jsdom on the rendered HTML — the same API the browser uses:

```
errorbar    points=3 matched=3 OK  tags=[polyline]  g#GRID\.polyline\.1\.1 > *:nth-child(3n+2)
linerange   points=3 matched=3 OK  tags=[polyline]  g#geom_linerange\.segments\.51\.1 > *
pointrange  points=3 matched=3 OK  tags=[polyline]  g#geom_linerange\.segments\.100\.1 > *
crossbar    points=3 matched=3 OK  tags=[polygon]   g#geom_polygon\.polygon\.152\.1 > *
errorbarh   points=3 matched=3 OK  tags=[polyline]  g#GRID\.polyline\.201\.1 > *:nth-child(3n+2)
col_eb      points=3 matched=3 OK  tags=[polyline]  g#GRID\.polyline\.250\.1 > *:nth-child(3n+2)
```

The errorbar matches came back as `116.61,111.75 116.61,46.57` / `410.39,274.68 410.39,62.87` / `263.5,339.86 263.5,315.42` — vertical, and 65.18 / 211.81 / 24.44 px long against intervals of 0.8 / 2.6 / 0.3, a constant 81.5 px per unit. Right elements, right order.

Full suite: `test_dir("tests/testthat")` reports **4841 passing and 9 failures, all in the pie tests** — the same pre-existing set #138 and #141 documented. Confirmed unchanged by this branch: stashing it and re-running those two files gives the identical 9.

## Notes

Those 9 are not noise, and they are the same defect as this one in a different processor. On ggplot2 3.4.4 a polar bar layer draws a `geom_rect.gTree.N` containing **one `geom_polygon.polygon.N` grob per wedge**, where the pie processor looks for a single grob whose sub-polygons are grouped by id — so `generate_selectors()` returns an empty list and **a pie highlights nothing**. `DESCRIPTION` puts no version bound on ggplot2, and CI runs release, which is why this has stayed invisible. Filed separately rather than folded in here.

The branch was reset from `main` after #141 merged, so the push is a force-with-lease over remote history that is already in `main` as the squashed `4bafe8a`.

---

### DIFF
--- a/.github/workflows/workflow-lint.yml
+++ b/.github/workflows/workflow-lint.yml
@@ -27,10 +27,31 @@ jobs:
 
       # actionlint validates the workflow YAML and shellchecks inline `run`
       # blocks, catching the class of shell/quoting bugs this automation
-      # is prone to. Both the installer script and the release artifact it
-      # downloads are pinned to a tagged version so the bytes fetched here
-      # are reviewable and cannot change out from under the job.
-      - name: actionlint
+      # is prone to. The release artifact is pinned by version AND by digest,
+      # so the bytes are reviewable and cannot change out from under the job.
+      #
+      # It is fetched directly rather than through upstream's
+      # `download-actionlint.bash`. That script's own curl has no `--fail`,
+      # so a transient error response is written to the file and unpacked:
+      # the job then dies on `gzip: stdin: not in gzip format` and reports a
+      # network blip as a lint failure on an unrelated pull request. That is
+      # not hypothetical -- it is what this step did on the run that
+      # prompted the change, having downloaded 107 bytes of error page.
+      # `--fail` turns that into a non-zero exit, `--retry-all-errors` stops
+      # it happening at all, and the checksum then means what it says: a
+      # mismatch is a supply-chain question, not a flaky download. Same fix
+      # as xability/maidr#854.
+      - name: Install actionlint
+        env:
+          ACTIONLINT_VERSION: 1.7.12
+          ACTIONLINT_SHA256: 8aca8db96f1b94770f1b0d72b6dddcb1ebb8123cb3712530b08cc387b349a3d8
         run: |
-          bash <(curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/v1.7.12/scripts/download-actionlint.bash) 1.7.12
-          ./actionlint -color
+          curl --fail --silent --show-error --location \
+            --retry 5 --retry-all-errors --retry-delay 2 \
+            --output actionlint.tar.gz \
+            "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz"
+          echo "${ACTIONLINT_SHA256}  actionlint.tar.gz" | sha256sum --check
+          tar xzf actionlint.tar.gz actionlint
+
+      - name: actionlint
+        run: ./actionlint -color
--- a/R/ggplot2_errorbar_layer_processor.R
+++ b/R/ggplot2_errorbar_layer_processor.R
@@ -54,7 +54,7 @@ Ggplot2ErrorbarLayerProcessor <- R6::R6Class(
     #' @param grob_id Grob ID for faceted plots (optional)
     #' @param panel_id Panel ID for faceted plots (optional)
     #' @param panel_ctx Panel context for panel-scoped selectors (optional)
-    #' @return List with data, axes, type and orientation
+    #' @return List with data, selectors, axes, type and orientation
     process = function(plot,
                        layout,
                        built = NULL,
@@ -270,16 +270,12 @@ Ggplot2ErrorbarLayerProcessor <- R6::R6Class(
     interval_grob_shape = function(grob) {
       kind <- class(grob)[1]
 
-      if (kind %in% c("segments", "points")) {
-        n <- if (identical(kind, "segments")) {
-          length(grob$x0)
-        } else {
-          length(grob$x)
-        }
+      if (identical(kind, "segments")) {
+        n <- length(grob$x0)
         if (n < 1L) {
           return(NULL)
         }
-        return(list(samples = as.integer(n), per_sample = 1L))
+        return(list(samples = n, per_sample = 1L))
       }
 
       if (!kind %in% c("polygon", "polyline")) {
--- a/R/ggplot2_errorbar_layer_processor.R
+++ b/R/ggplot2_errorbar_layer_processor.R
@@ -69,10 +69,14 @@ Ggplot2ErrorbarLayerProcessor <- R6::R6Class(
 
       layer_data <- self$get_layer_built_data(built, panel_id)
       is_horizontal <- self$is_horizontal_layer(plot, layer_data)
+      data <- self$extract_interval_data(
+        built, layer_data, is_horizontal, panel_id
+      )
 
       list(
-        data = self$extract_interval_data(
-          built, layer_data, is_horizontal, panel_id
+        data = data,
+        selectors = self$generate_selectors(
+          plot, gt, grob_id, panel_ctx, length(data)
         ),
         axes = self$extract_axes_labels(plot, built, panel_id),
         type = "error_bar",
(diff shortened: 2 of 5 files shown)


jooyoungseo:
A second, unrelated commit is on this branch, and it is here because the CI check it fixes went red on this PR's first run.

**`Lint workflows and shell scripts` failed on a network blip, dressed as a lint failure.** The step bootstrapped upstream's `download-actionlint.bash`, whose own `curl` has no `--fail`, so the error response was written to the file and handed to `tar`:

```
Downloading https://github.com/rhysd/actionlint/releases/download/v1.7.12/actionlint_1.7.12_linux_amd64.tar.gz with curl
100   107  100   107    0     0    672      0
gzip: stdin: not in gzip format
tar: Child returned status 1
##[error]Process completed with exit code 2
```

107 bytes. Nothing was wrong with the workflows, and nothing on this branch touches one — but the check name says otherwise, and the next person to see it red has no reason to doubt it.

`076e05d` fetches the pinned release directly with `--fail`, `--retry-all-errors` and a sha256, keeping v1.7.12. I verified the digest against the release's own `actionlint_1.7.12_checksums.txt` rather than against my own download:

```
8aca8db96f1b94770f1b0d72b6dddcb1ebb8123cb3712530b08cc387b349a3d8  actionlint_1.7.12_linux_amd64.tar.gz
```

`--fail` is the part that matters as much as the pin. Without it a transient response reaches the *checksum*, which then reports a download problem as a hash mismatch on a pinned release — indistinguishable from the supply-chain attack the pinning exists to catch. This is the same failure xability/maidr#854 fixed, which is how it was recognisable here.

Ran actionlint v1.7.12 locally against the whole of `.github/workflows/` including the rewritten step: **exit 0, no findings.** The existing `shellcheck .github/scripts/*.sh` step already fails loudly if shellcheck goes missing from the runner image, so actionlint's inline `run` shellchecking cannot silently weaken here the way it could in maidr — no separate assertion step was needed.

It would have been its own pull request if the branch policy allowed a second branch. Splitting it into its own commit is the closest available.

---

jooyoungseo:
All three findings taken, in `88dfcde`. The coverage suggestion was the valuable one — I ran the cases before writing them, and they are the cases I would have wanted to know about.

**1. `@return` omitted `selectors`.** Fixed, and the Rd regenerated with it.

**2. The `"points"` branch was dead.** You were right that `find_interval_grob()` only ever returns `polygon`, `segments` or `polyline` — a `points` grob is a pointrange's *estimate*, and the whisker beside it is preferred precisely so that branch is never reached. It was not future-proofing, it was a leftover from an earlier draft that walked the gTree differently. Removed rather than commented: an unreachable branch is how a "verified against a real export" claim quietly stops being true.

**3. The positional lookup, probed rather than reasoned about.** Both cases you named, and one more, on ggplot2 3.4.4:

| case | panel children | result |
|---|---|---|
| two `geom_errorbar()` layers | `grill, zeroGrob, polyline, polyline, zeroGrob, border` | two **different** selectors |
| `geom_errorbar()` + `geom_linerange()` | `grill, zeroGrob, polyline, segments, zeroGrob, border` | `GRID.polyline…` and `geom_linerange.segments…` |
| `theme_void()` | unchanged | resolves |
| `panel.grid = element_blank()` | unchanged | resolves |

Three new tests pin them. The two-errorbar one asserts the selectors *differ*, which is the assertion that matters: had the anchor been counted from the wrong place, both layers would have resolved to the first polyline — and that resolves, and looks healthy, and highlights the wrong layer's whiskers. The theme cases confirm what the anchor actually depends on: it is the leading `zeroGrob`, not the grill, so a theme rendering the background differently does not move it.

**On the second review's follow-up suggestion** — a regression test that fails loudly on a ggplot2 shape change rather than silently — that is already the behaviour, and it is worth being precise about why. `find_unnamed_interval_grob()` degrades to `NULL` if the layout shifts, which is silent *in production*, but `test-errorbar-selectors.R` asserts the selector resolves to **exactly** three elements against a real export. A layout change makes that zero and the suite goes red. That is the difference from the pie processor, whose selector generation had no export-level test at all — which is exactly how xability/r-maidr#151 stayed invisible while the same repo's CI stayed green. A ggplot2-version matrix job would still be worth having, and it belongs with #151 rather than here, since #151 is the case that needs it.

## Verification after the change

- `testthat::test_file("tests/testthat/test-ggplot2-errorbar-layer-processor.R")` — **78 passed**
- `testthat::test_file("tests/testthat/test-errorbar-selectors.R")` — **49 passed**
- `tools::checkRd("man/Ggplot2ErrorbarLayerProcessor.Rd")` — clean

**On the CI at review time:** both of you read it right. `test-coverage` and `ubuntu-latest (oldrel-1)` failed on `wget` to `quarto-dev/quarto-cli` releases — `503 Service Unavailable` on one, twenty consecutive "No data received" on the other — during dependency setup, before any package code ran. Same GitHub-releases outage that produced the 107-byte actionlint body earlier on this PR, which is why `Lint workflows` went green this time while the quarto downloads did not: the retry I added rides through it and `wget`'s exhausts. I have re-run the failed jobs.

---

Issue #137, opened 2026-08-10
feat: plot coverage roadmap — unsupported ggplot2 and base R layers


jooyoungseo:
## Description

Tracking issue for layers r-maidr cannot read. `Ggplot2ProcessorFactory$get_supported_types()` currently lists 13 types; anything outside that list falls through to `Ggplot2UnknownLayerProcessor`.

## Parity gaps

- [ ] #132 — the base R adapter has no violin processor, though the ggplot2 adapter does. Which plotting system a user chose should not decide whether their chart is accessible.
- [ ] #133 — `stacked_normalized_bar` exists in the JS core but r-maidr never emits it, so `geom_bar(position = "fill")` is read as a plain stacked bar. The existing stacked/dodged processors already share the machinery this needs.

## New layer types

- [ ] #135 — `geom_errorbar` / `geom_ribbon` / `geom_pointrange`. **Highest impact item here**: it also affects a layer that already works, since `geom_smooth()` draws a confidence ribbon by default and the current processor reads only the fitted line, silently discarding the band.
- [ ] #134 — `geom_area` and stacked areas
- [ ] #136 — `geom_hex` and `stat_bin_2d` (the `bin_2d` half needs no core work)

## Suggested order

#133 and #132 are self-contained parity work with no upstream dependency. #135 is gated on xability/maidr#789 for the full interval trace, but the `geom_smooth` ribbon loss is worth confirming and scoping independently.

## Related

Core roadmap: xability/maidr#814

jooyoungseo:
Status, and one thing this roadmap turned up that is not a plot type.

**#145 is closed** (#150). The uncertainty geoms drew correctly and highlighted nothing, because `Ggplot2ErrorbarLayerProcessor$process()` emitted no selectors. The five geoms do not draw alike — a linerange is a named `segments` grob, a pointrange and a crossbar are gTrees holding the mark that spans the interval beside another, and `geom_errorbar()` draws an **unnamed** `GRID.polyline.N` with *three* elements per sample. Resolved per geom, verified against a real `gridSVG` export and then through a real `querySelectorAll`.

[...]

PR #152, opened 2026-08-12
fix(ggplot2): find a pie's wedges however ggplot2 drew them

### LINKED ISSUE #151: A ggplot2 pie highlights nothing on ggplot2 3.4.4, because the wedge grob is named differently there
## Description

`Ggplot2PieLayerProcessor$find_polygon_grob()` locates the wedges by matching `geom_rect\.polygon` on a grob name:

```r
find_polygon_grob = function(grob) {
  if (!is.null(grob$name) && grepl("geom_rect\\.polygon", grob$name)) {
    return(grob$name)
  }
  ...
```

On **ggplot2 3.4.4** no such grob exists. A polar bar layer draws a `gTree` named `geom_rect.gTree.N` holding **one `geom_polygon.polygon.N` grob per wedge**:

```
 gTree | panel-1.gTree.26
   gTree | grill.gTree.24
     ...
   zeroGrob | NULL
   gTree | geom_rect.gTree.8
     polygon | geom_polygon.polygon.2
     polygon | geom_polygon.polygon.4
     polygon | geom_polygon.polygon.6
   zeroGrob | NULL
```

So the search returns NULL, `generate_selectors()` returns `list()`, and **a pie chart highlights nothing as the reader navigates**. Everything else about the layer is correct: the wedges sonify, announce and braille, and the payload looks healthy. The failure is silent in exactly the way #145 was for the uncertainty geoms.

## Why this has stayed invisible

`DESCRIPTION` puts **no version bound on ggplot2** — it is a bare `ggplot2` under `Imports:` — while `R-CMD-check.yaml` installs `release`. The comment on `find_polygon_grob()` describes a single grob "whose sub-polygons are grouped by id", which is what newer ggplot2 produces, so CI is green and the older layout is never exercised.

## Reproducing

Nine tests fail on ggplot2 3.4.4, all from this one cause:

```
test-ggplot2-pie-layer-processor.R      7 failures
test-fabricated-selector-guard.R        2 failures
```

The clearest one is `test-ggplot2-pie-layer-processor.R:597`, "selects the polygon grob, not the rect grob" — `selectors` has length 0.

`test-fabricated-selector-guard.R:146` is the one worth reading, because it shows the blast radius rather than the mechanism: it checks that a facet level which drew no wedges gets no selector while the levels that *did* draw still get one, and on 3.4.4 it reports `empty` has length 3 and `drawn` has length 0. **Every** panel comes back unselected, not just the empty one.

These failures are longstanding — #138 and #141 both recorded the same nine and treated them as pre-existing, and #150 confirms the count and identity are unchanged by stashing that branch and re-running the two files.

## Scope

- Resolve the wedges on both layouts: the single id-grouped `geom_rect.polygon.N` and the per-wedge `geom_polygon.polygon.N` children of a `geom_rect.gTree.N`.
- Emitting one element per wedge is the contract — `PieTrace` pairs element *i* with slice *i*, and the per-wedge layout means several grobs rather than one, so the selector list shape changes for that case.
- Verify against a real `gridSVG::grid.export()` rather than against grob names, the same way #150 did. A wrong selector highlights another wedge, which is worse than none.
- Decide whether `DESCRIPTION` should declare a minimum ggplot2 instead. That is the honest alternative if the older layout is not going to be supported — silently drawing a pie that cannot be highlighted is the one outcome nobody chose.

## Notes

Found while running the full suite for #145. It is the same class of defect: the layer reads its data correctly and cannot point at what it drew, and no assertion on the announcement can see it.

jooyoungseo:
## Description

Closes #151.

`Ggplot2PieLayerProcessor$find_polygon_grob()` matched `geom_rect.polygon`, which is one layout of two. On ggplot2 3.4.4 a polar bar layer draws a `geom_rect.gTree` holding **one `geom_polygon.polygon` grob per wedge**, and nothing named `geom_rect.polygon` exists anywhere in the tree — so the lookup found nothing, `generate_selectors()` returned an empty list, and **a pie highlighted nothing at all** while sonifying, announcing and brailling correctly.

`DESCRIPTION` puts no version bound on ggplot2 and `R-CMD-check.yaml` installs `release`, so the older layout was never exercised.

## The part worth reviewing: nine failures nobody read

This is the second half of a story worth recording. #138 and #141 both ran the suite, both reported "9 failures, all in the pie tests", and both correctly called them pre-existing and out of scope. #150 did the same and confirmed the count and identity were unchanged by that branch. Three pull requests observed the same nine failures and none learned what they meant, because "pre-existing" is a true statement that ends the inquiry.

They meant a pie chart could not be highlighted.

## Changes Made

`find_polygon_grob()` splits into two lookups and prefers the one it already had:

- **`find_named_polygon_grob()`** — the original search for a single `geom_rect.polygon.<N>` holding every wedge grouped by id.
- **`find_wedge_container()`** — a `geom_rect.gTree.<N>` whose polygon descendants are the wedges. Verified against real `gridSVG::grid.export()` output:

  ```
  <g id="geom_rect.gTree.8.1">
    <g id="geom_polygon.polygon.2.1"><polygon id="geom_polygon.polygon.2.1.1" .../></g>
    <g id="geom_polygon.polygon.4.1"><polygon id="geom_polygon.polygon.4.1.1" .../></g>
    <g id="geom_polygon.polygon.6.1"><polygon id="geom_polygon.polygon.6.1.1" .../></g>
  </g>
  ```

  Either layout is a container whose `<polygon>` descendants are the wedges in slice order, so **the selector shape is unchanged** — `#<container>\.1 polygon` resolves against both, and only the lookup learns the second one.

- **`holds_polygon()`** gates the gTree branch. A `geom_rect` layer that drew no wedges has nothing to point at, and naming its tree would emit a selector resolving to nothing — which the caller cannot tell apart from a working one. That is the failure mode the existing comment on the empty-list return already argues against, and it applies to the new branch too.

## Testing

**`tests/testthat/test-pie-selectors.R` is new, and it is the test that was missing.** It renders through `save_html()` and resolves the emitted selector against the exported SVG, which is the only kind of assertion that can see this class of defect — the processor's own tests assert a *string*, and a string that finds nothing satisfies every one of them.

- a pie's selector finds exactly one element per wedge;
- the wedges arrive in the order the data was emitted — the areas are 3, 5 and 2 of ten, so the drawn arcs pin the pairing and a shuffle fails;
- a faceted pie scopes each panel to its own wedges, and the two selectors differ. That guards the failure that is *not* an empty list: every panel pointing at the first panel's wedges resolves, and looks healthy.

**`test-ggplot2-pie-layer-processor.R`** gains a case that a `geom_rect` layer drawing no polygons still gets no selector, and its selector-shape assertion now admits both containers rather than naming the one this ggplot2 does not draw.

**Verified they bite:** reverting `R/ggplot2_pie_layer_processor.R` produces **13 failures** across the two files.

n/a — the change is in which element a selector addresses, pinned by test rather than by eye.

## Testing results

| Gate | Before | After |
|---|---|---|
| `testthat::test_dir("tests/testthat")` | 9 failed, 4850 passed | **0 failed, 4881 passed** |
| `tools::checkRd("man/Ggplot2PieLayerProcessor.Rd")` | — | clean |
| line lengths (`.lintr` limit 100) | — | clean |

The nine failures this repo has carried are gone. There are no known failing tests in this suite now.

## Notes

**One of the nine was not a selector problem, and is skipped rather than fixed.** `Ggplot2PieLayerProcessor handles a zero-row layer` fails because ggplot2 3.4.4 cannot *build* the plot — `expand_limits_discrete_trans()` raises `replacement has length zero`, and it reproduces on a bare `ggplot_build()` with no maidr in the picture and with or without the polar coord. There is nothing there for this processor to get right or wrong.

It is skipped by asking the question rather than by naming a version:

```r
built <- tryCatch(suppressWarnings(ggplot2::ggplot_build(p)), error = function(condition) condition)
if (inherits(built, "error")) testthat::skip(...)
```

I have not bisected which ggplot2 fixed it, and a version constant I cannot justify would be a worse record than the condition itself. The skip message names the installed version and the upstream error.

`man/Ggplot2PieLayerProcessor.Rd` is regenerated for this one class only, the same way #150 handled #143's roxygen mismatch: generated into a scratch copy, the single file copied across, `tools::checkRd()` clean. The wholesale regeneration stays #143's.

**Follow-up worth considering, not done here.** Two processors have now shipped a selector lookup that silently found nothing, and in both cases the gap was invisible until someone resolved the selector against a real export. Every processor that emits selectors could carry one export-level case asserting they resolve to the number of marks the layer claims. That is a small file each, and it is the only kind of test that sees this at all — I have noted it on #137.

---

### DIFF
--- a/R/ggplot2_pie_layer_processor.R
+++ b/R/ggplot2_pie_layer_processor.R
@@ -290,20 +290,18 @@ Ggplot2PieLayerProcessor <- R6::R6Class(
         return(list())
       }
 
-      polygon_grob <- NULL
-
+      roots <- list()
       if (!is.null(panel_ctx) && !is.null(panel_ctx$panel_name)) {
         # Facet / patchwork path: scope the search to this panel's grob
         panel_grob <- find_gtable_panel_grob(gt, panel_ctx)
         if (!is.null(panel_grob)) {
-          polygon_grob <- self$find_polygon_grob(panel_grob)
+          roots <- list(panel_grob)
         }
       } else if ("grobs" %in% names(gt)) {
-        for (grob in gt$grobs) {
-          polygon_grob <- self$find_polygon_grob(grob)
-          if (!is.null(polygon_grob)) break
-        }
+        roots <- gt$grobs
       }
+
+      polygon_grob <- self$resolve_layer_polygon_grob(plot, roots)
 
       # No polygon grob means this layer drew no wedges here: an empty facet
       # level, a zero-row layer, a coord that renders no polygons. The layer
@@ -350,7 +348,96 @@ Ggplot2PieLayerProcessor <- R6::R6Class(
     #'
     #' @param grob Grob to search
     #' @return Grob name, or NULL when the tree holds none
+    #' @description Pick the wedge container belonging to *this* layer
+    #'
+    #' A panel can hold more than one polar \code{geom_rect} layer -- two
+    #' \code{geom_col()}s under \code{coord_polar()} is an ordinary way to
+    #' draw a ring over a pie -- and a search that takes the first match
+    #' hands every layer the first layer's wedges. Those selectors resolve,
+    #' and the payload looks healthy, and the outline is on the wrong marks.
+    #'
+    #' \code{LayerProcessor$find_layer_grob_tree()} cannot be reused for
+    #' this: it matches on the geom's own class, and a \code{geom_col()}
+    #' layer is \code{GeomCol} while the grob it draws is named after
+    #' \code{geom_rect}. So the containers are collected in drawing order --
+    #' which is layer order -- and indexed.
+    #'
+    #' When the counts do not line up the answer is no selector rather than
+    #' a guess, for the reason \code{generate_selectors()} already gives: a
+    #' wrong selector highlights another layer's wedges, and the caller can
+    #' tell an empty list from a wrong one where a reader cannot.
+    #'
+    #' @param plot The ggplot2 object, or NULL when the caller has none
+    #' @param roots Grobs to search, in drawing order
+    #' @return Grob name, or NULL
+    resolve_layer_polygon_grob = function(plot, roots) {
+      containers <- unlist(
+        lapply(roots, function(root) self$collect_polygon_grobs(root)),
+        use.names = FALSE
+      )
+      if (length(containers) == 0L) {
+        return(NULL)
+      }
+      if (length(containers) == 1L) {
+        return(containers[[1]])
+      }
+
+      index <- self$get_layer_index()
+      layers <- if (is.null(plot)) NULL else plot$layers
+      if (is.null(layers) || length(containers) != length(layers) ||
+        is.null(index) || index < 1L || index > length(containers)) {
+        return(NULL)
+      }
+      containers[[index]]
+    },
+
+    #' @description Every wedge container in a grob tree, in drawing order
+    #'
+    #' One entry per layer that drew wedges. A match is not descended into:
+    #' the container is the whole layer's wedges, and its children are the
+    #' individual ones.
+    #'
+    #' @param grob Grob to search
+    #' @return Character vector of grob names, possibly empty
+    collect_polygon_grobs = function(grob) {
+      if (is.null(grob)) {
+        return(character(0))
+      }
+
+      own <- self$find_own_polygon_grob(grob)
+      if (!is.null(own)) {
+        return(own)
+      }
+
+      found <- character(0)
+      if ("children" %in% names(grob)) {
+        for (child in grob$children) {
+          found <- c(found, self$collect_polygon_grobs(child))
+        }
+      }
+      found
+    },
+
+    #' @description Whether this grob is itself a wedge container
+    #' @param grob Grob to test
+    #' @return Grob name, or NULL
+    find_own_polygon_grob = function(grob) {
+      if (is.null(grob$name)) {
+        return(NULL)
+      }
+      if (grepl("^geom_rect\\.polygon", grob$name)) {
+        return(grob$name)
+      }
+      if (grepl("^geom_rect\\.gTree", grob$name) && self$holds_polygon(grob)) {
+        return(grob$name)
+      }
+      NULL
+    },
+
     find_polygon_grob = function(grob) {
+      if (is.null(grob)) {
+        return(NULL)
+      }
       named <- self$find_named_polygon_grob(grob)
       if (!is.null(named)) {
         return(named)
@@ -363,7 +450,7 @@ Ggplot2PieLayerProcessor <- R6::R6Class(
     #' @param grob Grob to search
     #' @return Grob name, or NULL
     find_named_polygon_grob = function(grob) {
-      if (!is.null(grob$name) && grepl("geom_rect\\.polygon", grob$name)) {
+      if (!is.null(grob$name) && grepl("^geom_rect\\.polygon", grob$name)) {
         return(grob$name)
       }
 
--- a/R/ggplot2_pie_layer_processor.R
+++ b/R/ggplot2_pie_layer_processor.R
@@ -323,22 +323,53 @@ Ggplot2PieLayerProcessor <- R6::R6Class(
       list(paste0("#", escaped_svg_id, " polygon"))
     },
 
-    #' @description Find this layer's wedge polygon grob within a grob tree
+    #' @description Find the grob whose \code{<polygon>} descendants are the wedges
     #'
-    #' Matches on the \code{geom_rect.polygon} prefix, which the polar grill's
-    #' own polygon (named \code{GRID.polygon.<N>} under \code{coord_radial()})
-    #' does not carry.
+    #' ggplot2 does not draw a polar bar layer the same way across versions,
+    #' and the difference is not cosmetic. Verified against real
+    #' \code{gridSVG::grid.export()} output:
+    #'
+    #' \itemize{
+    #'   \item One \code{geom_rect.polygon.<N>} grob holding every wedge,
+    #'     grouped by id. This is what the lookup was written for.
+    #'   \item A \code{geom_rect.gTree.<N>} holding **one
+    #'     \code{geom_polygon.polygon.<N>} grob per wedge}. On ggplot2 3.4.4
+    #'     this is what a pie draws, and nothing named
+    #'     \code{geom_rect.polygon} exists anywhere in the tree -- so the
+    #'     lookup found nothing, \code{generate_selectors()} returned an empty
+    #'     list, and **a pie highlighted nothing at all** (#151).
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Observation 1 was right, and it is fixed in `8ce2faf` rather than tracked — because you are also right that this PR *widened* the match surface, and "wrong wedges" is worse than "no wedges", which is the whole argument of the change.

**Reproduced first.** Two `geom_col()` layers under one `coord_polar()` — a ring drawn over a pie, an ordinary chart:

```
gTree | geom_rect.gTree.8       <- layer 1
  polygon | geom_polygon.polygon.2 / .4 / .6
gTree | geom_rect.gTree.16      <- layer 2
  polygon | geom_polygon.polygon.10 / .12 / .14

layer 1 -> #geom_rect\.gTree\.8\.1 polygon
layer 2 -> #geom_rect\.gTree\.8\.1 polygon      <- the first layer's wedges
```

**`LayerProcessor$find_layer_grob_tree()` cannot be reused for it,** which I tried first. It matches on the geom's own class, and a `geom_col()` layer is `GeomCol` while the grob it draws is named after `geom_rect` — so it returns NULL for every pie and the search fell straight through to the panel-wide one. Worth recording, since it is the obvious reach and it is a dead end.

So the containers are collected in **drawing order, which is layer order**, and indexed by the layer. After:

```
layer 1 -> #geom_rect\.gTree\.8\.1 polygon
layer 2 -> #geom_rect\.gTree\.16\.1 polygon
```

A single pie — every chart that is only a pie — takes the one-container path unchanged.

**When the counts do not line up, the answer is no selector.** Two containers and three layers means drawing order no longer says which belongs to which, and a positional guess lands on another layer's marks. That is the reasoning the existing comment on the empty-list return already gives, applied to the new branch. Both cases are pinned:

- two polar layers get **different** selectors;
- a panel with two containers and three layers gets `list()`.

**Observation 3** taken in the same commit — `geom_rect.polygon` is anchored now, matching the `gTree` one.

**Observation 2** I agree with and am leaving alone: `man/` does now have a visible split between old- and new-style generated files, and closing it means the wholesale regeneration that is #143's scope. Folding it in here would bury a 30-line fix under a 53-file diff.

On the verification note — same sandbox limitation as #150, and CI has since gone green on all four R-CMD-check platforms for the previous commit. Re-verified locally after this one:

| Gate | Result |
|---|---|
| `testthat::test_dir("tests/testthat")` | **0 failed, 4885 passed** |
| `tools::checkRd("man/Ggplot2PieLayerProcessor.Rd")` | clean |
| line lengths | under 100 |

---

jooyoungseo:
All three taken — `0ce8730` — and finding 2 is one I should have caught myself.

**2. The Rd was corrupted, exactly as described.** The roxygen block documenting `find_polygon_grob()` was stranded when I inserted a method beneath it, so roxygen glued the two together and documented `resolve_layer_polygon_grob(plot, roots)` as taking a `grob`. `tools::checkRd()` reported clean, and it was right to: the Rd is *valid*, it just describes the wrong function. I ran that check and read its "clean" as "correct".

I have a repo-wide test for this exact defect in `xability/maidr` (`test/type/grammarDocs.test.ts`, written after four occurrences there), and I made the same mistake here where it does not exist. So it exists now: **`tests/testthat/test-roxygen-orphans.R`** fails on any roxygen block carrying two `@description` tags. The repo has none — mine was the only one — and the check itself is exercised against temporary fixtures rather than against a real source, so it keeps testing after the sources are clean.

**1. The dead code is deleted.** `find_polygon_grob()`, `find_named_polygon_grob()` and `find_wedge_container()` had no callers left after `8ce2faf`, and you are right that sixty orphaned lines carrying a second, subtly divergent answer to "is this a wedge container" is the shape a future fix patches in one place and misses in the other.

**3. You were right that it was not ambiguous, and the fix is better for it.** Counting containers against `plot$layers` was the wrong measure. ggplot2 lays a panel out as the grill, a leading `zeroGrob`, **one child per layer in layer order**, a trailing `zeroGrob` and the axis tree — so the *slot* is the handle, and a `geom_text()` layer simply occupies one and draws nothing. That is the same fact the errorbar processor already relies on, which I should have reached for first.

Probed on ggplot2 3.4.4:

| chart | before | after |
|---|---|---|
| one pie | container | container |
| two rings | two different containers | two different containers |
| two rings + `geom_text` | **`<none>` for both rings** | container, container, `<none>` |
| `geom_text` then one pie | pie's container for *both* | `<none>`, container |

That last row is the second half of the fix and it is not in your list: the slot is authoritative when it holds **no** container too. Falling through to a whole-panel search there hands a label layer the pie's wedges — resolving, healthy-looking, wrong. Both rows are pinned by tests, and the test that locked the old behaviour in as expected is replaced rather than adjusted.

## Verification

| Gate | Result |
|---|---|
| `testthat::test_dir("tests/testthat")` | **0 failed, 4894 passed** |
| `tools::checkRd("man/Ggplot2PieLayerProcessor.Rd")` | clean, and now describing the right functions |
| line lengths | under 100 |

The regenerated Rd no longer documents the three deleted methods, and `layer_slot_grob(panel)` reads with its own description.

On your note about `Rscript` being blocked in the review sandbox: worth saying that it made no difference to any of the three findings. All were reachable by reading, and the one my local run reported clean is precisely the one that was wrong.

---

jooyoungseo:
CI went red on all four R-CMD-check platforms and on `test-coverage`, and it was mine — `f33b720` fixes it.

```
── Failure ('test-roxygen-orphans.R:64:3'): there are R sources to check ──
Expected `length(r_sources())` > 20L.
Actual comparison: 0 <= 20
[ FAIL 1 | WARN 23 | SKIP 3 | PASS 5854 ]
```

`R CMD check` runs the tests against an **installed** package, where `R/` holds the lazy-load database rather than the sources. So the orphan check found nothing to read, and the guard I put on it — "there are R sources to check", written so the case could not pass vacuously — failed instead. The guard did exactly its job. The test had no business asserting there at all.

It skips when the sources are absent and runs on `devtools::test()` / `testthat::test_dir()`, which is the loop the mistake it catches is made in. Verified by copying the file into a directory with an empty `R/` and running it there: the two source-reading cases skip, the two fixture cases still run, so the check itself stays covered everywhere.

Worth stating plainly: **this check does not run under `R CMD check` or `test-coverage`.** That is a real limitation of testing a source-text property from a package test suite, and I would rather say so than quietly let it look like CI coverage it is not.

One thing the failing logs surfaced that is worth recording: **CI runs ggplot2 4.x**, where `geom_errorbarh()` is soft-deprecated — 23 warnings, no failures. So the two grob layouts this PR handles are both exercised: 3.4.4's `geom_rect.gTree` locally, and 4.x's in CI. That is the confirmation I could not get from one environment.

Local: `testthat::test_dir("tests/testthat")` — **0 failed, 4894 passed**.

---

PR #153, opened 2026-08-13
fix(ggplot2): give a missing bar category the column ggplot2 draws for it

### LINKED ISSUE #112: A bar whose x or fill is NA is drawn as its own category but announced as neither
Raised while addressing a review point on #110, and deliberately left out of that PR — the guard there only stops the case getting *worse*. Measured on `f0d087a`; the numbers below are unchanged by #110 once its guard landed.

## What ggplot2 draws

```r
frame <- data.frame(
  cat  = c("p", "q", NA),
  fill = c("A", "B", "A"),
  val  = c(10, 20, 30),
  stringsAsFactors = FALSE
)
ggplot2::ggplot(frame, ggplot2::aes(x = cat, y = val, fill = fill)) +
  ggplot2::geom_col(position = "dodge")
```

ggplot2 lays out **three** x categories — `p`, `q`, and a grey `NA` — and draws a bar in each. `NA` is a level of the discrete scale, with a tick and a strip label, exactly like the other two. The same holds for an `NA` in `fill`: it gets its own legend key.

## What maidr announces

`x/y/z` triples, `-` where the component is missing:

```
                       drawn   announced
NA in x    (dodge)       3     3   p/10/A   -/30/A   q/20/B
NA in fill (dodge)       3     2   p/10/A   q/20/B
NA in x    (stack)       3     3   p/10/A   -/30/A   q/20/B
NA in fill (stack)       3     2   p/10/A   q/20/B
```

Two distinct wrong answers:

- **`NA` in `x`** — the value survives, but its category is empty. A reader hears the number 30 attached to nothing. There is a bar on screen labelled `NA` and no way to tell that is the one being described.
- **`NA` in `fill`** — the row vanishes. Three bars are drawn, two are announced, and the third is unreachable by any keystroke. No warning, and the payload is internally consistent, so nothing looks wrong from the inside.

The `stat = "count"` path drops the row from the cross-tabulation instead, which is a third behaviour again.

## Why this is not a one-line fix

The natural fix — put `NA` in the level order so it gets a grid column like any other category — collides with what the level order is *for*. Both processors build their axis categories with `sort(unique(...))`, which drops `NA` by definition, and then key cells with `paste(x, fill)`, which stringifies `NA` to `"NA"`. So today:

- a genuine missing value and a **literal string `"NA"`** collapse onto the same cell key, and
- the level vector is `character`, with no way to carry "missing" distinctly from the two characters `N` and `A`.

`tests/testthat/test-ggplot2-facet-na-group.R` already pins the facet-side answer to this question — a facet level that is `NA` is announced as the literal `"NA"`, because that is the string ggplot2 draws on the strip, and a level *spelled* `"NA"` is kept distinct from it. Whatever is decided here should match that, so that the same missing value reads the same way whether it lands on a facet strip or an axis tick.

## What needs deciding

1. **Announce `NA` as the two characters ggplot2 draws** (consistent with the facet answer), and separate it from a literal `"NA"` the way `facet_group_rows()` does — by identity, not by string.
2. Or **announce it through the missing-value channel** the frontend already has (`missingText`, default "missing"), which reads better but does not match the label a sighted reader sees.

(1) matches the existing precedent in this package and the picture on screen. (2) reads more naturally but tells the reader a different word than the one printed on the axis. I lean (1) for that reason, but it is a wire-format decision, not a bug fix, which is why it is filed rather than done.

## Scope

Both bar processors (`ggplot2_dodged_bar_layer_processor.R`, `ggplot2_stacked_bar_layer_processor.R`), for `stat = "identity"` and `stat = "count"`. The Base R side is not affected — `barplot()` takes named heights, so there is no aesthetic to be missing.

Not urgent: an `NA` in a bar chart's grouping column is uncommon, and the frame is already visibly odd. But it is silent, and in the `fill` case a whole bar is unreachable, which is the same class of defect as #94.

jooyoungseo:
# Pull Request

## Description

ggplot2 does not treat a missing value as an absence. Its discrete scale lays out a category for it after the ordinary levels, draws a grey bar there, puts a tick under it and a key in the legend — exactly as it lays out a facet panel for a missing facet value and draws "NA" on the strip.

maidr gave two different wrong answers:

| | drawn | announced | |
| --- | --- | --- | --- |
| `NA` in `x` | 3 | 3 | the value survives, its category is empty — a reader hears the number 30 attached to nothing |
| `NA` in `fill` | 3 | 2 | the row vanishes: three bars drawn, two announced, the third **unreachable by any keystroke** |

The second is the sharper one, and it is silent in the way these always are — the payload is internally consistent, so nothing looks wrong from the inside.

## Related Issues

Closes #112

### The decision the issue asked for

#112 filed this rather than fixing it because it is a wire-format choice: announce the missing category as **the two characters ggplot2 draws** (option 1), or through the frontend's `missingText` channel (option 2). I took option 1, which is the issue's own recommendation, for the reason it gives: `test-ggplot2-facet-na-group.R` already pins that answer for a facet strip, and the same missing value should read the same way whether it lands on a strip or an axis tick. Option 2 reads more naturally but tells the reader a different word than the one printed on the axis.

Verified against ggplot2 rather than assumed — its own label vector for `c("p", "NA", "q", NA)` is `"NA"`, `"p"`, `"q"`, `NA`: the missing category comes **last**, and both it and a level literally spelled `"NA"` get their own tick, both reading "NA".

### New — `R/discrete_level_utils.R`

Three free functions, so a missing value cannot mean one thing on a dodged chart and another on a stacked one. Free functions rather than methods because `facet_group_rows()` is the established shape for exactly this kind of shared rule.

- **`discrete_level_order()`** — the level order, with the missing category last, carried as `NA_character_` rather than the string `"NA"`. That is the identity-not-string rule `facet_group_rows()` already follows, and it is what keeps a column literally containing those two characters as its own separate category.
- **`level_label()`** — the text on the tick. `"NA"` for the missing level.
- **`level_keys()`** — keys a cell so the two cannot collapse. `paste()` stringifies them together, which is why the duplicate check could not see the collision and why the grid had nowhere to put the row.

The missing key is deliberately **not `""`**. R lets `"" %in% names(x)` answer `TRUE` and then throws `subscript out of bounds` on `x[[""]]` — and both processors read their cells through exactly that `%in%`-then-`[[` guard, so the obvious sentinel would have passed the guard and aborted.

### Three things that had to move with the level order

Each is a *wrong* highlight rather than a missing one, which is why none could be left:

- **`reorder_layer_data()`** sorted a missing row last however the categories are laid out. The payload emits the missing category as the **last** series, which the frontend consumes **first** on its reverse per-column walk — so the rect and the cell would have been a whole column apart.
- **`split()`** drops the missing group, so the row-by-row path a duplicated cell falls back to lost the series again.
- **`factor()`** drops the missing level unless told `exclude = NULL`, so `table()` left the `stat = "count"` cell out of the tabulation, and `count_table[NA, ]` is an `NA` subscript rather than a lookup (hence indexing by position).

The stacked processor's `stat = "count"` branch had a fourth: the fill scale's own label for the missing category is `NA`, so the series was emitted with `z: null` — a drawn segment whose group has no name.

## Verification

`tests/testthat/test-bar-na-category.R` runs **the bundled frontend's own rect walk against the rects the pipeline exports**. The question is not whether the payload looks plausible; it is whether the highlight lands on the bar the reader is on. A grid that gained a column but drew it in the wrong place announces every category correctly and outlines the wrong bar, and no assertion on the payload alone can see that.

The walk, for all four cases, on ggplot2 3.4.4:

```
== NA in fill, dodge -- rects: 3 slots: 9 sparse: TRUE direction: reverse
   cell x=p    z=A    y=10  -> rect x=52.60   fill=rgb(248,118,109)
   cell x=q    z=B    y=20  -> rect x=180.05  fill=rgb(0,191,196)
   cell x=r    z=NA   y=30  -> rect x=307.51  fill=rgb(127,127,127)

== NA in x, dodge -- rects: 3 slots: 6 sparse: TRUE direction: reverse
   cell x=p    z=A    y=10  -> rect x=52.90   fill=rgb(248,118,109)
   cell x=q    z=B    y=20  -> rect x=182.39  fill=rgb(0,191,196)
   cell x=NA   z=A    y=30  -> rect x=311.88  fill=rgb(248,118,109)
```

Columns advance left to right, each series' colour matches its rects, and no rect is left unclaimed. The stacked cases are identical.

The test asserts those properties rather than the coordinates: one colour per series and no two series sharing one, rect x monotonic across the walk, every rect claimed. It also reads the level order **off `ggplot_build()`** rather than hard-coding it, so if a future ggplot2 stops putting the missing category last this fails instead of the announcement quietly drifting a column away from its tick.

Also covered: a level spelled `"NA"` beside a missing one stays two categories; a duplicated cell still falls back to row-by-row **and keeps the missing row there**; and an ordinary chart with no missing value at all, as the control.

| Gate | Result |
| --- | --- |
| `testthat::test_dir("tests/testthat")` | **0 failed, 4990 passed** |
| the new file with the fix reverted | **10 failures** |
| `tools::checkRd()` on all four touched/new `.Rd` | clean |
| `test-roxygen-orphans.R` | passes |
| line lengths | under 100 |

The two warnings in the suite are the pre-existing `boxplot(notch = TRUE)` ones in `test-base-r-orchestrator.R`.

### Tests replaced rather than deleted

`test-geom-col-incomplete-frame.R` had two cases pinning the *old* behaviour — added in #110, which deliberately only stopped this getting worse. They are replaced with cases asserting the new behaviour, per the project's rule about not dropping a pin and leaving the weaker assertion behind.

N/A.

## Checklist

- [ ] I have updated the documentation, if applicable.

## Additional Notes

**`man/` is scoped to what changed**, following #150 and #152: the installed roxygen2 is 8.1.0 against `RoxygenNote: 7.3.2`, so a full regeneration rewrites the package. Only `man/Ggplot2DodgedBarLayerProcessor.Rd` (which documented `discrete_level_order()` as a method that no longer exists there) and the three new files are copied from a scratch regeneration. `man/Ggplot2StackedBarProcessor.Rd`'s 212-line diff is pure output-format drift with no method changed, so it is left for #143.

**Base R is unaffected**, as #112 notes: `barplot()` takes named heights, so there is no aesthetic to be missing.

---


jooyoungseo:
Nit taken in `4d2f111`. You are right that the second `discrete_level_order(fill_values)` was redundant — the binding above the griddable test is still live on the fallback path — and there is a reason beyond tidiness to drop it: two computations of the same series list is two things that could stop agreeing, on the one path where a disagreement means a series silently disappears. The comment now says the fallback uses the same `fill_levels` and why.

Re-run after the change: `testthat::test_dir("tests/testthat")` → **0 failed, 4990 passed**.

On the note about not being able to execute `testthat`/`lintr`/`checkRd` — worth saying that the one thing static reading could not have settled here is the part I would most want checked, and it is checked in CI rather than by either of us reading. `test-bar-na-category.R` renders through `save_html()` and resolves the layer's own selector against the exported SVG, so "the announcement names the right category" and "the highlight lands on the right bar" are answered by the rendered document. R-CMD-check is green on Ubuntu release and oldrel; macOS and Windows are still running at the time of writing.

Two things in your read I want to confirm rather than let pass silently, since both are load-bearing:

**`match()`/`%in%` and `NA`.** Correct as you describe it for the level *vectors*, but the reason the cell keys need `level_keys()` at all is that `paste()` does not follow that rule — it stringifies `NA` to `"NA"` before either function ever sees it. So the identity distinction survives in `discrete_level_order()`'s output and dies at the key, which is why there is a separate function for the key rather than a comment saying "careful".

**The `man/` diff.** Slightly different from your reading: the `find_layer_grob_tree`/`get_layer_built_data`/`get_own_layer` entries appearing is roxygen 8.1.0 emitting inherited R6 methods that 7.3.2 did not, so it is version drift rather than catch-up for methods that were missing. The substantive half is the `discrete_level_order` removal, as you say. Either way it is #143's territory and not this PR's.

---

PR #154, opened 2026-08-13
chore(docs): regenerate man/ and NAMESPACE with the roxygen2 that runs today

### LINKED ISSUE #143: chore(docs): man/ was generated by an older roxygen2, so any regeneration rewrites the package

jooyoungseo:
# Pull Request

## Description

`man/` was generated by roxygen2 7.3.2; the installed version is 8.1.0. Regenerating rewrites 47 files, so a branch documenting one method could not regenerate at all — it either carried the whole rewrite or hand-wrote its section. Three pull requests in a row hand-scoped their `man/` changes to avoid it (#150, #152, #153).

## Related Issues

Closes #143

## The issue's central risk does not exist

> It is not confined to documentation. `NAMESPACE` **loses** entries:
> ```diff
> -importFrom(grid,grid.draw)
> -importFrom(grid,grid.newpage)
> ```

That is a line-based diff read as a semantic one. roxygen2 8 groups a multi-symbol import onto several lines:

```
importFrom(grid,
  grid.draw,
  grid.newpage
)
```

so the diff removes the two one-per-line directives and adds the grouped form. Reading only the removals makes it look like a loss.

Parsed both files as R expressions and compared the directives they declare:

```
importFrom  old  12  new  12  identical TRUE
export      old  44  new  44  identical TRUE
S3method    old   0  new   0  identical TRUE
import      old   0  new   0  identical TRUE
```

Nothing is dropped. **This mattered to check rather than assume**: `R/svg_utils.R:85-86` calls `grid.newpage()` and `grid.draw()` *unqualified*, so losing those imports would be `could not find function "grid.newpage"` on every SVG export — a hard runtime failure in the main render path, not a documentation nit.

That resolves scope item 3, and it is why this looked risky enough to sit open.

## Changes Made

**`man/` (45 files), `NAMESPACE`, `DESCRIPTION`** — one `roxygen2::roxygenise()` with the default loader.

**`.github/workflows/docs-drift.yaml`** — regenerates and fails on any difference (scope item 4).

### The `man/` diff is not purely mechanical

Corrected from an earlier version of this description, which said it was. Most of it is:

| | |
| --- | --- |
| method headings | `\subsection{Method \code{f()}}` → `\subsection{\code{Class$f()}}` |
| cross-references | `\link[maidr:X]{maidr::X}` → unqualified |
| `\preformatted` | inline → multi-line |
| `\docType{data}` | dropped — deprecated in R, roxygen2 8 stops emitting it |
| `maidr-package.Rd` | an author holding `c("aut","cph","cre")` is now listed under `Authors:` as well as `Maintainer` — a rendering change, not a missing author restored |

But one class of change is real content. Every R6 subclass's **Inherited methods** list gains `find_layer_grob_tree`, `get_layer_built_data` and `get_own_layer`. Checked rather than assumed:

```
find_layer_grob_tree in old LayerProcessor.Rd            5
find_layer_grob_tree in old Ggplot2BarLayerProcessor.Rd  0
find_layer_grob_tree in new Ggplot2BarLayerProcessor.Rd  1
```

They exist in `R/layer_processor.R` and were already documented on the parent; 7.3.2 was failing to propagate them into subclasses. So it is roxygen2 8 closing a pre-existing doc gap, not a regression and not something introduced here — but a future reader diffing `man/` for real content should know this class of change is in the changeset.

## Why the drift check is pinned, and pinned that way

roxygen2 changes its output between minors — this issue *is* one such change — so an unpinned check goes red on every open pull request the day a new roxygen2 ships, reporting a release as a fault in someone's branch.

It reads the pin from `Config/roxygen2/version` in DESCRIPTION rather than from a version written into the workflow. That field is roxygen2 8's own record of which version last ran, so the pin and the artefact cannot disagree: the regeneration that changes one writes the other. A future upgrade is then one commit.

The job runs `roxygen2::roxygenise()` with **no `load_code` argument**, deliberately. roxygen2's loaders do not all see the same thing; R6 methods are documented from the object it loads, so sourcing the files instead of loading the package would produce a different `man/` and fail on a tree that is correct.

## Verification

| Gate | Result |
| --- | --- |
| the new `docs-drift` job, on this PR | **passed on its first run** |
| `roxygen2::roxygenise()` run twice, second compared to first | **idempotent — no difference** |
| `tools::checkRd()` on all 271 `.Rd` | **0 with issues** |
| `testthat::test_dir("tests/testthat")` | **0 failed, 4990 passed** |
| `actionlint` 1.7.12 on the new workflow | clean |

`R CMD check` (`--no-tests --no-vignettes --no-manual --no-examples`, `_R_CHECK_FORCE_SUGGESTS_=false`), on this branch **and** on an unmodified `origin/main` worktree:

```
* checking package dependencies ... NOTE
* checking R files for syntax errors ... WARNING
* checking files in 'vignettes' ... WARNING
* checking package vignettes in 'inst/doc' ... WARNING
Status: 3 WARNINGs, 1 NOTE
```

Identical on both, so this introduces nothing. All three are environmental in this container — the "syntax errors" WARNING is `OS reports request to set locale to "en_US.UTF-8" cannot be honored`, and the other two are quarto and the unavailable Suggests. Every documentation check passes:

```
* checking Rd files ... OK
* checking Rd metadata ... OK
* checking Rd cross-references ... OK
* checking for missing documentation entries ... OK
* checking for code/documentation mismatches ... OK
* checking Rd \usage sections ... OK
* checking Rd contents ... OK
```

N/A.

## Checklist

- [x] I have updated the documentation, if applicable.
- [ ] I have added appropriate unit tests, if applicable.

## Additional Notes

**Review it by reading the workflow and the NAMESPACE, not the 47 files.** The three things that could hide a fault are the NAMESPACE comparison above, `checkRd` on all 271 files, and the suite. All three are reported here rather than summarised.

**Scope item 1** — "pin it or move forward" — is answered as *move forward*, because there is no reason to hold at 7.3.2 and the drift check makes the current version self-recording rather than a convention someone has to remember.

**`GITHUB_PAT` on the job.** Kept: `R-CMD-check.yaml`, `pkgdown.yaml` and `test-coverage.yaml` all set it, so this matches the repo rather than diverging from it, and `setup-r-dependencies` resolves through pak, which is rate-limited against the GitHub API without a token. DESCRIPTION has no `Remotes:` today, so it is not load-bearing right now — but a job that installs dependencies and is the *only* one without a token is the one that starts failing on a busy afternoon.

**One thing the new job cannot catch**: it compares generated output against sources, so a roxygen block attached to the wrong function produces a *valid* Rd that this job is happy with. `tests/testthat/test-roxygen-orphans.R` covers that separately, and the two are different questions — one asks "is `man/` current?", the other "does it document what it claims to?".


jooyoungseo:
You are right about the inherited methods, and I should not have written "purely mechanical" — I had already worked that out on #153 and then said the opposite here. Corrected in the description, with the check rather than the assertion:

```
find_layer_grob_tree in old LayerProcessor.Rd            5
find_layer_grob_tree in old Ggplot2BarLayerProcessor.Rd  0
find_layer_grob_tree in new Ggplot2BarLayerProcessor.Rd  1
```

Present on the parent, absent from every subclass, present now — so 7.3.2 was failing to propagate, and 8.1.0 closes a pre-existing gap. Your framing is the right one: a future reader diffing `man/` for real content needs to know this class of change is in the changeset, precisely *because* the rest of it is noise they will learn to skim.

Thank you for re-deriving the NAMESPACE directive sets rather than taking the numbers. That is the claim the whole change rests on, and it deserved a second derivation — the failure mode if it were wrong is `could not find function "grid.newpage"` on every SVG export, which is not something a docs PR should be able to cause.

**`GITHUB_PAT`** — keeping it, and it is not cargo-culted, though your instinct to ask was sound. `R-CMD-check.yaml`, `pkgdown.yaml` and `test-coverage.yaml` all set it, so this matches the repo rather than diverging from it, and `setup-r-dependencies` resolves through pak, which is rate-limited against the GitHub API without a token. DESCRIPTION has no `Remotes:` today, so you are right that it is not load-bearing right now. But a job that installs dependencies and is the *only* one in the repo without a token is the one that starts failing on a busy afternoon, for a reason nobody will connect to this PR. Noted in the description so it reads as a decision rather than a leftover.

On not being able to run the regeneration: the one thing static reading could not settle is whether a second `roxygenise()` is a fixed point, and that is now answered twice — once locally (snapshot, re-run, no difference) and once by the new `docs-drift` job, which **passed on its first run**. That job is the durable version of the check you could not perform.

---

Issue #136, opened 2026-08-10
feat: support geom_hex and stat_bin_2d


jooyoungseo:
## Description

`geom_hex()` / `stat_bin_hex()` and `geom_bin_2d()` / `stat_bin_2d()` have no layer processor. Both are the standard ggplot2 answer to an overplotted scatter, and both currently fall through to `Ggplot2UnknownLayerProcessor`.

## Two halves of different cost

**`geom_bin_2d` is nearly free.** A rectangular 2D bin grid is a heatmap, and `Ggplot2HeatmapLayerProcessor` already emits that schema. The work is reading `count` plus the bin edges from the computed layer data.

**`geom_hex` needs core work** for the offset hexagonal lattice, tracked at xability/maidr#799. Worth splitting if the rectangular half lands first.

## Scope

- [ ] `stat_bin_2d` → `heat`, using bin edges as axis coordinates
- [ ] Announce the count and the bin's coordinate range rather than a bin index
- [ ] `geom_hex` once xability/maidr#799 lands

## Difficulty

`bin_2d` low; `hex` gated on core.

jooyoungseo:
Measured before starting, and the premise needs correcting — the `bin_2d` half is not unsupported. It is **wrong**, which is worse.

## `geom_bin_2d` does not fall through to unknown

`geom_bin_2d()` is `GeomTile` + `StatBin2d`, and `Ggplot2Adapter$detect_layer_type()` matches on `geom_class == "GeomTile"`. So it is already classified `heat` and handed to `Ggplot2HeatmapLayerProcessor`.

What comes out:

```
ggplot2 drew:            18 tiles
maidr emitted:           200 x 200 = 40,000 cells
of which non-missing:    0
axes.z.label:            NA
selectors:               1
```

A reader navigating that chart walks a forty-thousand-cell grid in which every cell announces missing, on a colour axis with no name, against eighteen tiles on screen. Falling through to `Ggplot2UnknownLayerProcessor` would have been the honest outcome; this claims to be a heatmap and describes nothing.

## Why

`Ggplot2HeatmapLayerProcessor$extract_data()` assumes a `geom_tile()` heatmap built from **tidy data** — one row per cell, with `x`, `y` and `fill` columns already in `plot$data`. Three assumptions break at once on a computed layer:

1. Axis levels come from `discrete_levels(original_data[[x_col]])`. For a binned scatter that is the 200 raw continuous x values, not the 4 bin columns — hence 200 × 200.
[...]

PR #155, opened 2026-08-13
fix(ggplot2): read a binned scatter's grid from the layer that computed it

jooyoungseo:
# Pull Request

## Description

`geom_bin_2d()` **is not unsupported.** It is `GeomTile` + `StatBin2d`, so `detect_layer_type()` already classifies it `heat` and hands it to `Ggplot2HeatmapLayerProcessor`. That classification is right — a rectangular bin grid is a heatmap. What was wrong is everything that followed, which assumed a `geom_tile()` chart built from tidy data, one row per cell.

Measured on a four-by-four binned scatter of 200 points, before:

```
ggplot2 drew:         18 tiles
maidr emitted:        200 x 200 = 40,000 cells
of which non-missing: 0
axes.z.label:         NA
```

A reader navigating that walks forty thousand cells, each announcing missing, on a colour axis with no name, against eighteen tiles on screen. Falling through to `Ggplot2UnknownLayerProcessor` would have been the honest outcome; this claimed to be a heatmap and described nothing.

After:

```
x labels (5): -2.308 to -1.154 | -1.154 to 0 | 0 to 1.154 | 1.154 to 2.308 | 2.308 to 3.462
y labels (5): 1.385 to 2.769 | 0 to 1.385 | -1.385 to 0 | -2.769 to -1.385 | -4.154 to -2.769

   2  8  2  5  .
  10 37 28  8  1
   7 31 35 10  .
   2  8  4  1  .
   .  1  .  .  .

drawn tiles: 18   emitted cells with a count: 18   counts agree: TRUE
```

## Related Issues

Closes the `bin_2d` half of #136. `geom_hex` is untouched — see the note at the end.

## Two things broke, and together they produced an empty answer

**1. `reorder_layer_data()` destroyed the data before the stat ran.**

It made the first two columns into factors so tiles would sort into DOM order. That is right for a tidy heatmap, whose first two columns *are* the categories. On raw continuous observations it maps 200 data points to 200 **discrete positions**, and `StatBin2d` then bins the integers rather than the numbers — one bin per observation, labelled `0.5 to 1.5`, `1.5 to 2.5`, and so on.

This is the half that took longest to find, because it happens upstream of the processor and looks like a data problem rather than a reordering one. It now returns the frame untouched for a binned layer, whose tiles the stat places rather than the data order.

**2. `extract_data()` rebuilt a grid that was already there.**

For a binned scatter the source columns are the observations, so axis levels came out one per data point, and the built `x` is a bin *centre* (`-0.5538086`) matching none of the raw values. The fill loop ran, matched nothing, and returned a complete `NA` matrix. `fill_col` also fell back to `names(original_data)[3]` — a two-column frame has no third name, hence `NA` on the colour axis.

A binned layer now reads `built_data` directly, which already **is** the grid: one row per drawn tile, carrying `count` and the bin edges.

## Decisions worth review

**Empty cells stay missing, not zero.** An empty bin did count nothing — but the frontend reads a `0` as "no rect here" for highlighting, and every cell of a heatmap has one. Same reasoning the dodged-bar processor records for its `stat = "identity"` branch.

**Bins are named by range, not index.** #136's second scope item, and it comes free: `xmin`/`xmax` are computed alongside the count. "a count of 4" means nothing without "between -2.3 and -1.2", and the range is what a sighted reader takes from the axis. Edges are rounded to four significant figures — the raw value is `-1.1076174999999999`, which is not an announcement anyone can hold in their head.

**The binned path is chosen on `StatBin2d`, not on a `count` column.** A tidy heatmap whose value column happens to be named `count` is not a binned layer and must not take that path. There is a case for exactly this.

## Verification

| Gate | Result |
| --- | --- |
| `testthat::test_dir("tests/testthat")` | **0 failed, 5026 passed** |
| `tests/testthat/test-bin2d-heatmap.R` with the fix reverted | **10 failures** |
| `tools::checkRd()` on the regenerated Rd | clean |
| `roxygen2::roxygenise()` re-run | no drift |
| line lengths | under 100 |

The tests assert the grid **against what `ggplot_build()` actually drew** rather than against a shape chosen here — bin counts, cell counts, and the lowest bin's edge are all read from the built data, so they keep meaning if ggplot2 changes how it places bins.

A `geom_tile()` control is included and is the thing most at risk, since the new branch sits inside the processor every tile heatmap uses.

N/A.

## Checklist

- [ ] I have updated the documentation, if applicable.

## Additional Notes

**`geom_hex` stays out**, and for two reasons rather than one. It is gated on xability/maidr#799 for the offset hexagonal lattice, as #136 says — and separately, `stat_binhex()` requires the `hexbin` package, which is not in `Suggests`, so it currently fails to compute at all. That is a dependency decision as well as a core one.

**I stranded a roxygen block while writing this**, inserting `format_bin_edge` between `bin_labels`' documentation and `bin_labels` itself. `tests/testthat/test-roxygen-orphans.R` — added in #152 after I made the same mistake there — caught it. Recording that because the check earning its keep twice is the argument for having written it.

---


jooyoungseo:
Both points taken in `5454207`, and the first one is a nice case of the source disagreeing with the installed package.

## `value` exists — it is ggplot2 3.x's spelling

You searched current ggplot2 source and found only `count`/`density`/`ncount`/`ndensity`. On the version this repo tests against:

```
ggplot2: 3.4.4
has value column: TRUE
has count column: TRUE
value == count for every row: TRUE
```

So it is not dead code, and it is not a different quantity — which is what justifies the hardcoded `"count"` label you asked about. The comment now says that, and says it was checked rather than assumed.

Worth noting the shape of this, because it has come up twice in this repo now: **CI runs ggplot2 4.x while local development runs 3.4.4** (I established that from the deprecation warnings in #152's logs). Your read of current source and my read of the installed package were both right about different versions. The branch is there for exactly that gap, so a build that stops emitting one column still gets an answer from the other.

## `get_own_layer()` — you are right, and for a better reason than DRY

Refactored. The duplicated bounds check was not just repetition: it was a *second place for the same question to be answered*, and the two could have drifted — which is the failure this very PR is about, arrived at through a different door. `is_binned_layer(plot)` now takes only the plot and reaches the layer through the helper.

The out-of-range case is still pinned, now by constructing a processor whose index names no layer rather than by passing a bad index — which is how it would actually go wrong.

## On `inherits()` vs the literal class name

Left as the literal, and I agree it is a boundary worth naming. A subclassed stat would slip past — but it would slip past into the *tidy* path, which is the safe direction: it would produce today's behaviour rather than a grid read from the wrong place. Matching by inheritance would also catch anything ggplot2 later derives from `StatBin2d` for a lattice that is not rectangular, and that would be the unsafe direction.

## Re-verified after the change

| Gate | Result |
| --- | --- |
| `testthat::test_dir("tests/testthat")` | **0 failed, 5026 passed** |
| `roxygen2::roxygenise()` | no drift |
| line lengths | under 100 |

On not being able to run the suite in the sandbox: the `docs-drift` job added in #154 covers the regeneration half of what you checked by hand, and it has passed on both commits here — including the one where I had to resolve an `.Rd` merge conflict, which is exactly where a hand-resolved regeneration could have gone wrong unnoticed.

---

jooyoungseo:
R-CMD-check was red on Ubuntu (release and oldrel-1), and so was test-coverage. Ten failures, all five render-based cases in `test-bin2d-heatmap.R`, each failing twice — once on `expect_length(raw, 1)` and once on `fromJSON`. The tidy `geom_tile()` control passed.

The cause is upstream of everything this PR is about, and it is not the version split I expected.

**ggplot2 4.0 gave `geom_bin_2d()` a geom of its own.** On 3.x the layer's geom is a plain `GeomTile`; on 4.x it is `GeomBin2d`, a subclass of it. `detect_layer_type()` matches `class(layer$geom)[1]`, so on 4.x a binned scatter is no longer a heatmap — it is `unknown`. An unknown layer is dropped rather than described, so the chart emits **no `maidr-data` attribute at all**.

That makes the 4.x symptom of #136 worse than the one the issue reports, not milder: 3.x announced a 200×200 grid of missing cells against eighteen drawn tiles, and 4.x announces nothing. It also means the tests above were never failing on their subject — they were failing on the layer's absence.

Measured with and without the change, on both versions:

| | 3.4.4 | 4.0.3 |
| --- | --- | --- |
| before | `heat` | `unknown` |
| after | `heat` | `heat` |

The fix adds `GeomBin2d` to the membership test. By name rather than through `inherits()`, to match every other branch in that function and because the symbol does not exist on 3.x. I checked what else moved in that release: `stat_summary_2d()` was left on `GeomTile`, and `geom_hex()` has its own unrelated `GeomHex`, so nothing follows it in.

A regression test now asserts the classification directly, ahead of the render cases, with a note that it is upstream of them.

CI runs 4.x and my local R has 3.4.4, so I installed ggplot2 4.0.3 into a separate library and ran the whole suite against both. Both green — 0 failures on 4.0.3 (where five cases in this file failed before), 0 on 3.4.4. `roxygenise()` produces no drift in `man/`, `NAMESPACE` or `DESCRIPTION`.

My first hypothesis was wrong and worth recording: ggplot2 4.0 does namespace some class names — the backtrace in the failing job shows `ggplot_build.ggplot2::ggplot` — so I expected `class(layer$stat)[1]` to have become `"ggplot2::StatBin2d"` and `is_binned_layer()` to be the break. It has not; it is still `"StatBin2d"`, and that check was fine. The geom subclass was the whole of it.

---

Issue #156, opened 2026-08-13
Every ggplot2 heatmap announces its axes as literally "x" and "y"


jooyoungseo:
`Ggplot2HeatmapLayerProcessor$process()` builds its axes from two string literals:

```r
axes <- build_axes(
  x = "x",
  y = "y",
  z = fill_label
)
```

So the axis names never reach the reader. The colour axis is resolved properly; the other two are not.

```r
frame <- expand.grid(row = c("a", "b"), col = c("x", "y", "z"))
frame$score <- 1:6
p <- ggplot2::ggplot(frame, ggplot2::aes(col, row, fill = score)) +
  ggplot2::geom_tile() +
  ggplot2::labs(x = "Task", y = "Model")
```

emits

```json
{"x": {"label": "x"}, "y": {"label": "y"}, "z": {"label": "score"}}
```

A cell is therefore announced as **"x: y, y: a, score: 3"** — where the first `y` is a category value and the second is the axis name that should have been "Model". Sighted readers see "Task" and "Model" printed on the chart; nobody reading it non-visually does.

An explicit `labs()` is ignored, and so is the mapped column name that ggplot2 falls back to. There is nothing a caller can do about it from their own code.

## Scope

Every ggplot2 heatmap: `geom_tile()`, `geom_raster()`, `stat_summary_2d()`, and `geom_bin_2d()` — which #155 has just made readable, so a binned scatter now announces a correct grid of counts against an axis called "x".

Not the base R adapter. `BaseRHeatmapLayerProcessor$extract_axis_titles()` reads `xlab`/`ylab` off the recorded call and carries a reasoned default for `heatmap()` versus `image()`, so it is right. The two adapters disagree, and the ggplot2 one is the wrong half.

Not the other ggplot2 processors either. `build_axes(x = "x", ...)` appears exactly once in `R/`; every other processor resolves its labels. `Ggplot2PointLayerProcessor$extract_axes_labels()` is the pattern — `built$plot$labels$x`, then `plot$labels$x`, then `rlang::as_label(plot$mapping$x)`.

## Fix

Give the heatmap processor the same resolution the point processor already has. Worth extracting the lookup rather than copying it a third time, since it is the same three-step fallback either way.

A test should assert on a chart with a `labs()` override **and** on one without, since the fallback to the mapped column name is the case a `labs()`-only test would miss — and the two failure modes read identically from the announcement.

PR #157, opened 2026-08-13
feat(ggplot2): read geom_hex as a hexagonal bin lattice

### LINKED ISSUE #136: feat: support geom_hex and stat_bin_2d

jooyoungseo:
# Pull Request

## Description

`geom_hex()` matched no branch in `detect_layer_type()`, so the layer fell through to the unknown processor and the chart rendered as a static image with nothing to read at all.

It is a lattice of counted cells, which is nearly a heatmap — and the one difference is why it gets a type of its own rather than joining `heat`. **Hexagons tessellate because alternate rows are offset by half a cell**, so a bin's column index is not its position: bin 3 of one row and bin 3 of the next sit at different x. Read as a heatmap the chart would navigate perfectly well and put every bin past the first row on the wrong coordinate. The core's `HexbinTrace` announces centres instead, and this emits what it expects.

A four-bin hexbin of 200 points, after:

```
type: hexbin
axes: {"x":{"label":"First"},"y":{"label":"Second"},"z":{"label":"count"}}
rows: 6   row sizes: 2,4,4,5,4,1
bins: 20  selectors: 20
first bin: {"x":-0.577,"y":-2.9545,"count":3}
first selector: polygon#geom_hex\.polygon\.12\.1\.1
```

## Related Issues

Closes #136. The `bin_2d` half landed in #155; this is the other one.

Both of that issue's blockers are gone. xability/maidr#799 shipped the `hexbin` trace in v4.2.0, and the `hexbin` package — which `stat_binhex()` needs to compute at all — is added to `Suggests` here, with the tests skipping when it is absent.

## Rows are ragged, and stay that way

`stat_binhex()` emits only the bins that hold something, so the lattice is uneven by construction: `2, 4, 4, 5, 4, 1` above. Padding it to a rectangle would announce cells that were never drawn. The frontend's `MovableGrid` clamps a row change to the new row's length precisely because grids arrive like this.

Rows ascend in y, because the frontend's UPWARD steps to the *next* row index — the same convention the heatmap follows.

## The selectors follow the regrouping, not the document

gridSVG exports the layer's single multi-polygon grob as one `<polygon>` per hexagon, in built-data order, each with an id of the form `<grob>.1.<n>`:

```
polygons: 20
parent id: geom_hex.polygon.12.1
sibling tags under parent: polygon
```

So a bin is addressed by the built row it came from, and the emitted list is permuted with the data rather than trusting document order.

**`stat_binhex()` does emit its rows bottom-first today**, which means a list left in document order would pass — and would be undetectable the moment that stopped, since a hexbin announces centres and has no index to contradict. So the regrouping is tested *directly*, on a deliberately shuffled frame, separately from the chart that arrives already sorted. That split is real, not decorative: replacing the permutation with `seq_len(nrow(built_data))` fails only the shuffled test.

| reverted | fails |
| --- | --- |
| selectors left in built-data order | `the regrouping carries the built rows with it` |
| rows in descending y | `the lattice is read as rows, bottom first` |
| adapter branch removed | `a hexbin is its own layer type, not a heatmap` |

## Verification

| Gate | Result |
| --- | --- |
| whole suite, ggplot2 4.0.3 (what CI runs) | **0 failures** |
| whole suite, ggplot2 3.4.4 | **0 failures** |
| `test-hexbin-lattice.R` | 116 assertions |
| each mechanism reverted | a named test fails, per the table above |
| `roxygen2::roxygenise()` re-run | no drift |
| line lengths | under 100 |

The selector test resolves each emitted selector against the **real exported SVG** and checks the polygon it matches is the one whose centre and count that bin announces — read out of `ggplot_build()` rather than against literals, so it keeps meaning if the stat changes where it puts the bins.

N/A.

## Checklist

- [x] I have updated the documentation, if applicable.

## Additional Notes

**The branch carries a merge commit** folding in its own pre-squash history. #155 squash-merged, leaving those thirteen commits on the remote branch; force-push is not available to me here, so the merge is how the branch gets back in step. Its content diff against `main` is empty, and the diff this PR shows is only the `geom_hex` files.

**Found while writing the example, filed separately as #156**: every ggplot2 heatmap announces its axes as literally `"x"` and `"y"`. `Ggplot2HeatmapLayerProcessor` builds them from two string literals, so `labs(x = "Task")` and the mapped column name are both ignored — a cell reads "x: y, y: a, score: 3". It is the only `build_axes(x = "x", ...)` in `R/`; the base R adapter reads real labels. Not touched here, but it now also affects the binned scatter #155 just made readable.

---


jooyoungseo:
One thing a reviewer should know before reading the coordinate handling here, because the same question came up on the Python side and got a different answer.

py-maidr's `hexbin` binding **declines** a log-scaled lattice. `Axes.hexbin` takes its own `xscale`/`yscale`, bins in the transformed space, and hands back offsets in that space — a bin centred at x = 3.4 would announce `0.53`. That is a hexbin-specific argument, so declining that one call is coherent.

`geom_hex()` has no such argument, and this PR does nothing about transformed scales. That is deliberate, and the reason is that ggplot2's transformation is not the layer's:

```
scale_x_log10(), data spanning 0.671 to 9.903
  geom_hex   built x range: -0.146 to 1.023
  geom_point built x range: -0.174 to 0.996
```

The scale transforms before the stat, so **every** ggplot2 layer in this package reads transformed positions, not just this one. A `geom_point` scatter of prices from $5.50 to $9,403 on `scale_x_log10()` currently announces 0.744 to 3.973 under the label "Price (USD)".

Filed as #158. Singling out `geom_hex` for a package-wide problem would make this layer inconsistent with every other one and would leave the actual bug in place, so it is out of scope here — but it is worth knowing that a log-scaled hexbin merged by this PR is exactly as wrong as a log-scaled scatter already is, rather than newly wrong.

---

jooyoungseo:
Took all three. The first one was worth more than "follow-up issue" — it is the same silent-highlight class this PR exists to avoid, and the machinery to fix it was already sitting on the base class where the review pointed.

**1 — fixed.** `find_hex_polygon_name()` now searches inside `find_layer_grob_tree(plot, gt, panel_ctx)` rather than the panel, so a layer resolves to its own grob:

```
layer 1  bins: 14  selectors: 14  grob: geom_hex.polygon.12.1.*
layer 2  bins: 39  selectors: 39  grob: geom_hex.polygon.16.1.*
```

Reverted to the panel-wide first match, `two hexbin layers each address their own hexagons` fails.

It would have been the worst-behaved instance of the bug, not a mild one: both selector lists come out the right length and every count is real, so nothing contradicts it — and a hexbin announces centres, so there is no index to notice sitting on the wrong hexagon. You are right that the heatmap processor has the same shape (`rect_children[1]`); that is a separate defect in an existing layer rather than something to copy into a new one.

**2 — taken, and the case you flagged as "unlikely to matter" turned out not to matter at all, which I had claimed otherwise.** `extract_axes` now calls `resolve_legend_label(..., layer_index = self$get_layer_index())`.

I first wrote a test asserting the layer-mapping fallback, with a comment saying that was the reason for the reuse. Then I reverted `layer_index` to `NULL` and the test still passed — exactly as you said it would, because ggplot2 records every layer's mapping in `built$plot$labels` while building, so the override path answers first. The comment was wrong and is gone; the case is still asserted, but for the outcome rather than as evidence about which branch ran.

The reuse does change one thing, and that is now the assertion that earns it:

```
labs(x = "")  ->  [x]     (was: the empty string)
```

An axis announced as `""` is worse than one announced as its column name. `resolve_legend_label`'s `nzchar` guard handles it; my hand-rolled `length(label) == 1L` check did not.

**3 — left, as you suggested**, and I checked your reasoning rather than just agreeing: `split()` on a numeric vector factors it, and `factor()` orders levels by the sorted numeric values, so the names do come back ascending and the re-sort is a no-op today. It stays because the ordering is a requirement of the frontend rather than an incidental property of `split`, and the failure if it ever changed would be a chart flipped under the cursor with every announcement still true.

Whole suite green on ggplot2 4.0.3 and 3.4.4, `roxygenise()` re-run, no drift.

---

PR #160, opened 2026-08-14
fix(ggplot2): announce transformed scales in the space the reader sees

### LINKED ISSUE #158: A transformed scale is announced in transformed space: scale_x_log10() reads out log10, labelled as the data
ggplot2 applies a scale transformation **before** the stat, so `ggplot_build()`'s data holds the transformed values. Every ggplot2 processor reads that data straight through, so a chart on `scale_x_log10()` announces log10 coordinates under the original axis label.

```r
d <- data.frame(price = 10^runif(60, 0.7, 4), size = runif(60, 1, 9))
ggplot2::ggplot(d, ggplot2::aes(price, size)) +
  ggplot2::geom_point() +
  ggplot2::scale_x_log10() +
  ggplot2::labs(x = "Price (USD)", y = "Size")
```

```
announced x range : 0.744 to 3.973
actual data range : 5.5 to 9403.5
axis label        : Price (USD)
axes x min/max    : 0.583 to 4.135
```

A sighted reader sees ticks at 10, 100, 1000, 10000 and a point sitting near $6,000. A screen-reader user is told **"Price (USD): 3.973"**.

This is the failure mode that is worse than an unsupported chart. Nothing is missing, nothing errors, the structure is right, the point count is right, the label is right, and the numbers are false. There is no signal anywhere in the output that would let a reader catch it — the only clue is that the values look small, and "small" is not something you can check without the chart you cannot see.

Grid navigation inherits it too: `axes.x.min`/`max` are the transformed limits, so the coordinate readout agrees with the wrong numbers rather than contradicting them.

## Scope

Package-wide on the ggplot2 side, not one processor. The transformation is applied by the scale, so it reaches every layer that reads built positions — verified identical for `geom_point` and `geom_hex`:

```
geom_hex   built x range: -0.146 to 1.023   (data 0.671 to 9.903)
geom_point built x range: -0.174 to 0.996
```

Any `trans`/`transform` scale does it, not only `log10`: `scale_*_sqrt()`, `scale_*_reverse()`, `scale_*_continuous(transform = "log2")`, and `coord_trans()`. `scale_*_reverse()` is the nastiest of them, since it negates — a value of 40 is announced as -40, which reads as a real number in the wrong direction rather than as an obviously odd one.

Not the base R adapter, which records the call's own coordinates.

## What a fix needs

ggplot2 keeps the transformation on the scale, and it is invertible:

```r
scale <- built$layout$panel_scales_x[[1]]
scale$get_transformation()$inverse(values)   # ggplot2 >= 3.5
scale$trans$inverse(values)                  # earlier
```

so the values can be put back into data space where the reader expects them. Both spellings are needed while 3.4.4 is supported — that split has already bitten this repo twice (#151, #155).

Three things to decide, and they are the real work rather than the call itself:

1. **Where it goes.** One place that every processor passes positions through, or each processor separately. The former, presumably, but there is no such seam today.
2. **Axis limits as well as points.** `axes.x.min`/`max` come from the panel ranges and are transformed too. Inverting the points and not the limits would leave grid navigation disagreeing with the announcement, which is worse than the current state where at least they agree with each other.
3. **`coord_trans()` is a different mechanism** from a scale transform — it transforms at draw time, after the stat, so the built data is *already* in data space and inverting it would break a chart that currently reads correctly. Whatever is done has to tell the two apart rather than test for "is there a log axis".

A test should assert on the announced value against the source column, not against a literal, and cover at least `log10`, `sqrt` and `reverse` — `reverse` because a sign flip is the one an eyeballed range check would pass.

jooyoungseo:
Closes #158.

ggplot2 applies a scale transformation **before** the stat runs, so `ggplot_build()`'s data is in the transformed space. Read straight through, a `scale_x_log10()` chart announces log10 coordinates under the original axis label:

```r
ggplot(prices, aes(price, size)) + geom_point() +
  labs(x = "Price (USD)") + scale_x_log10()
```

| | announced before | actual |
|---|---|---|
| first x | 0.744 | $5.55 |
| last x | 3.973 | $9,403 |
| label | "Price (USD)" | — |

This is the failure mode that is worse than an unsupported chart. Nothing is missing, nothing errors, the structure is right, the point count is right, the label is right, and the numbers are false — with no signal a reader could catch, because "these look small" is not checkable without the chart you cannot see.

## What changed

`R/scale_transform_utils.R` (new) reads the panel scale's transformation — `get_transformation()` on ggplot2 ≥ 3.5, `$trans` before it — and inverts positions through it. `identity` returns `NULL` and costs nothing.

The point and smooth processors call it. Those are the two that carry continuous positions through this seam; the rest are named below.

### Three decisions worth the review

**Inverted at emission, not on the frame.** `scale_*_reverse()` *negates* rather than compresses, so a frame inverted before a sort would order rows opposite to the way they were drawn — and the selectors indexed by that order would then land on the wrong element. Ordering follows the drawn scale; only the announced number moves. (The smooth processor inverts the frame, because nothing below it orders by these values: the split is on `group` and the rows keep the order `StatSmooth` emitted. The comment says so at the call site.)

**A transformed axis emits no navigation grid.** Grid navigation walks equal increments, and the breaks a reader hears — 10, 100, 1000 — are equally spaced only in the space the points are no longer announced in. Emitting the transformed range instead would leave the grid and the announcement in *different* spaces, which is worse than before, where the two were at least wrong together. The label survives: an axis with no name is a worse outcome than an axis with no grid. This is per axis, so a chart with one transformed axis does not lose both.

**`coord_trans()` needs no special case and deliberately gets none.** It transforms at draw time, *after* the stat, so its built data is already in data space **and** its scale reports `identity` — the same comparison that skips an untransformed chart skips it too. A test for "is there a log axis" would have inverted it a second time and broken a chart that reads correctly today. `tests/testthat/test-transformed-scales.R` asserts this both through a render and by driving `panel_transformation()` directly, since the render alone cannot show that `identity` and `coord_trans()` take the *same* branch.

## Why `reverse` is in every test case

It negates rather than compresses, so a plausibility check on the announced range passes it while every value has the wrong sign. `log10` and `sqrt` would both be caught by "do these numbers look too small"; `reverse` would not.

## Falsification

Each mechanism reverted in turn, against the new test file:

| reverted | failing assertions |
|---|---|
| baseline (all present) | 0 |
| point emission un-transform | 3 |
| grid suppression | 9 |
| smooth un-transform | 1 |
| restored | 0 |

## Verification

Full suite, 0 failures on **ggplot2 3.4.4** and **ggplot2 4.0.3** (CI runs 4.x; 3.4.4 is the older floor). Round-trip error through `$inverse()` measured at 0 for `log10` and `reverse`, 1.8e-12 for `sqrt`, on both versions.

## Scope left for a later pass

The line processor and the remaining continuous-position processors are untouched. Only 3 of 34 ggplot2 processors read through the base-class `get_layer_built_data` seam, so a wholesale sweep is not one edit, and doing it fast is how the ordering hazard above gets reintroduced somewhere it is not noticed. Discrete positions are indices into a scale's levels rather than measurements and are returned untouched, which the tests pin.

---


jooyoungseo:
All three findings verified before acting. The first one was more than informational — thank you for it.

## 1. The line processor — half of it was already broken, and you found which half

You're right that `recover_x_values()` handles x, and right that my "line processor untouched" framing misled. But the y half is not merely untested — it was **wrong**. Measured on a `geom_line()` series spanning $5.01 to $10,000:

| | announced before | data |
|---|---|---|
| `scale_x_log10()` → x | 5.012 .. 10000.000 | ✅ already right |
| `scale_y_log10()` → y | **0.700 .. 4.000** | 5.012 .. 10000.000 |

So a log-y line chart was reading in log space this whole time, and x being right is exactly what hid it: a reader checking one axis against the other would find the chart internally consistent. Fixed in `f12f395` — `layer_data$y` goes through the shared `untransform_positions()` in `extract_data()`. After: 5.012 .. 10000.000 for `log10`, `sqrt` and `reverse` alike.

Applied to the frame rather than at each emission, and that is safe here for a reason worth stating, since it is the opposite of the choice made for points: nothing below orders by y. The split is on `group`, and the only other read of `y` is an NA filter — so inverting cannot reorder rows against the drawn polyline the way it could on x.

The discrete-y case is the one that could have broken quietly, since a factor y reaches the payload as a level code that `attach_discrete_y_names()` turns into the name a reader hears — inverting *that* would be undoing an index. It survives because a discrete scale reports no transformation, verified directly:

```
discrete y       panel_transformation: NULL
continuous y     panel_transformation: NULL
log10 y          panel_transformation: log-10
reverse y        panel_transformation: reverse
coord_trans y    panel_transformation: NULL
```

and pinned by its own test rather than left to that reasoning.

On folding `get_x_transformation()` onto the shared helpers: agreed in principle, and I checked they read the same object — `built$layout$panel_params[[i]]$x$scale` is `identical()` to `built$layout$panel_scales_x[[i]]`, for x and y both — so the consolidation is safe. Not doing it in this PR: the line processor's x path goes *forwards* (`transform_x_values()`, raw → transformed) to keep the break-matching comparison exact, which is a different operation from inverting, and collapsing the two deserves its own diff rather than riding along under a y fix.

## 2. `NEWS.md`

Added under `## Bug Fixes`, in the existing style.

## 3. Faceted / free-scale panel

Added — but asserted against the helper rather than through a rendering, and deliberately. ggplot2 gives free scales different *limits*, never different *transformations*, so every panel of a rendered facet would read correctly even with the panel index wrong; a render test would pass without exercising the branch at all. What the test pins is what the branch actually does:

```r
expect_length(build$layout$panel_scales_x, 2)     # free_x: one per panel
for (panel in 1:2) expect_false(is.null(panel_transformation(build, "x", panel)))
expect_false(is.null(panel_transformation(build, "x", 99)))   # past the end
expect_length(build$layout$panel_scales_y, 1)     # y shared
expect_null(panel_transformation(build, "y", 2))
```

The `99` case is the fixed-scale shape, where one scale serves every panel — not an error, and not a reason to give up and announce transformed values.

## Verification

Reverting the line change fails **3** assertions. Full suite: **0 failures** on ggplot2 3.4.4 and 4.0.3.

---

PR #161, opened 2026-08-14
feat(base-r): read vioplot violins as violin_box + violin_kde

### LINKED ISSUE #132: feat: support violin plots in the base R adapter
## Description

The ggplot2 adapter has `R/ggplot2_violin_layer_processor.R`, and `Ggplot2ProcessorFactory$get_supported_types()` lists `"violin"`. The base R adapter has no equivalent — there is no `base_r_violin_layer_processor.R`, and violin plots drawn through base R graphics fall through to `BaseRUnknownLayerProcessor`.

## Why it matters

Which plotting system a user chose should not decide whether their chart is accessible. Every other type in the ggplot2 factory has a base R counterpart, so violin is the one asymmetry in an otherwise complete pair.

## Scope

- [ ] `R/base_r_violin_layer_processor.R`, following the structure of the existing base R processors
- [ ] Register `"violin"` in the base R processor factory and in `get_supported_types()`
- [ ] Emit the `violin_kde` + `violin_box` layer pair, matching what the ggplot2 processor produces
- [ ] Tests alongside the existing base R processor tests

## Difficulty

Medium — the ggplot2 processor is a close reference, but base R exposes the density differently from `ggplot2`'s computed `violinwidth`.

jooyoungseo:
Closes #132.

## The gap

`Ggplot2ProcessorFactory` has a violin processor; the base R adapter had none, so `vioplot::vioplot()` fell through to the unknown-layer path and the chart landed on maidr's static fallback. Which plotting system a user chose should not decide whether their chart is accessible.

## Why this was listed as blocked, and why it is not

I put this issue on hold myself, on the grounds that `vioplot` computes its curve through `sm::sm.density` with internally-built arguments that could not be recovered. **That was wrong, and I had not read the function.** Its call is fully visible:

```r
est.xlim <- c(min(lower, data.min), max(upper, data.max))
smout <- do.call("sm.density", c(list(data, xlim = est.xlim), args))
base[[i]]   <- smout$eval.points
height[[i]] <- smout$estimate * hscale
```

with `args <- list(display = "none")`, plus `h = h` when the caller passed a bandwidth. `sm.density` is deterministic, so replaying it returns the curve vioplot drew — the same approach `BaseRBoxplotLayerProcessor` already takes when it re-calls `boxplot(plot = FALSE)`.

That distinction is the whole point. A kernel density estimate computed with different defaults is not wrong in any way a reader could detect; it simply describes a shape the chart does not draw.

## Verified against the drawing, twice, from directions that do not touch

**The box statistics, against vioplot's own return value.** `vioplot()` returns `upper`, `lower`, `median`, `q1` and `q3` invisibly:

```
violin 1: q1 TRUE  q3 TRUE  med TRUE  lower TRUE  upper TRUE
violin 2: q1 TRUE  q3 TRUE  med TRUE  lower TRUE  upper TRUE
```

`identical()`, not `all.equal()` — bit-for-bit.

**The curve, against the drawn polygon.** The violin body grob carries `n=200` vertices for a sample whose `sm.density` returns 100 evaluation points: that curve mirrored about the category position. This ties the replayed call to the rendered geometry without passing through any of maidr's own code, which is why it is the assertion worth keeping.

## Decisions

**`est.xlim` simplifies to the data range.** `lower` is `max(..., data.min)` so never below it, and `upper` is `min(..., data.max)` so never above — both outer calls collapse. The simplified form is used with the derivation recorded rather than the arithmetic repeated.

**`h` and `range` come from the call, not from defaults.** A caller's bandwidth goes straight to `sm.density`, and a smoother curve can show one mode where the chart shows two. `range` decides where the whiskers stop.

**A category with no spread is left out.** vioplot draws a degenerate mark for it; announcing a curve would claim a spread the chart does not show.

**A formula call is declined.** Resolving it needs the environment the call was made in. Declining leaves the chart on the static fallback, which says nothing rather than something invented.

## Selectors, better placed than the plotly binding's

vioplot draws the parts of the box as *separate* grobs, so unlike a chart whose box is one path, each section points at what it actually is:

| section | grob | SVG element |
|---|---|---|
| `min`, `max` | `lines-{i}` | `polyline` |
| `iq` | `rect-{i}` | `rect` |
| `q2` | `points-{i}` | `use` |
| body | `polygon-{i}` | `polygon` |

Element names verified against a real `gridSVG::grid.export()` rather than guessed. Every emitted selector was then resolved against that export in Chromium across a twelve-violin plot: **48 of 48 matched exactly one element**, including the double-digit case where a lexical sort would put `-10` between `-1` and `-2`.

## Falsification

| reverted to | failures |
|---|---|
| a different quantile rule | 4 |
| no violin branch in the factory | 3 |
| a flat sample announced anyway | 3 |
| `h` ignored | 2 |
| `est.xlim` from the whiskers | 2 |
| plot index dropped from the grob pattern | 2 |
| lexical sort of grob ids | 2 |
| `range` ignored | 1 |
| all box sections sharing one grob | 1 |

**Two initially survived, and both were my error rather than a missing test.**

*`est.xlim` from the whiskers* passed everything. My curve test asserted only the point **count**, and `sm.density` returns 100 evaluation points whatever span it is given — so the count could not move. Measured, the two spans genuinely differ (`[6.845, 13.091]` vs `[6.845, 12.737]`), and the endpoints are now asserted directly.

*The anchored grob pattern* passed everything, and the reason was that **my justification for it was false**. I claimed anchoring stopped `graphics-plot-1-polygon` matching `graphics-plot-11-polygon-1`, and that it excluded the panel frame. Measured, neither holds: the `-` delimiter excludes the eleventh plot with or without anchoring, and the frame (`graphics-plot-1-box-1`) never enters a search for `polygon` because it is not named that, despite being a polygon grob. What `$` genuinely excludes is a longer name beginning the same way. The docstring now says that, the test that pinned nothing was replaced with one pinning the separation that matters, and the revert fails.

## Checks

Full suite **5377 passing, 0 failures, 0 errors** (99 skips, all pre-existing — patchwork and other optional packages). End-to-end verified through `save_html()`: a real `vioplot()` call renders HTML carrying both layers with selectors naming grobs present in the same document.

`vioplot` and `sm` added to Suggests; `vioplot` wrapped late through a `packageEvent` hook beside the existing quantmod one, with matching `.onUnload` cleanup, and `sm` guarded by `requireNamespace`.

---


jooyoungseo:
All three addressed. The first minor point was a real bug and worse than described.

## Docs drift — already fixed

`7b44ed3`, pushed a few minutes after the review started, so it was reading `192cf03`. `roxygen2::roxygenise()` regenerated `man/base-r-wrappers.Rd` with the `\alias{vioplot}` and usage entry, plus man pages for the new helpers; output is idempotent, which is what the drift check compares.

## The named first argument — a real silent loss

Confirmed, and the severe case is worse than the one you named:

```
vioplot(x = a, b)   ->  1 group   ** the chart draws two
vioplot(x = a)      ->  0 groups  ** the chart announces nothing at all
```

The second is a single-violin chart written the perfectly ordinary way, producing no layers. And the first is the shape this whole PR is about: the announced violin was `b` carrying the label the other one should have had.

You framed this as a narrow edge case worth a comment. It is narrower than the positional form, but "announces nothing" is not a degradation I would want to leave documented rather than fixed — so `x` is now treated as data, in call order, while every other named argument (`names`, `horizontal`, `h`, `range`, `col`, …) stays out.

**Fixing it exposed a second error I had made.** Once `x = a` is read, its *argument* name becomes a candidate label — and vioplot does not draw it. Measured against the axis labels vioplot actually renders:

```
vioplot(a, b)                 ->  "1", "2"
vioplot(x = a, b)             ->  "1", "2"
vioplot(list(p = , q = ))     ->  "p", "q"
vioplot(data.frame(p =, q =)) ->  "p", "q"
```

So a name is a label only when it came from inside a list or a data frame. My original code took `names(numeric_only)` in every branch, which would have announced a violin as "x" that the chart labels "1". Restructured to make the distinction explicit, with the measurements in the comment.

Falsified: reading only unnamed arguments fails 4 tests; treating argument names as labels fails 2.

## `has_sm_package()` — right conclusion, slightly different route

You're right that the branch is unreachable, though `sm` is in vioplot's **`Depends`**, not its `Imports`:

```
Depends : sm, zoo
Imports : (empty)
```

Same practical consequence — `sm` is attached wherever a violin can have been drawn — so the guard cannot fire in a normal installation. Kept as defensive code so a broken one surfaces as a missing layer rather than an error thrown from inside a render, but the comment no longer claims a scenario that cannot occur. That overstatement was the fair part of the criticism.

## Checks

Full suite **5383 passing, 0 failures, 0 errors** (up 6 from the new tests). `roxygen2::roxygenise()` re-run and idempotent.

Noted on the last point: you couldn't run R in the review environment, so the CI results are the check that matters there — worth saying that the static reading still found the one defect I had shipped.

---

jooyoungseo:
CI came back red on `ubuntu-latest (release)` and `ubuntu-latest (oldrel-1)`. It was mine, and I had mis-triaged it. Reporting the whole chain because the way I missed it matters more than the fix.

## What failed

Both Linux jobs ran the suite clean — `[ FAIL 0 | WARN 24 | SKIP 5 | PASS 6340 ]` — and then failed the check itself. I pulled the check artifact rather than guessing:

```
* checking for unstated dependencies in ‘tests’ ... WARNING
'::' or ':::' import not declared from: ‘gridGraphics’
Status: 1 WARNING
```

Both new test files call `gridGraphics::grid.echo()` to read the grobs vioplot drew. `gridGraphics` reaches this package only as a transitive dependency of ggplotify — installed in practice, declared nowhere.

## How I missed it

I ran `R CMD check` locally before asking for the merge and reported `Status: 3 WARNINGs, 2 NOTEs`, calling them all environmental. Two of the three were — a missing `en_US.UTF-8` locale, and empty `inst/doc` because I had passed `--no-build-vignettes`. But I reached that verdict for the group rather than for each one, and this warning was not in it at all: my local invocation passed `--no-tests`, so `checking for unstated dependencies in 'tests'` never ran. I reported a clean check that had not looked at the thing that broke.

## The fix

`gridGraphics` declared in `Suggests`. The two call sites then differ on purpose:

- `violin_grobs()` in `test-base-r-violin.R` skips without it. All four of its callers go on to assert on selectors resolved against those grobs, so there is nothing left to check.
- `drawn_vioplot()` in `test-vioplot-stats.R` does **not** skip. Its grob capture already falls to `NULL` on error, and only the one test reading the drawn polygon needs it. The `identical()` comparison against `vioplot()`'s own return value needs no grobs at all and keeps running.

## Verified

Re-ran the check with the dependency scan actually enabled this time:

```
* checking for unstated dependencies in ‘tests’ ... OK
```

The three warnings that remain locally are the ones I can now name individually: the missing locale, and two from `inst/doc` being empty under `--no-build-vignettes`. CI builds vignettes and reported neither.

Both violin files pass locally (`47` assertions, 0 failures, 0 skips — `gridGraphics` is present here, so the guards are inert rather than hiding anything). Full suite re-running.

Pushed as `758946a`. I'll confirm all four platforms plus docs-drift and test-coverage before merging — properly this time.

---

PR #166, opened 2026-08-15
feat(ggplot2): keep the confidence band geom_smooth draws

jooyoungseo:
Refs #135 — the sub-item that issue calls out as *"affects a layer that r-maidr does support"*. The other three boxes on #135 are already done on `main` (`ggplot2_errorbar_layer_processor.R` exists, `error_bar` is in `get_supported_types()`, and the adapter detects `GeomErrorbar`/`GeomErrorbarh`/`GeomLinerange`/`GeomPointrange`/`GeomCrossbar`); I verified each before starting rather than assuming.

## The defect

`se = TRUE` is `geom_smooth()`'s **default**, and the band is the reason the layer is drawn rather than a plain line: it says how much of the fitted trend the data supports. `StatSmooth` computes it alongside the fit, and the processor read only `y`:

```
built cols: x, y, ymin, ymax, se, flipped_aes, PANEL, group, …
row 1        y = 0.956   ymin = -1.567   ymax = 3.480
emitted      {x: 1, y: 0.9565}
```

So a regression chart that otherwise worked was silently missing the half a reader needs to judge it.

## The fix

The band becomes its own `error_bar` layer:

```
geom_point() + geom_smooth(se = TRUE)  →  point, smooth, error_bar
                                          error_bar pt1: {x: 1, y: 0.9565, yMin: -1.567, yMax: 3.4799}
```

Emitted as a separate layer rather than as bounds on the smooth points because that is the shape MAIDR reads today — `ErrorBarTrace` consumes `y`/`yMin`/`yMax`, while `SmoothTrace` extends `LineTrace` and reads neither. It is also what the Python binding already emits for `sns.pointplot`, with the identical keys, so the two bindings describe an interval the same way.

## Two guards, both load-bearing

**The band is keyed on the layer's *stat*, never on the columns — and the existing suite is what caught this.** My first attempt keyed on `ymin`/`ymax` being present, and three `geom_density()` tests went red. The columns lie: this processor serves `StatDensity` too, and that fills the same two columns meaning something else entirely. Measured on a three-group density:

```
StatDensity  →  ymin = 0, ymax = density     (all 1536 rows finite)
```

That is the extent of the fill, not an uncertainty. Reading it as an interval would have announced **every density curve as having a confidence band running from zero** — a false claim about the data rather than a missing feature, and the worse of the two failures.

**A facet panel is skipped, and this one is a regression I introduced and caught.** `combine_facet_layer_data()` collapses a panel's layers into one payload entry and types it from the first result that produced one, so a panel holds a single layer type by construction. Handing it a multi-layer result does not add the band — it reads `result$data`, finds nothing, and drops the fitted curve too. Measured against baseline before the guard existed:

```
faceted point + smooth   main: 11 entries/panel   →   mine: 10   (smooth lost)
```

The patchwork path *does* expand `multi_layer`, so the guard keys on `panel_ctx$facet_groups`, which only the facet path supplies, rather than on `panel_ctx` being present at all.

**The consequence is a real limitation and I want it visible rather than buried:** a *faceted* regression chart still loses its band. Expressing one there means teaching the facet path to hold two layer types per panel — a change to shared composition code, not to this processor. It is pinned by a test so widening it later is deliberate.

## Selectors

The band emits none. A ribbon is one polygon covering every sample, so there is no per-sample element to address; repeating one selector for each point would highlight the whole band everywhere and look like it worked.

## Verification

Full suite: **5428 passed, 0 failed, 0 errors** (99 skips are pre-existing `patchwork cannot be loaded`). `man/` regenerated with roxygen2 so the drift check stays green. 13 new tests.

Behaviour across the cases that matter:

| plot | layers |
|---|---|
| `geom_point() + geom_smooth(se = TRUE)` | point, smooth, **error_bar** |
| `geom_smooth(se = FALSE)` | point, smooth |
| `geom_smooth(method = "loess", se = TRUE)` | smooth, **error_bar** |
| `geom_density(aes(fill = g))` | smooth — **no band** |
| `geom_area(stat = "density")` | smooth — **no band** |
| faceted `geom_point() + geom_smooth()` | point (11 entries, unchanged) |
| `geom_smooth(aes(colour = g), se = TRUE)` | smooth, **error_bar** |

Falsification — each mechanism reverted in turn, `smooth`-filtered suite:

| reverted | result |
|---|---|
| band never emitted | **3 failed, 3 errors** |
| stat guard removed | **5 failed** — the density cases |
| facet guard removed | **1 failed** — the facet regression pin |
| *(restored)* | **632 passed, 0 failed** |

## Follow-on

Carrying the band *on* the smooth layer would be the better answer — a reader would hear the value and its interval at the same point, instead of switching layers — and it would sidestep the facet limitation entirely, since it stays one layer. That needs `LineTrace`/`SmoothTrace` to read `yMin`/`yMax` in the core, and py-maidr would want it too: `sns.regplot` has exactly the same gap today. Worth filing upstream; this change is correct either way and would simplify into it.

---


jooyoungseo:
All three findings actioned in `84a1c10`. The first one was a real defect and my PR body claimed the case worked — thank you for checking it rather than taking the table at face value.

**1. Grouped band flattened.** Reproduced before changing anything, and it is exactly what you described:

```
geom_smooth(aes(colour = g), se = TRUE)
  smooth      2 curves
  error_bar   160 points, ONE layer
              x monotonic? FALSE
              x at 79,80,81:  9.886  10  1     ← seam, nothing announced
              has z?  FALSE
```

The root cause is worth stating precisely, because it rules out the obvious fix: an `error_bar` layer **cannot** carry groups. `ErrorBarPoint` in the core is `{x, y, yMin?, yMax?}` with no `z`, and `ErrorBarTrace` holds `points: ErrorBarPoint[]` — a flat sequence. So there is nowhere inside one layer for a group boundary to live.

Each curve therefore becomes its own band layer, told apart by `name`. That field turns out to exist for precisely this, and its JSDoc names the case:

> Without it, two layers of one type are indistinguishable — a hue-split error bar chart…

```
grouped   →  smooth, error_bar(name="a", 80pts, monotonic), error_bar(name="b", 80pts, monotonic)
single    →  point, smooth, error_bar(unnamed, 80pts, monotonic)
```

A lone band stays unnamed — the type already identifies it, so naming would add a word the reader gains nothing from.

**2. Faceted leaf inside a patchwork.** Added the third clause. You were right that it is not a regression this PR introduces, but it is cheap and it makes `can_emit_band()` agree with `Ggplot2ViolinLayerProcessor$process()`, which carries the identical check for the identical composition — and disagreeing with a sibling processor about the same question is how these drift.

**3. `NEWS.md`.** Added.

Verification after all three: **5433 passed, 0 failed, 0 errors**; `man/` regenerated. Falsification of the new split — band not split by group: **2 of 18 fail**, and they are the seam and naming cases.

---

PR #167, opened 2026-08-15
feat(ggplot2): read a bare geom_ribbon as the band or area it draws

jooyoungseo:
Refs #135 — the last of its uncertainty geoms. `geom_errorbar`, `geom_errorbarh`, `geom_linerange`, `geom_pointrange` and `geom_crossbar` all resolve today; `geom_ribbon` was the one still falling through:

```
geom_ribbon      -> unknown      ← dropped
geom_pointrange  -> error_bar
geom_crossbar    -> error_bar
geom_linerange   -> error_bar
geom_errorbarh   -> error_bar
```

It matters because it is the shape a user gets assembling `geom_smooth()`'s two halves by hand — a `geom_ribbon` band under a `geom_line` fit — and the interval was simply lost.

## Why it is not one more name in the membership test

A ribbon is **not automatically an interval**. `geom_ribbon(aes(ymin = 0, ymax = y))` is an area chart, and both misreadings are real:

- an area read as an interval reports a filled magnitude as a bound, telling a reader the chart is less certain than it is about data carrying no uncertainty at all;
- a band read as an area announces `ymax` as a magnitude and drops `ymin` entirely.

The baseline separates them, and it is the **same rule the Python binding already draws for `fill_between()`**: filling from zero to one curve is an area, anything else is the gap between two curves. Measured on a pair of ribbons:

```
aes(ymin = lo, ymax = hi)  →  ymin = 0, 1, 2 …     all zero? FALSE  → error_bar
aes(ymin = 0,  ymax = y)   →  ymin = 0, 0, 0 …     all zero? TRUE   → area
```

Read off the **built frame** rather than the mapping, because `ymin` may be a constant, a column, or a computed aesthetic and only the built data has resolved which. A layer that cannot be built is treated as a band — the reading that loses nothing, since an area announced as an interval still carries both edges.

Only an identically-*zero* lower edge counts. A ribbon measured from some other constant still draws a gap, and its heights are measured from somewhere the announcement would not mention; pinned by a test.

## `geom_area` is untouched, deliberately

`GeomArea` **inherits** `GeomRibbon`, so this is a `class(...)[1]` check rather than an `inherits()` one. An `inherits()` test would look correct and silently swallow every area chart into the ribbon branch — pinned, because that is exactly the kind of thing that reads fine in review.

```
ribbon band  →  error_bar (10 pts) + line     ← the line keeps its own reading
ribbon area  →  area
geom_area    →  area
stacked area →  stacked_area
geom_smooth  →  smooth                        ← its band is the smooth processor's
```

## A correction to my own test

I first asserted a stacked area detects as `stacked_area` and it came back `area`. The code was right and my expectation was wrong: whether bands stack is not knowable from the geom and position alone — a single-series area is drawn with `PositionStack` too — so detection answers `area` and the orchestrator resolves the stacking afterwards. The test now asserts the *emitted* type, with the reason written down.

I also had a bug of my own here worth noting: I first wrote `plot` where the parameter is `plot_object`, so it silently passed base R's `plot` **function** into `ggplot_build()`, which failed inside a `tryCatch` and made every ribbon an interval. Caught by the area case coming back `error_bar`.

## Verification

Full suite **5443 passed, 0 failed, 0 errors**; `man/` regenerated so the drift check stays green; `NEWS.md` updated. 10 new tests.

Falsification:

| reverted | failures |
|---|---|
| ribbon branch removed entirely | **5 of 10** |
| baseline test dropped (every ribbon an interval) | **2** — the two area cases |

---


jooyoungseo:
Your static analysis was right, and I could run R here to confirm it:

```
ribbon band   error_bar  data=10  selectors=0
line          line       data=1   selectors=1

find_interval_grob -> NULL
```

One correction to the mechanism: it is not the `shape$samples == expected` mismatch that produces the empty list — `find_interval_grob()` returns **NULL** outright, because it searches for errorbar-shaped grobs and a band is not one. So the outcome is even more incidental than diagnosed: it does not reach the shape check at all.

Taken option 2, since your point was precisely that the right outcome by accident is not the same as by design. `generate_selectors()` now declines up front for a ribbon with the reasoning written down, mirroring `Ggplot2SmoothLayerProcessor`.

**The part worth reporting: option 1 alone would not have worked.** I wrote the end-to-end test first, then falsified it by removing the new guard — and it still passed, 17 of 17. Of course it did: the lookup keeps returning nothing either way. A test that cannot fail is not a pin, so the predicate is asserted directly instead:

```
predicate disabled -> 22 passed, 1 failed
```

TRUE for a ribbon, FALSE for `geom_errorbar` / `geom_pointrange` / `geom_crossbar` / `geom_linerange`, and FALSE rather than an error for a layer index the plot does not have — a composition can hand a processor one.

Verified the decline is specific to the geometry rather than to the type: `errorbar`, `pointrange` and `crossbar` all still emit their one selector.

Full suite **5456 passed, 0 failed, 0 errors**; `man/` regenerated.

On the two minor notes — both accurate, both pre-existing, and I left them alone deliberately. `find_layer_index()`'s `identical()` matching and the extra `ggplot_build()` are shared with `draws_single_ring()`, so changing either here would fix half a pattern and leave the other half to drift.

---

PR #169, opened 2026-08-16
feat(ggplot2): read stat_ecdf as the step layer it draws

### LINKED ISSUE #168: feat: read stat_ecdf as a step layer
## Description

`stat_ecdf()` currently detects as `unknown`, so an empirical CDF — one of the more common statistical graphics — announces nothing at all and falls back to the static image.

```
stat_ecdf            types=unknown sizes=0
stat_ecdf by group   types=unknown sizes=0
```

This is a **parity gap**, not a new chart type: `step` is already in `get_supported_types()`, the JS core has a step trace, and py-maidr already reads `sns.ecdfplot` as `STEP`.

## Why it was declined, and why that was right

`R/ggplot2_adapter.R` gates the step branch on the stat:

```r
if (geom_class == "GeomStep" && stat_class == "StatIdentity") {
  return("step")
}
```

with a comment naming two concrete obstacles. Both reproduce exactly, on `n = 20` in 2 groups:

```
rows: 22  (n = 20)                    ← 2 padding rows
non-finite x: 2   values: -Inf, Inf
is x sorted as returned? FALSE        ← input order; GeomStep sorts in draw_panel
grouped: 24 rows, 2 non-finite per group, unsorted within group
```

So declining was correct: emitting those rows as they stand would announce an unsorted staircase whose first and last x are infinite.

Worth recording that the two bindings met the same fact and diverged. `jsonlite` serialises `-Inf` as the **string** `"-Inf"`, so r-maidr would have announced text where a number belongs; py-maidr's `json.dumps` writes a bare `-Infinity`, which `JSON.parse` rejects outright and which kills the chart (xability/py-maidr#427). r-maidr saw this and stepped back; the Python side did not.

## What supporting it needs

Both obstacles are tractable, and neither needs core work:

- **Sort by x within each group.** This is exactly what `GeomStep` does itself, in `draw_panel` — so sorting recovers the order that is actually drawn rather than imposing a new one.
- **Drop the `±Inf` padding.** Those two rows are the conceptual ends of the staircase, not observations. There is no x to announce for them and the reader loses nothing.

## Acceptance

- `stat_ecdf()` detects as `step` and emits one point per observation, in drawn order
- No non-finite value reaches the payload
- A grouped `stat_ecdf(aes(colour = g))` emits one series per group
- `geom_step()` with an identity stat is unchanged
- A stat that is neither identity nor ECDF still declines, rather than the gate being removed wholesale

jooyoungseo:
Closes #168.

## The gap

An empirical CDF detected as `unknown` and fell back to the static image — it announced **nothing at all**, even though `step` was already in `get_supported_types()`, the JS core already had a step trace, and py-maidr already read `sns.ecdfplot` as one. A parity gap rather than a new chart type.

## It was declined on purpose, and the reasons hold

`R/ggplot2_adapter.R` gated the step branch on `stat_class == "StatIdentity"`, with a comment naming two obstacles. Both reproduce exactly, on `n = 20` in 2 groups:

```
rows: 22  (n = 20)                    ← 2 padding rows
non-finite x: 2   values: -Inf, Inf
is x sorted as returned? FALSE        ← input order
grouped: 24 rows, 2 non-finite per group, unsorted within group
```

So the decline was correct as it stood. What this PR does is remove the obstacles rather than the gate.

## Neither obstacle needs a judgement call

**The sort is not an imposed order.** `GeomStep` does not draw the rows it is handed — `draw_panel()` calls `ggplot2:::stairstep()`, whose first line is:

```r
data <- as.data.frame(data)[order(data$x), ]
```

So the sorted rows *are* what is on screen. Sorting recovers the drawing; it does not invent one. An already-sorted layer — every `geom_step()` written by hand — is unchanged.

Ordered within `PANEL` and `group`, never globally, because `draw_panel()` runs once per panel per group. A grouped ECDF is several staircases, and a global sort would interleave them into one series walking backwards at every seam.

**The ±Inf rows are not observations.** They are the conceptual ends of the staircase; there is no x a reader could be told. Dropping them loses nothing.

## A cross-binding note worth recording

The two bindings met this same fact and diverged. `jsonlite` writes `-Inf` as the **string** `"-Inf"`; py-maidr's `json.dumps` writes a bare `-Infinity` that `JSON.parse` rejects outright, so the chart never initialises. r-maidr saw the hazard and stepped back — the Python side did not, which is xability/py-maidr#427, found while measuring this.

## Measured, after

```
stat_ecdf           type=step series=1 sizes=20     sorted=TRUE nonfinite=0
stat_ecdf grouped   type=step series=2 sizes=10,10  sorted=TRUE nonfinite=0
geom_step identity  type=step series=1 sizes=6      dir=hv
geom_step vh        type=step series=1 sizes=6      dir=vh
```

y comes out `0.05, 0.10, … 1.00` — the ECDF proportions, which is what would go first if the rows had been reordered without their y values travelling with them.

The gate was **relaxed by naming a second stat, not removed**: a step layer on any other computed stat still returns `unknown` and keeps its static-image fallback rather than being read on a guess.

## Verification

Full suite **5475 passed**, `man/` regenerated so the drift check stays green, `NEWS.md` updated. 21 new tests.

Two existing cases in `test-adapters.R` pinned the old decline and failed, which is the correct signal — the contract changed on purpose. They now assert the new one, with the old reasoning kept as the explanation of why the support has work to do.

Falsification:

| reverted | failures |
|---|---|
| gate back to `StatIdentity` only | **5** |
| `in_drawn_order()` not called | **8** |

Two of the new cases are guards rather than behaviour, and both would look fine while being wrong: `is.finite()` on a character vector is `FALSE` throughout, so an unguarded filter on a discrete x would delete **every** row rather than none; and a processor handed a layer index the built plot has no frame for answers instead of erroring.

---


jooyoungseo:
Item 1 was right and is fixed in `4b5a9f4` — thanks, it was a real inconsistency rather than a style point.

You are correct that the y half was redundant with `ggplot2_line_layer_processor.R:640-647` / `:704-705`, and the two filters were there for genuinely different reasons: mine for JSON-safety, the existing one for keeping `data.length` aligned with the rendered polyline so the highlight index mapping does not shift. `in_drawn_order()` now asks about **x** alone, and its docstring says why the two cases differ.

Verified the ECDF padding is unaffected: its two rows are infinite in **x** and finite in **y** (`0` and `1`), so the x-only test removes exactly the same two rows. `test-ggplot2-ecdf.R` and `test-adapters.R` both still pass (21 and 106).

Worth noting the same question came up independently in the Python binding an hour earlier and landed the same way, which is why the docstring now cross-references it. There, filtering on both coordinates **broke a test**: `seaborn.pointplot` NaN-pads a hue level missing from a category, and that padding keeps two estimate lines the same length. A row with a real x and no y has a position and no reading, which is a different thing from a row with no position at all — so x-only is the rule in both bindings now (xability/py-maidr#430).

On item 2 — agreed, and it is the reason the gate names stats rather than testing a property. Each future stat needs its own answer to "what did this stat do to the row order and the padding", and the allowlist is what forces that question to be asked instead of assumed. The test at `test-ggplot2-ecdf.R:166-176` fails if the gate is ever widened wholesale.

On item 3 — fair. I have trimmed nothing for now because each copy is load-bearing where it sits (the adapter comment explains a gate, the roxygen explains a method, `NEWS.md` addresses users), but the cross-binding JSON note is the part most likely to drift, so it now lives in full in one place and is referenced from the others.

---

Issue #170, opened 2026-08-16
geom_point emits a point ggplot2 did not draw, so every highlight after a missing value is off by one


jooyoungseo:
## Problem

A scatter layer with a missing value emits **more points than the chart draws**:

```r
df <- data.frame(x = c(1, 2, 3, 4), y = c(1, NA, 3, 4))
ggplot(df, aes(x, y)) + geom_point()
```

Measured:

```
points emitted: 4
built rows:     4    finite y: 3     ← ggplot2 removed the NA row
selectors:      1  ->  g#geom_point\.points\.2\.1 > use

payload: [{"x":1,"y":1},{"x":2,"y":null},{"x":3,"y":3},{"x":4,"y":4}]
```

ggplot2 says so itself: `Removed 1 rows containing missing values (geom_point())`. There are three `<use>` elements on the page and four entries in `data`.

## Consequence

The selector resolves to the drawn elements in order, so from the gap onward every data point is paired with the **next** observation's mark:

| data | should highlight | actually highlights |
|---|---|---|
| `x = 1` | the mark at 1 | the mark at 1 |
| `x = 2` (gap) | nothing | the mark at **3** |
| `x = 3` | the mark at 3 | the mark at **4** |
| `x = 4` | the mark at 4 | nothing left |

This is the failure worth prioritising over a crash: a sighted-with-low-vision reader is shown a mark that does not correspond to the value being announced, and nothing in the output says so.

## This package already knows the rule

`R/ggplot2_line_layer_processor.R:640-647` drops NA-y rows for exactly this reason, and says so:

> Drop rows whose y is NA. The corresponding gridSVG polyline only contains coordinates for non-NA points ... so emitting placeholder null rows here would make `data.length > polyline.points.length` and shift the MAIDR JS highlight-to-point index mapping by the number of leading NAs.

The point processor never adopted it. So the two processors in the same package disagree about the same question, and only one of them is right about the geometry.

## Suggested fix

Match the line processor: a sample ggplot2 did not draw is not emitted.

Dropping does lose the position, which is a real cost — xability/maidr#925 has just given the core a way to express "positioned, no reading", and this is the shape that wants it. But that is a `LineTrace` capability today, this is a point trace, and `MAIDR_VERSION` is pinned at `4.2.0` which predates it. A wrong highlight is worse than an absent point, so the drop is the right answer now and the richer one is a follow-up once the pin moves.

## Acceptance

- A scatter layer emits exactly as many points as ggplot2 draws
[...]

PR #175, opened 2026-08-16
fix(ggplot2): read a jittered point's data, not the offset it was drawn at

### LINKED ISSUE #174: geom_jitter announces the jitter as the data — including on the y axis, where it changes the measurement
## The defect

`position_jitter` displaces every point by a random amount so overlapping observations stay separable. The displacement is a drawing decision, not an observation — but it is what lands in the built data frame, and that is what the point layer processor reads and announces.

Measured on `ggplot(df, aes(g, v)) + geom_jitter()` against the same data drawn with `geom_point()`:

```
geom_jitter        x= 1.310872   y= -0.6257439
geom_point         x= 1.000000   y= -0.6264538   <- the actual datum
```

Two separate things are wrong, and the second is the serious one.

### 1. x — a random number where the axis has a level name

`max|dx|` measured at **0.399**, so a point can be announced almost halfway to the neighbouring category. The x axis here is discrete: its levels are `a`, `b`, `c`, drawn at 1, 2, 3. A reader navigating the chart is told `g` is `1.310872`.

This half also affects plain `geom_point()` on a discrete scale, which announces `x = 1` — honest about position, still silent about the name. That part is a grammar gap and is filed upstream as xability/maidr#927.

### 2. y — the announced value is not the value in the data

`geom_jitter()` jitters **both** axes by default: `height` defaults to 40% of the data's resolution. On continuous data the resolution is tiny and the error is negligible. On the case the geom actually exists for it is not.

A Likert-style chart, 60 responses on a 1–5 scale, three groups:

| | `max｜dx｜` | `max｜dy｜` | first point: drawn | first point: true |
|---|---|---|---|---|
| `geom_jitter()` on integer y | 0.3896 | **0.3960** | **1.1477** | 1.0000 |
| `geom_jitter(height = 0.5)` | 0.3955 | **0.4969** | **1.3736** | 1.0000 |
| `geom_jitter()` on continuous y | 0.3973 | 0.0000 | 0.3411 | 0.3411 |
| `geom_jitter(height = 0)` | 0.3935 | 0.0000 | 0.3411 | 0.3411 |

A respondent who answered **1** is announced as having answered **1.1477**. On a five-point scale that is not a possible answer — the value is not merely imprecise, it is outside the response set. Nothing in the output says so; the chart loads, sounds complete, and is wrong.

Overplotted integer data is precisely the situation `geom_jitter` is reached for, so the defect is worst exactly where the geom is most used.

## Reproduce

```r
library(ggplot2)
set.seed(1)
likert <- data.frame(g = rep(c("a","b","c"), each = 20),
                     score = sample(1:5, 60, TRUE))
p <- ggplot(likert, aes(g, score)) + geom_jitter()

res <- maidr:::Ggplot2PlotOrchestrator$new(p)$generate_maidr_data()
str(res$subplots[[1]][[1]]$layers[[1]]$data[1:3])
```

## The recovery is exact

Rebuilding the plot with the layer's position adjustment replaced by `position_identity()` returns the untouched values, in the same row order:

```r
b  <- ggplot_build(p)$data[[1]]                                   # jittered
q  <- p; q$layers <- list(geom_point(mapping = p$layers[[1]]$mapping))
qb <- ggplot_build(q)$data[[1]]                                   # true

nrow(b) == nrow(qb)   # TRUE — row order preserved
qb$y[1]               # 1  — exactly the datum
```

So the true positions are recoverable from ggplot2's own machinery rather than by trying to invert a random displacement, which cannot be done.

## Scope

- [ ] Detect a layer whose `position` is `PositionJitter` (also `PositionJitterdodge`) and read its positions from an identity-positioned rebuild rather than from the jittered frame
- [ ] Snap a discrete-scale x to its level position, so the category is at least the tick a sighted reader sees it against
- [ ] Leave continuous scales untouched — there the drawn position *is* the value, and snapping would round every measurement onto a gridline
- [ ] Build the replacement layer without mutating the caller's plot: ggplot2 `Layer` objects are ggproto and have reference semantics, so assigning `p$layers[[i]]$position` would alter the user's own object

## Related

- xability/py-maidr#439 — the same defect in the Python binding for `sns.stripplot` / `sns.swarmplot`, fixed there for x; ggplot2's y jitter has no seaborn equivalent, since neither seaborn function displaces the value axis
- xability/maidr#927 — why the level's *name* still cannot be announced
- #170 — the point layer processor and its drawn-sample filter

jooyoungseo:
Closes #174.

## What was wrong

`position_jitter()` displaces every point by a random amount so overlapping observations stay separable. The displacement is a drawing decision, not an observation — but it is what lands in the built data frame, and that is what the point layer processor read.

Measured on `ggplot(df, aes(g, v)) + geom_jitter()` against the same data drawn with `geom_point()`:

```
geom_jitter        x= 1.310872   y= -0.6257439
geom_point         x= 1.000000   y= -0.6264538   <- the actual datum
```

Two things were wrong, and the second is the serious one.

**`x`** was a random number where a discrete axis has a level, reaching `0.399` of a category width — almost halfway to the neighbouring category.

**`y`** — and this has no seaborn equivalent, since neither `stripplot` nor `swarmplot` displaces the value axis. `geom_jitter()` jitters **both** axes by default, `height` being 40% of the data's resolution, so the number announced as *the measurement* was not the measurement. On continuous data the resolution is tiny and the error is negligible. On the overplotted integer data the geom exists for it is not:

| 60 responses on a 1–5 scale | `max｜dy｜` | first point drawn | first point true |
|---|---|---|---|
| `geom_jitter()` | 0.3960 | **1.1477** | 1 |
| `geom_jitter(height = 0.5)` | 0.4969 | **1.3736** | 1 |

A respondent who answered **1** was announced as having answered **1.1477**. On a five-point scale that is not merely imprecise — it is not a possible answer. The chart loads, sounds complete, and is wrong; and it is worst exactly where the geom is most reached for.

## How it's fixed

A random displacement cannot be inverted, so the positions are recovered by asking ggplot2 to lay the layer out again with `position_identity()`. That is exact rather than approximate, and it assumes nothing about column names or row order: a position adjustment neither adds nor removes rows, so the rebuilt frame lines up one for one — and the substitution is skipped if it ever does not.

Three things this had to get right, each with a test of its own:

- **The replacement is a fresh `ggproto` child, not an assignment.** ggplot2 `Layer` objects are ggproto and have *reference* semantics, so `plot$layers[[i]]$position <- ...` would strip the jitter from the caller's own plot — the object they still hold and may draw again.
- **`PositionJitterdodge` does not inherit from `PositionJitter`.** Both descend straight from `Position`, so it is named rather than caught by inheritance.
- **Its dodge goes with its jitter**, which is safe because the level the dodge separates travels as an aesthetic of its own and is still announced. Asserted, rather than assumed — otherwise this would be trading one missing fact for another.

## Verification

| reverted | tests that fail |
|---|---|
| the recovery entirely | 7 |
| `PositionJitterdodge` dropped from the list | 2 |
| assigning to the caller's layer instead of shadowing | 2 |

Disjoint sets. The tests compare against the **source column** rather than against a rebuild, so a recovery that drifted while staying self-consistent would still fail.

**Two honest caveats:**

- The row-count guard in `undisplace_layer()` has no reproducing case — it is defensive, not covered, and the commit says so rather than implying otherwise.
- My first `position_jitterdodge` test asserted on `y`, which that position never displaces by default (`jitter.height` is 0). It passed with the class removed from the list and therefore pinned nothing. It asserts on `x` now, and a separate case sets `jitter.height` explicitly for the `y` half.

- Full suite: **5673 passed, 0 failed** (99 skipped, 8 warnings — unchanged from before)
- `man/` regenerated with roxygen2, so the drift check from #143 stays green

## What this does not fix

The category's **name**. `geom_point()` on a discrete scale announces `x = 1` — honest about position, still silent about the level. `ScatterPoint.x` is typed `number` in the grammar and the trace does arithmetic on it, so that half is filed upstream as xability/maidr#927.

## Related

- xability/py-maidr#439 — the same x-axis defect in the Python binding for `sns.stripplot` / `sns.swarmplot`
- xability/maidr#927 — the grammar gap that keeps the level's name out
- #170 — the point layer processor's drawn-sample filter, which this sits beside

---


jooyoungseo:
Both review points addressed in `f73e007`, with one correction to the diagnosis.

**The per-panel rebuild is real.** Traced `ggplot2::ggplot_build` through facets of 2, 4 and 8 panels: **3, 5 and 9** calls. One is the original build; the other P are the same answer recomputed, each laying out all P panels.

One correction: the *call count* is O(P), not O(P²) — but the characterisation of the **work** as quadratic is right, since each of those P rebuilds builds every panel. Memoised, the count is flat at **2** for all three panel counts.

Keyed by layer index, with each entry holding the layer it came from. Both halves earn their place:

- the **index** bounds the cache at the plot's layer count, and stops two jittered layers evicting each other once per panel — which is the cost being removed;
- the **layer** makes the entry self-validating. `Layer` objects are ggproto, ggproto objects are environments, and `identical()` on two environments is a pointer comparison — so it is O(1) and exact, where hashing or deep-comparing the plot could cost as much as the rebuild it saves.

That second half is not theoretical. Reverting it (keying on the index alone) fails 3 tests, including one that reads two different plots in turn: without the check, the second plot is answered with the first plot's values — every number wrong, every number plausible.

**The `layer_index < 1` guard** is added. You're right that no caller can reach it today.

Falsification: removing the memo fails the panel-count test; dropping the layer check fails 3. Full suite **5676 passed, 0 failed**.

On the "not verified" note — understood, and worth saying that the numbers above are measured on this branch rather than reasoned about.

---

jooyoungseo:
You were right, and this was a real bug I introduced with the memo. Reproduced exactly as described, fixed in `fef1e78`.

```
shared <- geom_jitter()
p1 <- ggplot(df1, aes(g, score)) + shared
p2 <- ggplot(df2, aes(g, score)) + shared

same Layer object across plots : TRUE
p1 announced == df1$score      : TRUE
p2 announced == df2$score      : FALSE
p2 announced == df1$score      : TRUE     <- the bug
```

Every one of `p2`'s announced values was `p1`'s, and the row-count check waved it through because the two frames were the same length — exactly as you predicted. My reasoning that "a layer belongs to one plot" was simply false, and ggplot2's own `?layer` docs say so.

**And your point about the test is the sharper one.** `"a cached frame is never handed to a different plot"` built two plots from two separate `geom_jitter()` calls — different environments, so it passed against the bug while appearing to cover it. That's the third test in this session I've written that named a case without exercising it. It stays as the different-layers case, and a new one shares a single layer object and asserts `identical(p1$layers[[1]], p2$layers[[1]])` *before* reading either plot, so the fixture can't silently stop being the case it claims.

**The fix** is your suggestion: clear the cache when a plot starts being processed. `Ggplot2PlotOrchestrator$initialize` is the single entry point for the faceted, patchwork and plain paths, so one call covers all three — and it's the boundary the cache was scoped to all along, since it exists so a faceted plot doesn't rebuild per panel and a panel only belongs to the run that's starting.

The layer check stays: it costs nothing and still catches the ordinary case of a second plot with its own `geom_jitter()` inside one run.

**`undisplace_layer()`** now bounds-checks before dereferencing rather than relying on the guard one call later.

Falsification: removing the reset fails the shared-layer test and nothing else. Full suite **5680 passed, 0 failed**.

On the side benefit you spotted with `x_looks_like_indices` — I hadn't noticed that, and it's a good catch: whole-number x values do now let the faceted mapping resolve category labels where fractional jittered x would have fallen through.

---

PR #177, opened 2026-08-16
fix(ggplot2): keep a chart interactive when a reference line is drawn on it

### LINKED ISSUE #176: One geom_hline turns a fully supported chart into a static image
## What happens

`geom_hline()` and `geom_vline()` classify as `"unknown"` in `detect_layer_type()`, `has_unsupported_layers()` therefore returns `TRUE`, and `should_fallback()` renders the **whole plot** as a base64 image. A supported chart loses all of its interactivity because of one annotation.

Measured with `save_html()`:

| plot | output | size |
|---|---|---|
| `geom_boxplot()` alone | interactive SVG | 44,353 bytes |
| `geom_boxplot() + geom_hline()` | **base64 image** | 14,680 bytes |
| `geom_boxplot() + geom_vline()` | **base64 image** | 14,480 bytes |
| `geom_line() + geom_hline()` | **base64 image** | 30,808 bytes |

```r
library(ggplot2)
df <- data.frame(g = rep(c("a","b"), each = 30), v = rnorm(60))

maidr::save_html(ggplot(df, aes(g, v)) + geom_boxplot(), "ok.html")     # interactive
maidr::save_html(ggplot(df, aes(g, v)) + geom_boxplot() +
                   geom_hline(yintercept = 0.5), "gone.html")
#> Warning: Plot contains unsupported elements. Rendering as static image
#>          instead of interactive MAIDR plot.
```

## Why it is worth changing

Credit where due: this is the *safe* failure. It warns, and it never announces a wrong number — unlike the defects in #170 and #174. So this is an enhancement rather than a correctness bug.

But the cost is total and the trigger is ordinary. A threshold line is one of the most common ggplot2 annotations — a target, a control limit, a prior year's median, a significance cutoff — and adding one takes a box plot that read perfectly and leaves a screen reader user with an image.

## The precedent

The Python binding decided this exact question in xability/py-maidr#434: a reference line is **decoration rather than data**, the grammar has no annotation shape to put it in, so it is left out of the layer and the chart keeps working. `MultiLinePlot._is_in_data_space()` is the filter, and the reasoning there applies unchanged here.

Reading a reference line as *data* would be worse than dropping it. On the Python side an `axhline` announced its endpoints as `0` and `1`, because a blended transform puts its coordinates in axes-fraction space — a confident reading of a series that is not there. Whatever is done here, the line should not become a `line` layer.

## Scope

- [ ] Classify `GeomHline`, `GeomVline` and `GeomAbline` as annotation rather than `"unknown"`
- [ ] `has_unsupported_layers()` ignores annotation layers, so they no longer force the fallback
- [ ] The orchestrator omits them from the emitted layers — they are not data and there is no trace type for them
- [ ] **Guard:** a plot whose layers are *all* annotation still falls back. There is nothing to read, and "no unsupported layers" must not come to mean "no layers at all"
- [ ] `annotate("text", ...)` already leaves the chart interactive (measured), so this is specifically about the reference-line geoms

## Worth deciding separately

Announcing *that* a threshold is drawn, and at what value, is genuinely useful and neither binding does it. That needs a grammar shape for annotations (xability/maidr), and it should not hold up stopping one line from costing a chart everything.

## Related

- xability/py-maidr#434 — the same question answered for matplotlib, with the measurements
- #170, #174 — the ggplot2 layer-reading defects fixed recently; this is the safe-failure counterpart

jooyoungseo:
Closes #176.

## What was wrong

`detect_layer_type()` had no branch for `GeomHline` / `GeomVline` / `GeomAbline`, so they fell through to `"unknown"` — and one `"unknown"` layer makes `has_unsupported_layers()` true, which drops the **whole plot** to a base64 image.

Measured with `save_html()`:

| plot | rendering | size |
|---|---|---|
| `geom_boxplot()` | interactive SVG | 44,353 bytes |
| `geom_boxplot() + geom_hline()` | base64 image | 14,680 bytes |
| `geom_boxplot() + geom_vline()` | base64 image | 14,480 bytes |
| `geom_line() + geom_hline()` | base64 image | 30,808 bytes |

The chart itself was fully supported. What it lost — sonification, braille, keyboard navigation, the text description — it lost to a threshold line, which is among the most ordinary things to draw on a chart: a target, a control limit, last year's median, a significance cutoff.

## The fix, in three parts

**1. `R/ggplot2_adapter.R` — tag reference lines `"skip"`.**

A reference line carries no observations, and the grammar has no annotation shape to put it in. `"skip"` is the existing mechanism that already keeps `geom_text()` and a candlestick's wick from forcing a fallback.

Skipping rather than *reading* it is the same answer the Python binding reached in xability/py-maidr#434, and for the stronger of the two reasons: an `axhline` read as a line layer there announced its endpoints as **0 and 1**, because a blended transform puts its coordinates in axes-fraction space rather than data space. That is not a partial reading — it is a confident reading of a series that is not there.

Announcing *that* a threshold is drawn, and where, is worth doing. It needs a grammar shape for annotations, and it is not a reason to keep costing a chart everything in the meantime.

**2. `R/ggplot2_plot_orchestrator.R` — fall back when every layer is skipped.**

The trap on the other side of part 1 is "no unsupported layers" quietly coming to mean "no layers at all". A plot made only of skipped layers emits zero layers, and announcing that as an interactive chart is worse than an image, because an image at least says what it is.

That case was **already live before this change**: `annotate("text")` alone reported `should_fallback()` FALSE and emitted 0 layers, because `GeomText` was already tagged `"skip"`. Measured on pristine `main`:

```
should_fallback: FALSE
layers emitted: 0
```

**3. `R/ggplot2_facet_utils.R` — give the facet path the `"skip"` guard the others have.** *(found during review)*

`process_facet_panel()` types a panel from "the first layer that actually produced a result". `create_processor()` has no `"skip"` arm, so a skipped layer fell to the switch default, got an `Ggplot2UnknownLayerProcessor`, and won that scan:

```
geom_hline() + geom_point() + facet_wrap(~f)   layers [skip skip]
geom_point() + geom_hline() + facet_wrap(~f)   layers [point point]
```

Same chart, same data, typed by which layer happened to be written first — with the point data still underneath, so it is a scatter announced under a type the core does not know. Also pre-existing: `geom_text() + geom_point() + facet_wrap()` typed both panels `"skip"` on `main`. `create_layer_processors()` and both patchwork paths already guarded this; the facet path is catching up.

## Measured after

```
boxplot alone                  svg=TRUE  img=FALSE  warn=-
boxplot + hline                svg=TRUE  img=FALSE  warn=-
point + vline                  svg=TRUE  img=FALSE  warn=-
point + abline                 svg=TRUE  img=FALSE  warn=-
hline ALONE                    svg=FALSE img=TRUE   warn=fallback
vline ALONE                    svg=FALSE img=TRUE   warn=fallback
annotate text ALONE            svg=FALSE img=TRUE   warn=fallback
boxplot + rug (unsupported)    svg=FALSE img=TRUE   warn=fallback
```

`"skip"` stays narrow: a genuinely unsupported geom is still `"unknown"` and still falls back, including when a reference line is drawn alongside it. Widening `"skip"` into a general "ignore what we do not understand" would leave charts announcing a partial reading as a complete one, which is what the fallback exists to prevent.

## Tests

`tests/testthat/test-ggplot2-reference-lines.R` — **32 assertions**, driven through `save_html()` so what is asserted is what a reader receives.

Beyond "does not fall back", which is satisfiable by rendering an interactive shell with nothing in it, the box layer is asserted to carry **identical `$data`** with and without the threshold. The faceted cases pin the panel type as a *value*, since two orderings agreeing on the wrong type would otherwise pass.

### Falsification

Each part reverted separately, against the same file:

| reverted | result |
|---|---|
| the `"skip"` classification | **FAIL 15** — every "+ reference line" case, plus the adapter's three type assertions |
| the all-skip fallback guard | **FAIL 7** — the three annotation-only cases, and the `should_fallback()` assertion |
| the facet `"skip"` guard | **FAIL 3** — both orderings, and the pre-existing `geom_text` case |
| nothing | **FAIL 0 / PASS 32** |

The failure sets are disjoint apart from one test that genuinely depends on the first two — it asserts the annotation-only plot falls back *because it reads as nothing*, which needs the line skipped and the guard present.

Full suite: **5712 passed, 0 failed** (5680 before). `roxygen2::roxygenise()` produces no `man/` drift, no line exceeds the 100-char `.lintr` limit, and `NEWS.md` records the fix under `## Bug Fixes`.


jooyoungseo:
`NEWS.md` entry added in `828bddc` — you're right about the convention, and this is exactly the kind of symptom the existing entries describe.

Your three structural checks all hold, and I confirmed the third by running the suite you couldn't: **5712 passed, 0 failed**.

## What checking the fourth path turned up

You verified `"skip"` was already honoured in `create_layer_processors()` and in both patchwork paths. There is a fourth caller you didn't reach — `process_facet_panel()` in `R/ggplot2_facet_utils.R` — and it was **not** guarded. `create_processor()` has no `"skip"` arm, so a skipped layer falls to the switch default, gets an `Ggplot2UnknownLayerProcessor`, and its result then wins the "first layer that actually produced a result" scan that types the panel.

Measured on a faceted scatter:

```
geom_hline() + geom_point() + facet_wrap(~f)   layers [skip skip]
geom_point() + geom_hline() + facet_wrap(~f)   layers [point point]
```

Same chart, same data, typed by which layer happened to be written first — and the panel still carries the *point* data underneath, so it is a scatter announced under a type the core does not know.

**Not something this PR introduced.** `GeomText` has been tagged `"skip"` all along, so I checked the same ordering on pristine `main`:

```
geom_text() + geom_point() + facet_wrap(~f)   ->  skip skip
```

Pre-existing. Skipping reference lines only makes it far easier to reach, since a threshold on a faceted chart is much more ordinary than a text layer written before the data layer. Fixed in `dc3cca5` by giving the facet path the guard the other three already had.

Four tests cover it: both orderings agreeing, the line-first ordering pinned to `c("point", "point")` as a value (two orderings agreeing on the *wrong* type would otherwise pass), the `geom_text` ordering that predates this PR, and a faceted annotation-only plot still falling back. Reverting the guard alone fails 3 of them and nothing else.

Updated totals: **32 assertions** in `test-ggplot2-reference-lines.R`, full suite **5712 passed, 0 failed**, no `man/`/`NAMESPACE` drift, no line over the 100-char limit.

---

Issue #178, opened 2026-08-16
A faceted categorical scatter puts a string in a point's x, where the grammar requires a number


jooyoungseo:
## The defect

The same chart emits two different shapes depending on whether it is faceted:

```r
df <- data.frame(g = rep(c("a","b","c"), each = 6), v = rnorm(18), f = rep(c("p","q"), 9))

ggplot(df, aes(g, v)) + geom_jitter()                    # x = 1     (mapped_discrete)
ggplot(df, aes(g, v)) + geom_jitter() + facet_wrap(~f)   # x = "a"   (character)
```

Measured on the first point of each. The faceted path runs `apply_scale_mapping()`, which replaces the numeric position with the category label:

```r
mapped_values <- scale_mapping[char_values]
```

so the position is gone and the name is in its place.

## Why the string is the wrong half to keep

`ScatterPoint.x` is typed `number` in the grammar, and `ScatterTrace` does arithmetic on it:

```ts
const sortedByX = [...data].sort((a, b) => a.x - b.x || a.y - b.y);
this.xIndexByValue = new Map(this.xValues.map((x, index) => [x, index]));
const distance = Math.hypot(center.x - _x, center.y - _y);
```

With `x` a string, `a.x - b.x` is `NaN`. A comparator returning `NaN` leaves `Array.prototype.sort` with no ordering information, so the points stay in input order rather than the x order every downstream index assumes; and `Math.hypot(NaN, …)` is `NaN`, so `findNearestPoint` — the hover/highlight resolver — has no nearest point to find.

So a faceted categorical scatter announces the right *name* while giving the core a payload it cannot sort, index or highlight against. The unfaceted one gives a usable position and (until now) no name.

## What changed around it

`ScatterPoint.xLabel` / `yLabel` exist as of xability/maidr#927, so the name no longer has to displace the position — it travels alongside it. The unfaceted point path now emits both, and the Python binding does the same for `sns.stripplot` and `sns.swarmplot` (xability/py-maidr#439).

That leaves the faceted path as the odd one out, and it is the one whose shape the core cannot use.

## The fix

Have the faceted path emit what the unfaceted one now does: the numeric position in `x`, the name in `xLabel`. `discrete_axis_labels()` / `category_at()` already do the lookup, so the change is to stop `apply_scale_mapping()` overwriting the position for point layers rather than to add new machinery.

[...]

PR #179, opened 2026-08-16
feat(ggplot2): name the category a point sits in

jooyoungseo:
## What was wrong

ggplot2 maps a discrete scale onto consecutive integers and marks the result `mapped_discrete`, so a point in category "a" arrives as `x = 1`. That integer is what the point layer emitted:

```
geom_jitter(x = g)   first point  x = 1 (mapped_discrete)
geom_point(x = g)    first point  x = 1 (mapped_discrete)
```

A reader heard **"g is 1"** where the chart says **"a"**. Not a partial reading — the number is an internal coordinate, announced under the axis' own label as though it were the observation.

#174 fixed the neighbouring half for `geom_jitter` (the *displacement* was being announced as the measurement) and left the naming open, because the grammar had nowhere to put it. xability/maidr#927 has since added `ScatterPoint.xLabel` / `yLabel`, so the name now travels **alongside** the position rather than in place of it. The position stays numeric because the core sorts on it, measures distance with it, and groups columns by it.

The Python binding emits the same two fields for `sns.stripplot` and `sns.swarmplot` (xability/py-maidr#439).

## Measured after

```
geom_jitter (x=g)     x=1         xLabel=a   yLabel=-
geom_point  (x=g)     x=1         xLabel=a   yLabel=-
horizontal  (y=g)     x=1.370958  xLabel=-   yLabel=a
position_dodge        n=24        named=24   (0.875, 1.125, 1.875, 2.125)
numeric both axes     x=1.5       xLabel=-   yLabel=-
```

## Reading the scale took two attempts

Both are recorded in the tests, because both were wrong in a way that looked right:

- **`get_breaks()` returns the level *names*** on a discrete scale (`"a" "b" "c"`), carrying the integer positions in a `pos` attribute. `break_positions()` is the other candidate and is the wrong one — it rescales to 0..1 (`0.1875, 0.5, 0.8125` for three levels), which is not the space the layer's own `x` is in.
- **A missing position needs `[`, not `[[`.** `[[` *throws* "subscript out of bounds" where `[` answers `NA`.

## A displaced point is named by the slot it was moved within

My first version matched positions **exactly**, on the argument that a point drawn off a tick "was moved there deliberately, and naming it after the nearest tick would claim a category the renderer did not put it in."

That argument is wrong, and review caught what it cost. Measured on a dodged scatter over two groups:

```
plain point       n=24  named=24  distinct_x=1,2
position_dodge    n=24  named=0   distinct_x=0.875,1.125,1.875,2.125
```

**Zero of 24 named.** `position_dodge` shifts a point sideways to make room for a **sibling series within the same category**, and `position_jitter` scatters it **inside its own slot**. Neither moves a point out of its category, so the tick it was moved from *is* the category it is in — naming it is correct, not a false claim.

That is also the answer the Python binding already reached for the identical situation, in `_named_coordinate`: *"still that group, and still named by the tick it was moved from."* Two bindings were disagreeing on the same question with the better-argued one already written down.

Now rounds, **bounded at half a tick**, which is what keeps it honest: ggplot2 keeps both displacements inside the slot (a dodge divides the category's width, `position_jitter` defaults to 40% of the resolution), so anything further out is not a displaced member and stays unnamed.

## The guard that matters most

**A continuous axis is declined outright**, on the scale's own `is_discrete()`. A continuous axis has breaks and labels too — "0", "25", "1.00" — but those are formatted renderings of the numbers rather than names for them, so substituting one would cost the value both its type and its precision. It is also what makes the rounding above safe: on a continuous axis there are no names to round onto, so a measurement cannot be renamed after whichever tick it falls nearest.

## Tests

`tests/testthat/test-ggplot2-point-category-label.R` — **26 assertions**.

The pairing is asserted, not just presence: `{"1 a", "2 b", "3 c"}`. A layer that named every point "a" would satisfy a set-of-labels check and be useless.

Guards: the measurement axis is not named in either orientation; a numeric scatter carries no names and keeps its exact values (`1.5`, `2.5` unrounded); the helper declines a continuous axis; a position past half a tick is left unnamed; a real dodged scatter is named end to end.

### Falsification

Removing the two `xLabel`/`yLabel` assignments fails **4** and leaves the guards passing — the guards are about what must not change, so a fix that broke them would be caught from the other side.

## A regression I introduced and caught

The first version made the suite go from **8 warnings to 228** — 220 from the jitter tests alone, all "NAs introduced by coercion".

The cause is worth stating because it turned up a real defect. On the **faceted** path `apply_scale_mapping()` replaces the numeric position with the category label *before* emission, so `x` arrives as the character `"a"` rather than the number `1`, and `as.numeric("a")` warns once per point.

The lookup now declines a non-numeric position, and two tests pin that it stays silent — including one that renders a faceted jitter and asserts `expect_silent`.

But the underlying disagreement is a defect of its own: `ScatterPoint.x` is typed `number`, and with a string there `a.x - b.x` is `NaN`, so the core's sort has no ordering information and `Math.hypot` gives it no nearest point to highlight. **Filed as #178** rather than fixed here — it needs `apply_scale_mapping`'s other callers checked first (a bar's `x` is legitimately `string | number`, so the same call may be right there and wrong only here).

## Verification

- full suite: **5738 passed, 0 failed**, and **8 warnings** — the pre-change baseline
- `roxygen2::roxygenise()` — clean, two new `man/` pages committed
- no line over the 100-char `.lintr` limit in the changed files


jooyoungseo:
The `position_dodge` observation is the important one, and checking it showed my reasoning was wrong. Fixed in `10926ad`.

## What you spotted

Measured exactly as you predicted — a dodged scatter over two groups:

```
plain point       n=24  named=24  distinct_x=1,2
position_dodge    n=24  named=0   distinct_x=0.875,1.125,1.875,2.125
```

**Zero of 24 named.** You flagged it as "a real-world case this PR doesn't attempt to cover, consistent with the stated design". It is consistent with what I wrote — but what I wrote was wrong.

I justified the exact match as: *"a point drawn off one has been moved there deliberately… naming it after the nearest tick would claim a category the renderer did not put it in."* That is false for both displacements this could meet. A dodge shifts a point sideways to make room for a **sibling series within the same category**; a jitter scatters it **inside its own slot**. Neither moves a point out of its category, so naming it after the tick it was moved from is *correct*, not a claim the renderer never made.

The Python binding had already reached that conclusion for the identical situation, in `_named_coordinate`:

> A string-category axis puts its groups on consecutive integers, so a point drawn *off* one is a group a `dodge` shifted aside to make room for its neighbour — still that group, and still named by the tick it was moved from.

So I had two bindings disagreeing on the same question, with the better-argued one already written down, and I did not check.

**Now rounds, bounded at half a tick.** ggplot2 keeps both displacements inside the slot — a dodge divides the category's width, `position_jitter` defaults to 40% of the resolution — so anything further out is not a displaced member and stays unnamed. The `is_discrete()` gate you singled out does the rest: on a continuous axis there are no names to round onto, so a measurement still cannot be renamed after whichever tick it falls nearest.

After: **24 of 24 named**, `a` and `b` correctly split across the four dodged positions.

Three tests replace the one that asserted the old rule: a displaced point is named by its slot; a position past half a tick is not; and a real dodged scatter is named end to end rather than silently unlabelled.

## The other two

- **`stats::setNames` vs bare** — agreed, and thanks for checking whether it was mine. It is pre-existing either way, so I have left it rather than making this PR touch a convention it did not introduce.
- **The duplicated header comment** — fair. I have left it: the two say different things to different readers (the source comment is *why the field exists*, the test comment is *what was measured before and after*), and a pointer would cost a reader in the test file a file-open to learn what the test is defending. Happy to trim it if you would rather.

Full suite after: **5738 passed, 0 failed**, 8 warnings — the pre-change baseline. `roxygen2::roxygenise()` clean.

---

PR #180, opened 2026-08-17
fix(ggplot2): keep a faceted point's position, and carry its name beside it

### LINKED ISSUE #178: A faceted categorical scatter puts a string in a point's x, where the grammar requires a number

jooyoungseo:
Closes #178.

The same chart emitted two different shapes depending on whether it was facetted:

```r
ggplot(df, aes(g, v)) + geom_jitter()                    # x = 1     (mapped_discrete)
ggplot(df, aes(g, v)) + geom_jitter() + facet_wrap(~f)   # x = "a"   (character)
```

`ScatterPoint.x` is typed `number` in the grammar, and `ScatterTrace` does arithmetic on it — it sorts with `a.x - b.x`, indexes columns by the value, and resolves the nearest point with `Math.hypot`. A string makes the subtraction `NaN`, and a comparator returning `NaN` leaves `Array.prototype.sort` with no ordering to apply, so the points stay in input order rather than the x order every downstream index assumes; `Math.hypot(NaN, …)` is `NaN`, so `findNearestPoint` has no nearest point to find.

The faceted chart announced the right **name** while handing the core a payload it could not sort, index or highlight against. The unfaceted one gave a usable position and, since #179, the name as well.

## The fix is a removal

`ScatterPoint.xLabel` exists as of xability/maidr#927, and `Ggplot2PointLayerProcessor` already fills it from `discrete_axis_labels()`. So the name never had to displace the position — the relabelling block simply goes, and both paths emit what the unfaceted one already did:

```
                     before              after
jitter unfaceted     x=1,   xLabel=a     x=1, xLabel=a
jitter faceted       x="a", xLabel=—     x=1, xLabel=a
point  faceted       x="a", xLabel=—     x=1, xLabel=a
continuous faceted   x=-0.626            x=-0.626   (unchanged)
bar    faceted       x="a"               x="a"      (unchanged)
```

## The issue names the wrong branch

Worth recording, since the next person will read #178 before the code. It attributes this to `apply_scale_mapping()`, and that call **cannot run here**: `R/ggplot2_facet_utils.R:188` passes `scale_mapping = NULL` for every panel, so the branch guarded on it is unreachable from a faceted layer.

Measured rather than reasoned. Restoring the `apply_scale_mapping()` call alone leaves **all 22** assertions in the new test file passing. What actually relabelled is the other branch:

```r
layer_data$x <- x_values[layer_data$x]
```

which indexed the panel's sorted category values by the drawn position. Restoring **that** alone fails all eleven test blocks. The code comment says so at the point of the removal, so a reader who arrives via the issue is not sent to a function that never ran.

## Scope

Confined to the point processor. `Ggplot2BarLayerProcessor` calls the same helper and keeps it — a bar's `x` is `string | number` in the grammar, and a bar chart is navigated by category rather than by distance, so nothing there subtracts one x from another. A faceted bar is pinned as unchanged in the last test.

Two branches disappear with the relabelling that were hazards in their own right: the `as.character(layer_data$x)` fallback, which stringified an x whose column could not be found, and the index lookup itself, which was guarded by a heuristic (`all(x == floor(x)) && min(x) >= 1 && …`) that a continuous axis could satisfy by coincidence.

The `category_at()` comment is updated with it. It described the faceted path as handing the lookup a character and named the disagreement as filed-not-fixed; both paths now hand it a number. The non-numeric guard stays — it is the honest answer for any caller, and a bar layer's `x` legitimately is not one.

## Tests

`tests/testthat/test-ggplot2-faceted-point-position.R`, 22 assertions across 6 blocks: the position's type on both paths, the name carried beside it, the shape being *the same* faceted or not, each panel naming its own categories against the positions ggplot2 mapped them to, a continuous faceted scatter untouched, and a faceted bar unchanged.

Full suite: **0 failures**, 99 skipped (all `patchwork`, which is not installed here), 8 warnings — the same 8 as before this branch, and the count #178 mentions as the baseline the label lookup temporarily took to 228.

No `man/` regeneration: no roxygen block changed, only a function body and two comments.

---


jooyoungseo:
Thanks for tracing the misattribution independently rather than taking the PR body's word for it — that is the check worth doing here, since the whole point of the correction is that #178 named a function that cannot run.

Two notes, one of which extends your finding.

## The dead code is wider than the point processor

I checked your `extract_scale_mapping()` observation and it holds — and it is not confined to this processor. **No caller anywhere passes a non-`NULL` mapping**, so `apply_scale_mapping()` is unreachable from every processor that has a branch behind it:

| file | unreachable branch |
|---|---|
| `R/ggplot2_line_layer_processor.R` | 173-174, 247-248 |
| `R/ggplot2_bar_layer_processor.R` | 170 |
| `R/ggplot2_point_layer_processor.R` | removed here |

That changes one line of reasoning in the PR body, so it is worth stating plainly: I wrote that the bar processor "calls the same helper and is right to." The *reasoning* stands — a bar's `x` is legitimately `string | number` and a bar chart is navigated by category, so relabelling would be sound there — but the branch is unreachable in the bar processor too, so keeping it is not load-bearing. A faceted bar's `x = "a"` comes from `Ggplot2BarLayerProcessor`'s own extraction, not from that call. The last test in this file pins the behaviour either way, which is what actually matters for this PR.

Filed as #181 with the call-site evidence and a suggested scope, rather than widening this PR — agreed it does not belong here. Worth removing rather than leaving, I think, for the reason this PR exists: the plumbing cost real time twice in the same week, once in #178's diagnosis and once in the `category_at()` comment that repeated it. A guard that looks load-bearing and cannot execute invites exactly that.

## The unused parameter

Agreed on leaving `scale_mapping` in the signature — every processor takes it, and dropping it from one would be the odd one out. #181 proposes removing it from all of them together, which is the version that does not break parity.

## Verification

You noted no R toolchain was available to the job. I ran the suite locally against this branch: **0 failures**, 99 skipped (all `patchwork`, not installed here), 8 warnings — the same 8 as before the branch, and the number #178 cites as the baseline that the `xLabel` lookup temporarily took to 228. R 4.3.3 / ggplot2 3.4.4.

---

PR #193, opened 2026-08-20
feat(ggplot2): read geom_raster as the heatmap it draws

### LINKED ISSUE #192: `geom_raster()` is unread although it computes the same grid `geom_tile()` does
`geom_raster()` is ggplot2's fast path for a heatmap on a regular grid — its own documentation recommends it over `geom_tile()` whenever the cells are evenly spaced. r-maidr reads the `geom_tile()` spelling and not this one, so the same chart is accessible or not depending on which the author reached for.

## Measured

ggplot2 3.4.4, R 4.3.3, one 3 × 2 grid of values, through `save_html`:

| call | computed data columns | layers |
|---|---|---|
| `geom_tile(aes(g, h, fill = v))` | `fill,x,y,PANEL,group,xmin,xmax,ymin,…` | `heat` |
| `geom_raster(aes(g, h, fill = v))` | `fill,x,y,PANEL,group,xmin,xmax,ymin,…` | **none** |

The computed columns are identical — `ggplot_build` gives both the same `x`/`y` and the same `xmin`/`xmax`/`ymin`/`ymax` cell bounds, because a raster *is* a tile grid with the spacing assumed regular rather than derived per cell. Nothing the heatmap processor reads is missing.

## Why

`ggplot2_adapter.R:142` matches on the geom's class name:

```r
if (geom_class %in% c("GeomTile", "GeomBin2d")) {
  return("heat")
}
```

and `GeomRaster` is not one of them. It is not caught by inheritance either, which is the part worth measuring rather than assuming:

```
GeomRaster class chain:  GeomRaster < Geom < ggproto < gg
GeomTile   class chain:  GeomTile < GeomRect < Geom < ggproto < gg
```

`GeomRaster` does **not** descend from `GeomTile`. It is a sibling that reimplements drawing with a single raster grob, so no `inherits()` test against `GeomTile` would have picked it up either.

## Proposed

Add `"GeomRaster"` to that branch. The comment above it already records the same reasoning for the same situation — `geom_bin_2d()` got its own `GeomBin2d` class in ggplot2 4.0 where 3.x drew it with a plain `GeomTile`, and *"It is the same tile grid and still a heatmap, so both names land here. Matched by name rather than by `inherits()`."* A raster is the third name for that grid.

Worth checking the highlighting alongside the reading, and this is where the two spellings genuinely differ: `geom_tile()` draws one `<rect>` per cell, while `geom_raster()` draws **one raster grob for the whole grid**. So the reading transfers exactly and the per-cell selectors may not — which would make this the highlight-only blind spot xability/maidr#814 names, where audio, text and braille all read correctly from the unaffected side and nothing in an accessibility-focused suite can see that the wrong thing lit up. If per-cell highlighting is not recoverable from a single grob, that is worth saying out loud in the reading rather than shipping a chart that announces correctly and highlights nothing.

## Found alongside

A sweep of 19 ggplot2 geoms through `save_html`. The others that read nothing today, for the record and not for this issue:

- **`geom_segment` / `geom_curve`** — a segment whose ends share a coordinate is an interval in a lane, which is a gantt; xability/maidr#1094 and #1100 settled that reading for the Observable adapter and the trace already exists.
- **`geom_rect`** — the same interval shape with both bounds given.
- **`geom_contour` / `geom_density_2d`** — MAIDR has a contour trace (xability/maidr#802), and ggplot2 computes the `level` column, so the value is a number rather than a colour.
- **`geom_dotplot`**, **`geom_quantile`** — a dot histogram and a set of quantile curves.
- **`geom_rug`**, **`geom_spoke`** — declined for the reasons already established elsewhere: a rug is a one-dimensional distribution with no second field, and a spoke encodes a direction the grammar has nowhere to put.

jooyoungseo:
Closes #192.

`geom_raster()` is what ggplot2's own documentation steers users to whenever a heatmap's cells are evenly spaced. It was read as nothing — the chart came out as a static picture — while the `geom_tile()` spelling of the same chart read as `heat`.

## Measured

ggplot2 3.4.4, R 4.3.3, one 3 × 2 grid, through `save_html`:

| call | computed columns | layers | rendered |
|---|---|---|---|
| `geom_tile(aes(g, h, fill = v))` | `fill,x,y,PANEL,group,xmin,xmax,ymin,…` | `heat` | 64,471 bytes |
| `geom_raster(aes(g, h, fill = v))` | `fill,x,y,PANEL,group,xmin,xmax,ymin,…` | **none** | 16,612 bytes — the fallback page |

The computed columns are identical. `ggplot_build` gives both the same `x`/`y` and the same `xmin`/`xmax`/`ymin`/`ymax` cell bounds, because a raster *is* a tile grid with the spacing assumed regular rather than derived per cell. Nothing the heatmap processor reads was missing.

## Why it was missed

`ggplot2_adapter.R` classifies by the geom's class name, and `GeomRaster` was not one of them. Nor would `inherits()` have helped, which is the part worth measuring rather than assuming:

```
GeomRaster class chain:  GeomRaster < Geom < ggproto < gg
GeomTile   class chain:  GeomTile < GeomRect < Geom < ggproto < gg
```

`GeomRaster` is a **sibling** of `GeomTile`, not a subclass — it reimplements drawing from scratch as a single grob. So this is the third name on the same branch, for the reason the comment above it already gives for the second: ggplot2 4.0 gave `geom_bin_2d()` its own `GeomBin2d` where 3.x used a plain `GeomTile`, and *"it is the same tile grid and still a heatmap, so both names land here."*

## What does not transfer, and why it ships anyway

The single grob is also what it costs. Measured on the same two charts after the change:

| | `<rect>` | `<image>` | layer selectors |
|---|---|---|---|
| `geom_tile` | 11 | 1 | `g#geom_rect\.rect\.2\.1 > rect` |
| `geom_raster` | 5 | 2 | **none** |

The raster's grid is that second `<image>`. There is no per-cell element for a selector to name, so **the chart announces every value and highlights nothing.**

This is exactly the blind spot xability/maidr#814 names — audio, text and braille all read correctly, so nothing in an accessibility-focused suite can see that the wrong thing (or nothing) lit up. It is the same shape as #145.

I shipped the reading anyway, and want to be explicit about the judgement rather than bury it:

- a blind reader had a static image and now has every number in the grid — a strict gain;
- a low-vision reader had no highlight before either, because there was no layer at all — so nothing is lost;
- `geom_tile()` draws the same chart with a `<rect>` per cell and reads identically, so anyone who needs both has a supported way to get it.

Withholding the reading would trade a real gain for no gain. But it is a genuine limit, so it is asserted in the tests rather than described in prose — `test_that("a raster announces its values but has nothing to highlight")` fails the day a ggplot2 or renderer starts emitting per-cell elements, which is the reminder to give it selectors and delete the test.

## Verification

Full suite: no failures (`raster-heatmap: ........`, 8 passing; the two `W` markers elsewhere are pre-existing ggplot2 missing-value warnings).

Falsified by mutation — removing `"GeomRaster"` from the branch reddens **6 of the 8**:

```
== MUTATION: GeomRaster removed ==   [ FAIL 6 | WARN 0 | SKIP 0 | PASS 2 ]
== restored ==                       [ FAIL 0 | WARN 0 | SKIP 0 | PASS 8 ]
```

The highlighting test needed a second pass to be honest about this: it originally passed under the mutation, because `NULL$selectors` is also `NULL` — a chart that emitted no layer at all looked like a chart that emitted one without selectors. It now asserts the layer exists first.

## Found alongside, not in this PR

A sweep of 19 ggplot2 geoms through `save_html`. Also unread today:

- **`geom_segment` / `geom_curve` / `geom_rect`** — a segment whose ends share a coordinate is an interval in a lane, and the gantt trace already exists; xability/maidr#1094 and #1100 settled that reading for the Observable adapter.
- **`geom_contour` / `geom_density_2d`** — MAIDR has a contour trace (xability/maidr#802), and ggplot2 computes a `level` column, so the value is a number rather than a colour.
- **`geom_dotplot`, `geom_quantile`**.
- **`geom_rug`, `geom_spoke`** — declined on the reasoning already established: a rug is a one-dimensional distribution with no second field, and a spoke encodes a direction the grammar has nowhere to put.

All recorded in #192.

---

### DIFF
--- a/R/ggplot2_adapter.R
+++ b/R/ggplot2_adapter.R
@@ -139,7 +139,22 @@ Ggplot2Adapter <- R6::R6Class(
       # inherits(), to match every other branch in this function and because
       # the symbol does not exist on 3.x. The same release left
       # `stat_summary_2d()` on GeomTile, so nothing else moves with it.
-      if (geom_class %in% c("GeomTile", "GeomBin2d")) {
+      #
+      # `geom_raster()` is the third name for that grid, and the one ggplot2's
+      # own documentation recommends whenever the cells are evenly spaced.
+      # `ggplot_build` computes the same columns for it -- the same `x`/`y` and
+      # the same `xmin`/`xmax`/`ymin`/`ymax` bounds -- so the reading is the
+      # tile reading unchanged. inherits() would not have found it either:
+      # GeomRaster descends from Geom directly, a sibling of GeomTile rather
+      # than a subclass, because it reimplements drawing as a single grob.
+      #
+      # That single grob is also what it costs. The SVG holds one `<image>`
+      # for the whole grid where the tile spelling holds a `<rect>` per cell,
+      # so there is no element for a per-cell selector to name and the chart
+      # announces its values while highlighting nothing (#192). Read anyway:
+      # a reader who had a picture now has every number, and `geom_tile()` is
+      # there for anyone who needs both.
+      if (geom_class %in% c("GeomTile", "GeomBin2d", "GeomRaster")) {
         return("heat")
       }
 
--- a/R/ggplot2_adapter.R
+++ b/R/ggplot2_adapter.R
@@ -139,21 +139,6 @@ Ggplot2Adapter <- R6::R6Class(
       # inherits(), to match every other branch in this function and because
       # the symbol does not exist on 3.x. The same release left
       # `stat_summary_2d()` on GeomTile, so nothing else moves with it.
-      #
-      # `geom_raster()` is the third name for that grid, and the one ggplot2's
-      # own documentation recommends whenever the cells are evenly spaced.
-      # `ggplot_build` computes the same columns for it -- the same `x`/`y` and
-      # the same `xmin`/`xmax`/`ymin`/`ymax` bounds -- so the reading is the
-      # tile reading unchanged. inherits() would not have found it either:
-      # GeomRaster descends from Geom directly, a sibling of GeomTile rather
-      # than a subclass, because it reimplements drawing as a single grob.
-      #
-      # That single grob is also what it costs. The SVG holds one `<image>`
-      # for the whole grid where the tile spelling holds a `<rect>` per cell,
-      # so there is no element for a per-cell selector to name and the chart
-      # announces its values while highlighting nothing (#192). Read anyway:
-      # a reader who had a picture now has every number, and `geom_tile()` is
-      # there for anyone who needs both.
       if (geom_class %in% c("GeomTile", "GeomBin2d", "GeomRaster")) {
         return("heat")
       }
--- a/tests/testthat/test-raster-heatmap.R
+++ b/tests/testthat/test-raster-heatmap.R
@@ -1,0 +1,118 @@
+# A raster is the tile grid ggplot2 recommends for regular spacing (issue #192)
+#
+# `geom_raster()` is what ggplot2's own documentation steers users to whenever
+# a heatmap's cells are evenly spaced -- it draws the grid as one raster grob
+# instead of one rectangle per cell, which is much faster. The chart is the
+# same chart, and `ggplot_build` computes the same columns for both: `fill`,
+# `x`, `y`, and the `xmin`/`xmax`/`ymin`/`ymax` cell bounds.
+#
+# It was read as nothing at all, because the classifier matches the geom's
+# class name and `GeomRaster` was not one of the names. Nor would inheritance
+# have caught it:
+#
+#   GeomRaster   <  Geom      <  ggproto  <  gg
+#   GeomTile     <  GeomRect  <  Geom     <  ggproto  <  gg
+#
+# `GeomRaster` is a sibling of `GeomTile`, not a subclass -- it reimplements
+# drawing from scratch. So the fix is the third name on the same branch, for
+# the reason the comment above it already gives for the second: ggplot2 4.0
+# gave `geom_bin_2d()` its own `GeomBin2d` where 3.x used a plain `GeomTile`,
+# and "it is the same tile grid and still a heatmap, so both names land here".
+#
+# What does NOT transfer is the highlighting, and that is asserted here rather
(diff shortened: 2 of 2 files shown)


jooyoungseo:
Added in fde6678 — you were right that it belonged, and for a reason worth stating: every other test in the file reaches the classification *through* `save_html`, so a regression in the branch surfaces as three failures about grids and selectors rather than one about the thing that actually broke. Asked directly it also covers `geom_tile` again, the way the bin2d test does, so widening the branch cannot quietly narrow it.

It earns its place on the mutation table too. Removing `"GeomRaster"` now reddens **7 of 10** rather than 6 of 8, and the first failure names the classifier.

Thank you for tracing the downstream path rather than taking the description's word for it. One of your findings is load-bearing for a claim I made and could only assert indirectly: that `generate_selectors()` falls through to `selectors <- list()` when it finds no `geom_rect` children, rather than erroring or fabricating one. That is precisely what makes *"announces values, nothing to highlight"* a measured outcome instead of an aspiration, and it is better evidence than my own test, which observes the empty result without knowing why it is empty.

On `NEWS.md` — agreed, and left alone deliberately for the reason you identified: the preceding commit did not add one either, so adding one here would make the file inconsistent in the other direction. Batching before a release seems like the maintainer's call.

## CI

`macos-15-intel (release)` is red, and it is not this change. It failed **78 seconds in**, before R was installed and long before any test ran:

```
Error: Failed to resolve R version release at
https://api.r-hub.io/rversions/resolve/release/mac/x86_64.
```

That is `r-lib/actions/setup-r` failing to reach the r-hub version resolver. `ubuntu-latest (release)`, `ubuntu-latest (oldrel-1)`, `test-coverage` and `pkgdown` are all green on the same commit. I tried to re-run the failed job and GitHub declined with *"This workflow is already running"* — Windows was still in progress — so I will re-run it once the workflow settles, and report what it does.

Locally: 10 passing in this file, and no failures across the whole suite.

---

Issue #195, opened 2026-08-20
`gridSVG` cannot export a `curve` grob, so `geom_curve()` cannot be read at all


jooyoungseo:
Found while implementing #194. `geom_curve()` computes exactly the same four columns as `geom_segment()` — `x`, `xend`, `y`, `yend` — so it would read as the same gantt, with the curvature being a drawing instruction that never becomes a position. It cannot ship, and the blocker is not in this package.

## Measured

ggplot2 3.4.4, gridSVG 1.7-5, R 4.3.3. Same four-interval chart, drawn both ways:

| geom | grob under the panel | `gridSVG::grid.export()` |
|---|---|---|
| `geom_segment()` | `segments  GRID.segments.1` | exports; one element per interval, ids `GRID.segments.1.1.1` … `.4` |
| `geom_curve()` | `curve  GRID.curve.85` | **`Error in svgStyleAttributes(style, svgdev) : All SVG style attribute values must have length 1`** |

## Why it matters more than it looks

An unsupported layer takes the static-image fallback, which does not go through `grid.export()` — so today a `geom_curve()` chart quietly renders as a picture and nothing raises. The moment the layer is *claimed*, the same call takes the gridSVG path and `save_html()` throws:

```
save_html(ggplot(tasks) + geom_curve(aes(x = start, xend = end, y = task, yend = task)), f)
# Error: All SVG style attribute values must have length 1
```

So claiming `GeomCurve` turns a chart that renders into one that errors. That is why #194's implementation reads `geom_segment()` only, and says so in the branch rather than leaving the omission to look like an oversight.

## Where the fix goes

This is the same family as the repair already in `R/svg_utils.R` for gridSVG 1.7.7's `justTovjust()` crash on an `NA` justification — an upstream incompatibility worked around locally, with a note saying to drop the repair if gridSVG ever handles it. A `curve` grob's style is arriving with a length > 1 somewhere; finding which attribute and normalising it before export would unblock the reading, and `geom_curve()` would then need only its name added to the adapter's segment branch and to `find_segments_name()`'s position count.

Worth checking whether the exported element ids follow the same `<grob>.1.<n>` shape a `segments` grob's do before assuming the selectors transfer — a curve is drawn as an X-spline and may not be one element per row.

PR #196, opened 2026-08-20
feat(ggplot2): read geom_segment as the gantt of intervals it draws

### LINKED ISSUE #194: `geom_segment()` and `geom_curve()` draw gantt charts that read as nothing
The first of the readings #192 recorded alongside the raster, now that #193 has landed. ggplot2 3.4.4, R 4.3.3.

## The chart

`geom_segment()` with the two ends of a span on one axis and a lane on the other is how ggplot2 draws a schedule, a range plot and a high–low chart:

```r
ggplot(tasks) + geom_segment(aes(x = start, xend = end, y = task, yend = task))
```

Today that is a static picture. `save_html` emits the fallback page — no layers at all.

## The numbers are already there

`ggplot_build` computes both ends and the lane, exactly:

```
   x xend y yend group
1  0    3 1    1     1        # design, 0 to 3
2  3    8 2    2     2        # build,  3 to 8
3  8   11 3    3     3        # test,   8 to 11
4 12   15 2    2     2        # build, 12 to 15
```

and the panel's scale names the lanes, in drawn order and keyed by exactly those positions:

```r
b$layout$panel_params[[1]]$y$get_labels()
# "design" "build" "test"
```

`GanttPoint` claims `{x, start, end}` — a lane and two ends — which is precisely what the four rows above hold. Nothing is inverted from a pixel and nothing is inferred. MAIDR has had the gantt trace since xability/maidr#801, and the same reading was settled twice for the Observable adapter: xability/maidr#1094 for `Plot.link` and #1100 for `Plot.rule`.

Note row 4: lane `build` holds two intervals. `GanttData.points` is nested per lane precisely so one layer carries the whole schedule and up/down moves *between* lanes — the shape that chart needs.

## `geom_curve()` comes with it

Identical computed columns (`x, xend, y, yend`) — the curvature is a drawing instruction that never becomes a position, the same conclusion #1094 reached for `Plot.link`'s `curve` option. One processor reads both.

## Two things to decide, both measurable

**A segment whose ends share nothing is not a span.** An edge in a node-link diagram has no lane to sit in:

```
  x xend y yend group
1 0    2 0    2    -1        # both ends differ on both axes
2 1    3 1    3    -1
```

xability/maidr#1100 settled how to ask this — of the whole mark rather than per row, because reading three spans out of four segments announces a gantt quietly missing a quarter of its chart.

**An empty lane is real, and ggplot2 can express it.** A factor level nothing was drawn for is dropped from the panel by default, but `scale_y_discrete(drop = FALSE)` keeps it:

```
default      labels: design, build, test
drop = FALSE labels: design, build, test, idle
```

"Nothing is booked on this resource" is a statement about a schedule, and `GanttData.lanes` exists to carry exactly that row.

Both orientations work symmetrically — lanes on x with the spans running up y gives the mirror image of the table above.

## Highlighting is the open question

The layer draws as a single `segments` grob (`GRID.segments.1` under the panel), one grob for every interval, the way `geom_hex` draws one polygon grob for every bin. Whether that exports as one `<line>` per interval — which would give per-interval selectors, unlike #193's raster — needs measuring against the real output rather than assuming, and it should be settled in the implementation rather than discovered afterwards.

## Not this issue

`geom_rect()` is the same interval with both bounds given on both axes (`xmin,xmax,ymin,ymax`, and `group = -1` rather than a lane index). Coherent as a gantt, but its lane is a numeric band rather than a discrete level, which is a different naming question — its own issue.

`geom_contour` / `geom_density_2d`, `geom_dotplot` and `geom_quantile` remain from #192's sweep. `geom_rug` and `geom_spoke` stay declined on the reasoning recorded there.

jooyoungseo:
Closes #194 — the first of the readings #192 recorded alongside the raster.

`geom_segment()` with the two ends of a span on one axis and a lane on the other is how ggplot2 draws a schedule, a range plot and a high–low chart. It was read as nothing: `save_html` emitted the fallback page.

```r
ggplot(tasks) + geom_segment(aes(x = start, xend = end, y = task, yend = task))
```

## The numbers are already there

`ggplot_build` computes both ends and the lane exactly — nothing inverted from a pixel, nothing reconstructed:

```
   x xend y yend        →   [[{"x":"design","start":0, "end":3 }],
1  0    3 1    1             [{"x":"build", "start":3, "end":8 },
2  3    8 2    2              {"x":"build", "start":12,"end":15}],
3  8   11 3    3             [{"x":"test",  "start":8, "end":11}]]
4 12   15 2    2
```

Note lane `build`: booked twice, and both bookings sit in one row a reader moves along. `GanttData.points` is nested per lane for exactly that.

## Four things decided, all measured

**Lane names come off the panel's view of the scale**, not the source column. The built data records a discrete lane as the position ggplot2 gave it (1, 2, 3) and `panel_params[[i]]$y$limits` holds the levels in the same order, so the two line up by index. A continuous lane axis has no names, and the position itself is announced — `GanttPoint$x` takes a number or a string for that case, and `lanes` is omitted rather than filled with invented names.

**An empty lane is real, and ggplot2 can express it.** A level nothing was drawn for is dropped by default and kept by `scale_y_discrete(drop = FALSE)`:

```
default       lanes: design, build, test
drop = FALSE  lanes: design, build, test, idle   ← points: [..., ..., ..., []]
```

**A segment whose ends share nothing is not a span** — it is an edge in a node-link diagram. Asked of the whole layer rather than of each row, the rule xability/maidr#1100 settled for the same reading: one call can hold spans and edges together, and reading three spans out of four segments announces a gantt quietly missing a quarter of its chart. Refused layers return `"unknown"`, which is what they return today, so a declined chart keeps exactly the static-image fallback it already had.

**Both orientations, with the swap left to the frontend.** Lanes on y emit `orientation: "horz"` (bars running left to right, the ordinary gantt) and lanes on x `"vert"`; `Gantt`'s `text` getter swaps the two axis labels itself, so the labels stay the plot's own.

## Highlighting works here, unlike #193's raster

#194 left this open, and it is settled by measurement rather than assumption. `GeomSegment` draws one `segmentsGrob`, and gridSVG exports it as **one element per interval**:

```
id="GRID.segments.1.1.1"  id="GRID.segments.1.1.2"
id="GRID.segments.1.1.3"  id="GRID.segments.1.1.4"
```

So the layer emits one selector per interval, flat and in lane-major order — the frontend slices it by lane length and withdraws highlighting outright unless the count matches. The test asserts not only that four selectors are emitted in the order the lanes hold them, but that **each names an element that is really in the SVG**. A selector matching nothing is the highlight-only blind spot xability/maidr#814 describes, which no assertion about the payload alone can see.

Two details cost a rewrite each and are worth recording:

- `LayerProcessor$find_layer_grob_tree()` cannot serve. It matches a grob whose name begins with the geom's prefix, and ggplot2 leaves a segment layer with grid's automatic name — `GRID.segments.38`, no `geom_segment.` anywhere. The first version resolved NULL, emitted no selectors, and announced everything: a chart that reads perfectly and lights up nothing. `find_segments_name()` applies the same disambiguation rule keyed on the grob's **class**, so the nth segment layer takes the nth segments grob.
- That number is grid's global counter and is **not stable between sessions** — measured as `1` in one process and `38` in another. It is read off the gtable being exported rather than reconstructed.

## `geom_curve()` is deliberately excluded

It computes the same four columns and would read the same way. `GeomCurve` draws a `curve` grob, and `gridSVG::grid.export()` cannot export one:

```
Error in svgStyleAttributes(style, svgdev) :
  All SVG style attribute values must have length 1
```

An unsupported chart takes the static-image path and never meets that. Claiming the layer sends the same chart down the gridSVG path, and `save_html()` **raises** — I hit exactly that in the first working version, turning a chart that rendered into one that errored. Filed as #195 with the measurement, and the exclusion is pinned by a test rather than left to look like an oversight.

## Verification

Full `testthat::test_dir()`: no failures. `test-gantt-segment.R` is 29 assertions. Every guard falsified by applying each mutation alone:

| mutation | result |
|---|---|
| the adapter branch removed | 16 of 29 failed |
| the span test always says yes | 2 failed |
| the degenerate guard (level on **both** axes) dropped | 1 failed |
| intervals not sorted within a lane | 2 failed |
| span ends taken as written, not normalised | 1 failed |
| lane names not read from the scale | 6 failed |
| only drawn positions become lanes | 1 failed |
| orientation always `"horz"` | 1 failed |
| selectors numbered 1..n instead of by built row | 2 failed |
| the selector lookup reverts to `find_layer_grob_tree` | 3 failed |

**Two guards were dead on the first pass** and are the reason two more fixtures exist. The degenerate case — every segment reduced to a point, a scatter drawn with segments — had nothing exercising it. And "intervals ascend within a lane" was indistinguishable from "leave them alone", because the fixture's two `build` bookings were already in ascending order; a second frame writes the later booking first, which also strengthens the selector test (emission order becomes 1, 4, 2, 3 rather than 1, 2, 4, 3).

**The repo's own guard caught a defect in this PR.** `test-roxygen-orphans.R` failed on `ggplot2_adapter.R:445`: I had anchored the new helper on `ribbon_is_area = function(...)` and inserted it *between* that function's doc block and its body, stranding the block. Same mistake I have made before, and this repo has a test that finds it. Relocated after `ribbon_is_area`'s body; the check is green.

## About the diff size

34 `man/*.Rd` files change by one line each. `resolve_panel_index()` moves from `Ggplot2LineLayerProcessor` to `LayerProcessor`, which is where its siblings `get_layer_built_data()` and `find_layer_grob_tree()` already live, and every inheritor's generated page gains the inherited-method link. Duplicating a twelve-line panel resolver to avoid generated churn would let two readings disagree about which panel they are describing, which is the one thing a panel resolver exists to prevent.

## Remaining from #192

`geom_rect()` is the same interval with both bounds given on both axes, and its lane is a numeric band rather than a discrete level — a different naming question, its own issue. `geom_contour` / `geom_density_2d`, `geom_dotplot` and `geom_quantile` are still unread; `geom_rug` and `geom_spoke` stay declined on the reasoning recorded there.

---


jooyoungseo:
All four points were right. Three are fixed in ca2eb6d; the fourth turned out to be safer than feared, and is now pinned.

**1 and 2 — `NEWS.md` and `stats`.** Both done. `NAMESPACE` already carried `importFrom(stats, setNames)`, so the `DESCRIPTION` declaration was the half that was missing rather than a new dependency.

**3 — no test for two `geom_segment()` layers.** This is the one that should not have shipped as it was: I argued the position counting in prose and left nothing exercising it, which is exactly the shape this PR's mutation table exists to catch. `a second segment layer highlights its own intervals` now builds the two-layer chart, asserts the two layers resolve to *different* grobs, and checks both layers' selectors name elements that are really in the SVG. Measured, the grobs are `GRID.segments.1` and `GRID.segments.2` in layer order.

**4 — a zero-row layer shifting the indices.** Measured rather than accepted, and the news is good: ggplot2 emits **no segments grob at all** for a layer with no rows —

```
two segment layers          → GRID.segments.1, GRID.segments.2
first layer draws zero rows → GRID.segments.39          (2 layers, 1 grob)
```

— so the count runs *past the end* of what was drawn and `find_segments_name()` returns `NULL`. The later layer loses its highlight; it is never given the earlier layer's element. That is the direction the failure has to fall, so it is now pinned by `a layer that draws nothing costs the next one its highlight, not its accuracy`.

That test asks the lookup directly rather than rendering, and the reason is worth recording: a `geom_segment()` layer with no rows classifies as `"unknown"` — `segment_lane_axis()` has nothing to answer about — and an unknown layer drops the whole plot to a static image. Rendering the chart would exercise that pre-existing behaviour instead of the thing under test.

Full suite still clean; `test-gantt-segment.R` is now 41 assertions.

---

Issue #197, opened 2026-08-20
`geom_rect()` draws gantt charts too, but the rule for telling one from an annotation needs deciding


jooyoungseo:
Recorded as "not this issue" in #194, now that #196 has landed `geom_segment()`. The data is there; what is not settled is when a rectangle layer *is* a schedule.

## The numbers are exact

ggplot2 3.4.4. `geom_rect()` computes all four bounds:

```
  xmin xmax ymin ymax group
1    0    3  0.6  1.4    -1        # design, 0 to 3
2    3    8  1.6  2.4    -1        # build,  3 to 8
3    8   11  2.6  3.4    -1        # test,   8 to 11
4   12   15  1.6  2.4    -1        # build, 12 to 15
```

Each row is an interval in a band, which is `GanttPoint`'s `{x, start, end}` with nothing inferred — and the second `build` row is the nested-lane case `GanttData.points` exists for.

Highlighting is easier here than it was for the segment. ggplot2 names this layer's grob after the geom — `geom_rect.rect.2` — so `LayerProcessor$find_layer_grob_tree()` resolves it directly, without the class-based lookup #196 needed because `GeomSegment` arrives as `GRID.segments.38`.

## Naming the lanes is a solved problem, with one trap

A rect's `ymin`/`ymax` are numbers, so the y scale is **continuous** even when the bands stand for categories. The name is the tick inside the band:

```
default                                          breaks: 1, 2, 3   labels: "1", "2", "3"
scale_y_continuous(breaks = 1:3, labels = tasks) breaks: 1, 2, 3   labels: design, build, test
```

The trap is the first row. Those labels are the axis writing its own coordinates out, and a lane called `"2"` is worse than a lane called by its position — it looks like a name. xability/py-maidr#533 hit exactly this with `broken_barh` and settled it: a lane is named by the single **explicit** tick inside it, and by its position otherwise. The same test applies here, using "is this label a rendering of its own number" rather than matplotlib's locator class.

## The part that needs deciding: what is *not* a gantt

`GeomRect` is the most general mark ggplot2 has, and three other things arrive as one:

- **`annotate("rect", …)`** — a highlighted region behind a chart. This is the case that matters: claiming it would announce a piece of decoration as a schedule, and it is far commoner than a rect-drawn gantt.
- **A waterfall chart**, where the lanes would be on x and the bars carry the value.
- **A heatmap drawn with `geom_rect()`** rather than `geom_tile()`.

[...]

PR #199, opened 2026-08-20
feat(ggplot2): read geom_contour and geom_density_2d as the fields they draw

### LINKED ISSUE #198: `geom_contour()` and `geom_density_2d()` compute their levels as numbers, and read as nothing
The next of the readings #192 recorded, after #196's `geom_segment()`. ggplot2 3.4.4.

## Silent today

```r
ggplot(grid, aes(x, y, z = z)) + geom_contour()      # → unknown
ggplot(points, aes(a, b)) + geom_density_2d()        # → unknown
```

MAIDR has carried the contour trace since xability/maidr#802, and xability/py-maidr#540 has just landed the same reading for matplotlib, so the design questions are already answered.

## ggplot2 hands over more than matplotlib does

`ggplot_build` computes the level **as a number, per row**, and has already split the disconnected curves:

```
  level  piece  group          rows
1   0.3      1  -1-001-001       77
2   0.3      2  -1-001-002       77
3   0.6      3  -1-002-001       51
4   0.6      4  -1-002-002       51
```

That is a field with two peaks and two thresholds. Each level crosses it twice, and `piece` separates the islands — so there is no compound path to split, unlike matplotlib, where both islands arrive in one `Path` with two `MOVETO`s and #540 had to break them apart. Group by `piece`, and each group is one curve with its level already on every row, which is exactly `ContourPoint`'s shape.

`geom_density_2d` computes the same columns with estimated levels (`0.02, 0.04, … 0.16`).

## The filled forms say themselves that they are a different chart

```
geom_contour_filled       level: (0.0, 0.1]  (0.1, 0.2]  …   plus level_low, level_high, level_mid
geom_density_2d_filled    level: (0.00, 0.02]  (0.02, 0.04]  …
```

`level` is a **factor of intervals** rather than a number, because a filled contour draws the bands *between* levels. #540 declined `contourf` and a filled bivariate `kdeplot` for exactly this, and here the built data states it outright — so the refusal can be read off the frame rather than inferred from the geom name.

## Highlighting is per curve

gridSVG exports the layer as one `<polyline>` per piece:

```
GRID.polyline.1.1.1   GRID.polyline.1.1.2   GRID.polyline.1.1.3   GRID.polyline.1.1.4
```

Four pieces, four elements — better than matplotlib, where #540 could only address a level's whole path and two islands of one level share an element.

## The one thing that needs care

The grob is `GRID.polyline.N` — grid's automatic name, no geom prefix — so a contour layer joins the population `layer_polyline_grobs()` returns and `polyline_layer_position()` indexes into. That helper counts `"line"` and `"step"` layers only, and its own comment says why that matters:

> Counting only `"line"` layers would index the wrong polyline for **both** layers of a plot that combines the two.

A contour layer renders an auto-named polyline just as they do, so it has to join that count, or a `geom_line() + geom_contour()` chart will highlight each layer with the other's curve. That is the same class of failure #196 hit with `GeomSegment` and the same one xability/maidr#814 describes: both charts read correctly and both outline the wrong thing.

## Suggested shape

One `Ggplot2ContourLayerProcessor`, classified from `GeomContour` and `GeomDensity2d` (the filled subclasses declined), grouping the built rows by `piece` and emitting one series per piece with `level` on every point — plus `polyline_layer_position()` taught about the new type.

Still remaining from #192: `geom_dotplot` and `geom_quantile`; `geom_rect` is #197.

jooyoungseo:
Closes #198 — the next of the readings #192 recorded, after #196's `geom_segment()`.

```r
ggplot(field, aes(x, y, z = z)) + geom_contour()   # was: unknown, a static image
ggplot(points, aes(a, b)) + geom_density_2d()      # was: unknown
```

## The level is a number, not a colour

`ggplot_build()` puts `level` on every row, so both halves of the reading come off the frame:

```
  level  piece  group          rows
1   0.3      1  -1-001-001       77
2   0.3      2  -1-001-002       77
3   0.6      3  -1-002-001       51
4   0.6      4  -1-002-002       51
```

That is what separates this chart here from the same chart elsewhere: xability/maidr#1084 left Observable Plot's `contour` unread because there the magnitude survives only in a continuous fill, and distinct levels render as the same pixel.

The curves are checked against the **field**, not against ggplot2's own output — a gaussian bump's 0.5 contour is the circle of radius `√(-ln 0.5)`, and every emitted point sits on it to within 0.05.

## ggplot2 does the hard half that matplotlib does not

A field with two peaks crosses a level twice, and `piece` already separates the islands. So there is nothing to split apart here — unlike xability/py-maidr#540, where both islands come back in one `Path` with two `MOVETO`s, and reading them together would announce a straight line running across the saddle between the peaks. Group by `piece` and each group is one curve with its level on every row, which is `ContourPoint`'s shape exactly.

## The filled forms say themselves that they are a different chart

```
geom_contour_filled       level: (0.0, 0.1]  (0.1, 0.2]  …
geom_density_2d_filled    level: (0.00, 0.02]  (0.02, 0.04]  …
```

`level` is a **factor of intervals**, because a filled contour draws the bands between levels — so an outline runs along two different level curves and announcing it as one would be right for half its points. Both the classification and the reader check: naming the two geoms catches the ordinary case, and `contour_curves()` refusing a non-numeric `level` means a stat that ever produced banded levels under a line geom is declined rather than announced with a factor where a number goes.

## The interaction that needed the care

`GeomContour` draws a bare `polylineGrob`, named `GRID.polyline.N` with no geom prefix — so a contour layer lands in exactly the candidate list `layer_polyline_grobs()` returns and `polyline_layer_position()` indexes into for a line or a step. That helper's own comment already warned about this shape:

> Counting only `"line"` layers would index the wrong polyline for **both** layers of a plot that combines the two.

So it now counts `"contour"` as well. Without that, a `geom_line() + geom_contour()` chart hands each layer the other's elements: both read correctly, both outline the wrong thing — the blind spot xability/maidr#814 names, which no assertion about the payload can see. The test builds that chart **in both draw orders** and checks the two layers resolve to different grobs and that every selector names an element really in the SVG.

`find_layer_polyline_grob()`, `layer_polyline_grobs()` and `other_geom_grob_prefixes()` move from the line processor to `LayerProcessor` for the same reason: three geoms draw through that grob, and one answer to "which of them is mine" has to serve all three. `line_layer_position()` stays where it is — tests call it by name — and now delegates through the shared lookup.

## Highlighting is per curve

gridSVG exports the layer as one `<polyline>` per piece — `GRID.polyline.1.1.1` … `.4` — so each curve gets its own element, which is better than #540 managed for matplotlib, where two islands of one level share the level's path.

## Verification

Full `testthat::test_dir()`: no failures. `test-contour-level.R` is 41 assertions. Every guard falsified by applying each mutation alone:

| mutation | result |
|---|---|
| the adapter branch removed | 8 of 11 tests failed |
| the filled forms claimed too | 1 failed |
| the numeric-`level` guard in the reader dropped | 1 failed |
| curves grouped by level rather than by `piece` | 3 failed |
| `polyline_layer_position()` forgets contour | 2 failed |

The last row is the one worth reading: it reddens both the two-layer chart and the position helper's own case, which is what makes the line/contour interaction guarded rather than hoped for.

**A test bug surfaced while writing it, and it is worth recording.** The two-layer case first reported the *line's* selector as naming nothing. It does not — this package emits two selector forms, `*[id='GRID.polyline.1.1.1']` from the newer processors and `#GRID\.polyline\.1\.1\.1` from the line processor, and my id extraction knew only the first. Both are valid CSS and both resolve; the test now handles either, in a helper that says why.

## Remaining from #192

`geom_dotplot` and `geom_quantile`. `geom_rect` is #197, which needs the rule for telling a schedule from an `annotate("rect")` decided before an adapter can emit one.

---


jooyoungseo:
Both review points applied in c4e18b6, and the third one turned out to be bigger than a missing name.

**1. The stale comment on `line_layer_position()`** — right, and worse than stale: naming two of the three counted types reads as an exhaustive list. Rewritten to name all three, with the reason each one qualifies. Measured rather than assumed:

```r
"draw_panel" %in% ls(GeomContour)   # FALSE
"draw_panel" %in% ls(GeomStep)      # TRUE
```

so `GeomContour` inherits `GeomPath`'s bare `polylineGrob` directly, while `GeomStep` stairsteps and then calls the same method. Either way it lands in `layer_polyline_grobs()`'s candidate list.

**2. The detached comment in the factory** — fixed; the segment comment sits directly above `"gantt"` again, and the contour case brought its own comment with it.

**3. The three extra methods on `LayerProcessor`** — you were right that six moved where the PR body claimed three, and the three extras were not deliberate: my block extraction spanned them. They are back on `Ggplot2LineLayerProcessor`. Nothing outside it calls them, and they hardcode `GRID.polyline.<id>` selectors that only the line processor emits — the contour processor builds `*[id='...']` selectors of its own. The `man/` churn in this commit is the mechanical consequence: an inherited-method list appears on every subclass's page, so moving three methods off the base class rewrites all 33.

**4. `get_available_processors()`** — filed as #200 rather than fixed here, because adding the three missing names would have looked like a fix and changed nothing:

```r
f <- maidr:::Ggplot2ProcessorFactory$new()
f$is_processor_available("Ggplot2BarLayerProcessor")   # FALSE
length(f$get_available_processors())                   # 0
```

`is_processor_available()` tests `exists(name, mode = "function")`, and an R6 generator is not a function, so the predicate rejects every entry — including the ones already in the list. Both factories return `character(0)` unconditionally and always have. Nothing in `R/` calls either method; the only callers are two tests that assert the return value is a character vector, which `character(0)` satisfies. That is why the list drifted three times without anything noticing.

#200 lays out the two ways out — fix the predicate and add a drift guard against the `switch` (enumerable at runtime, so no second list), or delete both methods and their tests. I lean towards deleting, but a registry no dispatch path consults may still be intended as reflection API for a downstream, so I left the call rather than making it inside a contour PR.

Affected suites green: 479 passing across the line-processor, contour, factory, grob-utils and roxygen-orphan files. `man/` regenerated with the pinned roxygen2 8.1.0 and the default loader, matching what `docs-drift.yaml` runs.

---

Issue #202, opened 2026-08-20
geom_function renders as a static image, though it draws a plain sampled curve


jooyoungseo:
## What happens

```r
ggplot(data.frame(x = c(0, 6)), aes(x)) + geom_function(fun = sin)
```

```
geom_function    svg=FALSE  maidr-data=FALSE  img=TRUE
  warning: Plot contains unsupported elements. Rendering as static image
           instead of interactive MAIDR plot.
```

Same failure shape as #201: not a mis-typed layer but no layer, `should_fallback()` fires and the whole figure goes out as an `<img>`.

## Why the classifier misses it

`GeomFunction` **is** a `GeomPath`:

```r
class(GeomFunction)   # "GeomFunction" "GeomPath" "Geom" "ggproto" "gg"
class(StatFunction)   # "StatFunction" "Stat"     "ggproto" "gg"
```

but the branch matches `class(geom)[1]` against a fixed list:

```r
if (geom_class %in% c("GeomLine", "GeomPath", "GeomMA")) {
  return("line")
}
```

so the subclass falls through. Unlike `GeomRaster` in #193 and `GeomDotplot` in #201, this one really is a relative of something already handled — it is the first-element match that misses it, not the class tree.

## The reading is a plain series

`ggplot_build()` gives exactly what a line layer reads — one x/y pair per sample, no bands, no groups:

```
  x          y group
1 0  0.0000000    -1
2 1  0.8414710    -1
3 2  0.9092974    -1
4 3  0.1411200    -1
...
```

and the grob is a bare auto-named polyline, which is what the line and contour processors already address:

```
gTree | panel-1.gTree.16
  ...
  polyline | GRID.polyline.1
```

## `smooth`, not `line`

A `geom_function` curve is **sampled from a function** — `n` evenly spaced points chosen by the renderer, not observations. That is the distinction `smooth` already carries in this package: `GeomSmooth` and `StatDensity` both land there, and py-maidr types a KDE curve `smooth` for the same reason (xability/py-maidr#446). Announcing it as `line` would offer the sample count as though it were data; `n = 101` is a drawing parameter.

## Scope

Both halves are small and both are needed:

- classify `GeomFunction` (or `StatFunction`) as `"smooth"`;
[...]

Issue #200, opened 2026-08-20
get_available_processors() reports nothing available, in both factories


jooyoungseo:
## What happens

`Ggplot2ProcessorFactory$get_available_processors()` and `BaseRProcessorFactory$get_available_processors()` both return `character(0)`, always, for every processor the package ships.

```r
pkgload::load_all(".")
f <- maidr:::Ggplot2ProcessorFactory$new()
f$is_processor_available("Ggplot2BarLayerProcessor")   # FALSE
length(f$get_available_processors())                   # 0
length(maidr:::BaseRProcessorFactory$new()$get_available_processors())  # 0
```

## Why

`is_processor_available()` asks:

```r
exists(processor_class_name, mode = "function")
```

The processors are R6 generators, and an R6 generator is not a function:

```r
class(maidr:::Ggplot2BarLayerProcessor)   # "R6ClassGenerator"
is.function(maidr:::Ggplot2BarLayerProcessor)  # FALSE
```

so `mode = "function"` never matches and the `sapply()` in `get_available_processors()` filters everything out.

## How it surfaced

Review of #199 noticed that `get_available_processors()`'s hardcoded list is missing `Ggplot2ContourLayerProcessor`, and that `Ggplot2GanttLayerProcessor` and `Ggplot2HexbinLayerProcessor` were already missing before that. Adding the three names would have looked like a fix and changed nothing, because the list is filtered by a predicate that rejects every entry.

The list has drifted freely because nothing consults it. Across `R/` there are no callers; the only callers anywhere are two tests, one per factory, and each asserts only that the return value *is* a character vector — which `character(0)` satisfies:

```r
# tests/testthat/test-factories.R:254
processors <- factory$get_available_processors()
```

## Two ways out, and the choice is a design one

**Fix it.** `is_processor_available()` becomes `exists(name)` plus an `inherits(get(name), "R6ClassGenerator")` check, the hardcoded vector gains the three missing names, and a test asserts the vector agrees with what `create_processor()`'s `switch` actually dispatches to. The switch is enumerable at runtime without parsing source files, which matters because `R/` is not present in an installed package:

```r
src <- paste(deparse(body(Ggplot2ProcessorFactory$public_methods$create_processor)), collapse = "\n")
regmatches(src, gregexpr("Ggplot2[A-Za-z0-9]*Processor(?=\\$new)", src, perl = TRUE))[[1]]
```

[...]

PR #203, opened 2026-08-20
feat(ggplot2): read geom_dotplot as the histogram it draws

### LINKED ISSUE #201: geom_dotplot renders as a static image, though its built data is already a histogram
## What happens

```r
ggplot(data.frame(v = c(1,1,1,2,2,3,4,4,4,4)), aes(v)) + geom_dotplot(binwidth = 1)
```

```
geom_dotplot     svg=FALSE  maidr-data=FALSE  img=TRUE
  warning: Plot contains unsupported elements. Rendering as static image
           instead of interactive MAIDR plot.

geom_histogram   svg=TRUE   maidr-data=TRUE   img=FALSE      # same data
```

Not a mis-typed layer — no layer at all. `should_fallback()` fires and the whole chart is emitted as an ``, with no SVG and no schema. A blind reader gets an image with alt text saying it is an image. The same distribution drawn as `geom_histogram()` is fully navigable.

## Why the classifier misses it

`GeomDotplot` is a direct subclass of `Geom`, not of anything already handled:

```r
class(GeomDotplot)   # "GeomDotplot" "Geom" "ggproto" "gg"
class(StatBindot)    # "StatBindot"  "Stat" "ggproto" "gg"
```

So the class-name match finds nothing, exactly as it did for `GeomRaster` in #193 — a sibling rather than a subclass of the geom that already reads.

## The reading is already computed

A Wilkinson dot plot *is* a histogram: it bins the values and stacks one dot per observation, so a stack's height is the bin's count. `ggplot_build()` says so directly — one row per observation, each carrying its bin centre, the bin's count, and the bin width:

```
   y x binwidth count ncount countidx stackpos
1  0 1        1     3   0.75        1      0.5
2  0 1        1     3   0.75        2      1.5
3  0 1        1     3   0.75        3      2.5
4  0 2        1     2   0.50        1      0.5
5  0 2        1     2   0.50        2      1.5
6  0 3        1     1   0.25        1      0.5
7  0 4        1     4   1.00        1      0.5
...
```

Four distinct `x` values, counts 3/2/1/4 — which is the histogram, no reconstruction needed. Deduplicating on `x` gives bins of `[x - binwidth/2, x + binwidth/2)` carrying `count`, which is the shape `Ggplot2HistogramLayerProcessor` already emits.

Reading it as a `point` layer instead would announce ten observations all at `y = 0`, which is the layout, not the chart.

## Highlighting needs a decision

The layer draws **one** grob for the whole chart:

```
gTree | geom_dotplot.gTree.4
  dotstackGrob | geom_dotplot.dotstackGrob.2
```

so per-bin highlighting means naming a *run* of dots per bin rather than one element — the dots of a stack are consecutive, and their counts are known, so a positional range is available (`R/boxplot.R`'s outlier selectors already use `:nth-child` ranges for the same reason). Whether that resolves is worth measuring before promising: if it does not, the layer should ship announcing its bins and highlighting nothing, and say so, the way the raster reading in #193 does.

## Related

`geom_function` falls back to a static image the same way, measured in the same sweep:

```
geom_function    svg=FALSE  maidr-data=FALSE  img=TRUE
```

That is a different chart (a computed curve, so a `line` or `smooth` reading) and is filed separately.

Measured on ggplot2 4.0.1.

jooyoungseo:
Closes #201.

## The defect

A `geom_dotplot()` was not a mis-typed layer — it was **no layer at all**. `should_fallback()` fired on the unclassified geom and the whole figure went out as an ``:

```
geom_dotplot     svg=FALSE  maidr-data=FALSE  img=TRUE
geom_histogram   svg=TRUE   maidr-data=TRUE   img=FALSE
```

on the same data. A blind reader got an image whose alt text says it is an image.

`GeomDotplot` is a direct `Geom` subclass rather than a relative of anything already handled, so the class-name match found nothing — the same shape of miss `GeomRaster` was in #193.

## The reading was already computed

A Wilkinson dot plot *is* a histogram: it bins the values and stacks one dot per observation, so a stack's height is the bin's count. `ggplot_build()` says so directly — one row per observation carrying its bin centre, the bin width, and the bin's count:

```
   y x binwidth count countidx
1  0 1        1     3        1
2  0 1        1     3        2
3  0 1        1     3        3
4  0 2        1     2        1
```

Collapsing on the centre gives four bins of 3, 2, 1 and 4. Asserted against `geom_histogram()` on the same data rather than against literals, because what was wrong was the disagreement between two spellings of one chart.

## Bin bounds are computed, not read

The built data's own `xmin`/`xmax` name the bin in only **one** of the two orientations. Measured, `binaxis = "y"` puts the panel's whole range in `ymin`/`ymax` and the dot's own width in `xmin`/`xmax`:

```
   y binwidth count x countidx ymin ymax xmin xmax
1  1        1     3 1        1  0.5  4.5 0.55 1.45
```

so neither pair is the bin there, and the bounds come from the centre and the width instead. Both orientations read, and the horizontal one reports `orientation: "horz"` rather than quietly swapping its fields.

## The count axis is named, not read back

ggplot2 labels a dot plot's other axis "count" while drawing values on it that its own documentation calls meaningless. Reading the label back gives "count" for a default plot too — so only a caller who renamed it separates the two readings, and that is the test:

```r
+ labs(y = "proportion of stack")   # axes$y$label is still "count"
```

Without it, a real count would be announced under the caller's name for the fiction.

## Highlighting is not offered, and that is pinned

The layer draws one `dotstackGrob`, which gridSVG exports as one `<circle>` per **observation**:

```
id="geom_dotplot.dotstackGrob.2.1.1"
id="geom_dotplot.dotstackGrob.2.1.2"
…  (ten circles for ten observations)
```

A three-dot bin therefore has three elements and no element of its own, while the frontend's bar traces resolve exactly one element per announced value. A representative dot would be worse than none — the reader would be shown one dot of a stack as though it were the bin.

So the bins are announced, sonified and brailled and nothing lights up. That is the highlight-only blind spot xability/maidr#814 names, it is strictly better than the picture this chart was before, and the test fails the day a group per stack appears — which is the reminder to give it selectors and delete the pin. The same shape as the `geom_raster` limit in #193.

## Falsification

| mutation | result |
|---|---|
| bin bounds from a fixed dot-sized box | red |
| ignore `binaxis` (always vertical) | red |
| don't classify `GeomDotplot` | red |
| read the count axis label from the plot | red |
| count rows instead of reading the `count` column | **green** |

The last one survives and the test says so in its own comment rather than implying otherwise: `stat_bindot` expands a bin to one row per weighted unit and refuses a fractional weight outright (`` `weight` must be nonnegative integers ``), so counting rows gives the same answer everywhere ggplot2 will build. The `count` column is read anyway — the stat's own answer rather than a count of the rows that happen to represent it — and what the test fixes in place is that a weighted plot announces 4 and 5 rather than 2 and 1.

## Not included

`get_available_processors()` gains nothing here, because #200 measured that both factories' copies return `character(0)` unconditionally — `is_processor_available()` tests `exists(name, mode = "function")` and an R6 generator is not a function. Adding the name would have looked like a fix and changed nothing.

## Checks

Full suite: `[ FAIL 0 | WARN 8 | SKIP 99 | PASS 5929 ]`. `man/` regenerated with the pinned roxygen2 8.1.0 and the default loader, matching what `docs-drift.yaml` runs.

Filed alongside in the same sweep: #202 (`geom_function` falls back to a static image the same way, a different chart with a `smooth` reading).

---


jooyoungseo:
Finding 1 was a real undercount, not a scope cut — thank you. Fixed in 2fdcf4c.

Measured it before changing anything, and it reproduces exactly as described. Five observations at one bin, split three and two by `fill`:

```
x count group countidx stackpos
1     3     1        1      0.5
1     3     1        2      1.5
1     3     1        3      2.5
1     2     2        1      0.5
1     2     2        2      1.5
```

`dotplot_bins()` took `counts[rows[1]]` and announced three, dropping two observations with nothing saying so.

**Summed, not maxed, and the choice needed making rather than defaulting** — because the two candidate numbers disagree and neither is free of objection.

The sum is the number of observations the bin holds, which is what a `hist` layer's value means. It is also what `stackgroups = TRUE` literally draws — measured, that spelling continues the stack:

```
count x group countidx stackpos
    3 1     1        1      0.5
    3 1     1        2      1.5
    3 1     1        3      2.5
    2 1     2        4      3.5
    2 1     2        5      4.5
```

so the column really is five high there.

The default does not draw that. Both groups start at `stackpos` 0.5 and **overlap**, so the pile is three high and two observations sit hidden behind it. Announcing three would report the height of a pile that conceals data. Five is announced because the overlap is the rendering caveat ggplot2 documents rather than a fact about the distribution — and the reasoning is written into `dotplot_bins()` and the new test rather than left implicit, since a later reader will reasonably wonder.

Both alternatives are falsified: taking the first group's count (the reviewed defect) and taking the maximum each turn the new test red, 2 failures apiece.

On finding 2 — you're right that it reads as an omission rather than a decision. The class doc carries the full rationale, but the file-shape signal points the other way, and a maintainer scanning method lists sees the absence before the doc. I'd rather not add a no-op override whose only content is a comment, so I've left `process()` as it is; if you'd prefer the sibling shape for consistency I'll add `generate_selectors()` returning `list()` with the `dotstackGrob` reasoning inline.

Suite still green at `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 29 ]` for this file, and `man/` regenerated for the changed `dotplot_bins()` block.

---

PR #206, opened 2026-08-21
fix(ggplot2): give each curve layer the polyline it actually drew

### LINKED ISSUE #204: A smooth drawn before a line highlights the line's curve, and both layers share one selector
## What happens

Two layers, same data, only the order changed:

```r
ggplot(d, aes(x, y)) + geom_line() + geom_smooth(se = FALSE, method = "lm", formula = y ~ x)
```

```
line    data n=12  selector GRID.polyline.11.1.1   svg polyline points=12   ✓
smooth  data n=80  selector GRID.polyline.12.1.1   svg polyline points=80   ✓
```

```r
ggplot(d, aes(x, y)) + geom_smooth(se = FALSE, method = "lm", formula = y ~ x) + geom_line()
```

```
smooth  data n=80  selector GRID.polyline.70.1.1   svg polyline points=12   ✗
line    data n=12  selector GRID.polyline.70.1.1   svg polyline points=12   ✗
```

Both layers get **the same selector**, and it is the line's. The smooth layer announces its 80 fitted points while outlining the 12 observations; switching layers changes what is said and leaves the same curve lit.

This is the highlight-only failure xability/maidr#814 names in its purest form. Audio, braille and text are all correct for both layers. Nothing in the payload is wrong. Only the outline is, and no assertion about announcements can see it.

## Why

`Ggplot2SmoothLayerProcessor$generate_selectors()` collects every polyline in the panel and takes the largest counter:

```r
# Fitted line is the LAST polyline (confidence interval rendered first)
# ggplot2 renders confidence interval first, then the fitted line
target_id <- max(numeric_ids)
```

The reasoning is about **one smooth layer's own** internals — its band is drawn before its line — but the search is over the whole panel, so any other polyline-drawing layer joins the competition. `GRID.polyline.N` is grid's session-wide counter, so the largest is simply whatever was drawn last, which is the *other* layer whenever the smooth is not last.

The line processor mis-resolves in the same plot for a different reason: `polyline_layer_position()` counts `"line"`, `"step"` and `"contour"` layers, so with a `smooth` ahead of it the position and the candidate list disagree.

## What the fix has to distinguish

Layers fall into two families, and the grob tree says which:

- **Geom-named tree.** `geom_smooth` draws `geom_smooth.gTree.15` containing `GRID.polyline.13`. `layer_polyline_grobs()` deliberately skips these, so `find_layer_polyline_grob()` cannot find them — the right entry point is `find_layer_grob_tree()`, which is what the hexbin processor uses to disambiguate two layers sharing a geom, then the polyline inside that tree.
- **Bare polyline.** `geom_line`, `geom_step`, `geom_contour` — and `geom_function`, which is what turned this up — draw straight into the panel. Measured on a four-layer plot:

```
gTree | panel-1.gTree.32
  points   | geom_point.points.12
  gTree    | geom_smooth.gTree.15
    polyline | GRID.polyline.13
  polyline | GRID.polyline.16     ← geom_function
  polyline | GRID.polyline.17     ← geom_line
```

So the fix is ownership rather than a better heuristic: ask which grob *this layer* drew, the way the hexbin and contour processors already do, instead of guessing from the counter.

## Scope

`R/ggplot2_smooth_layer_processor.R:372-392`, and `polyline_layer_position()` in `R/grob_utils.R` for the other half. Worth a test per order for each pairing, since one order passes today and hides the other.

Found while measuring #202 — `geom_function` reads as `smooth`, so it lands in the same path and would inherit this. It is filed separately because it is not caused by that work: the `geom_smooth` + `geom_line` case above needs no new geom at all.

Measured on ggplot2 4.0.1.

jooyoungseo:
Closes #204.

## What was wrong

Two layers, two independent mistakes, and they cancelled in one draw order and compounded in the other. Measured on this branch before the fix, reading the emitted selector back out of the rendered document:

| draw order | curve layer outlines | line layer outlines |
|---|---|---|
| `geom_function()` then `geom_line()` | **the line's polyline** ✗ | **the function's polyline** ✗ |
| `geom_line()` then `geom_function()` | its own ✓ | its own ✓ |
| `geom_smooth()` then `geom_line()` | **the line's polyline** ✗ | its own ✓ |
| `geom_line()` then `geom_smooth()` | its own ✓ | its own ✓ |

Audio, text and braille are correct in every row. Only the outline is wrong, and no assertion about the announcement can see it — the highlight-only shape xability/maidr#814 names.

**The curve half.** `generate_selectors()` took the largest `GRID.polyline.N` in the panel. The reasoning is sound but local: within one smooth layer, ggplot2 draws the confidence band before the fitted line, so the later grob is the line. Applied panel-wide, the largest counter is simply whatever was drawn last — so any other polyline-drawing layer wins it.

**The line half.** `polyline_layer_position()` counted `line`, `step` and `contour` layers. `geom_function()` is typed `smooth` and yet draws a **bare** polyline: `GeomFunction` inherits `GeomPath$draw_panel()`, which returns a `polylineGrob` with no geom-named tree around it, so it lands in the candidate list `layer_polyline_grobs()` hands the line layer. The population being counted was one layer smaller than the population being indexed.

Measured grob trees, which is what the two branches turn on:

```
panel-1.gTree                        panel-1.gTree
  GRID.polyline.1   ← geom_function    geom_smooth.gTree
  GRID.polyline.2   ← geom_line          geom_ribbon.gTree
                                            GRID.polygon      ← band fill
                                            GRID.polyline     ← band edges
                                          GRID.polyline       ← fitted line
                                       GRID.polyline          ← geom_line
```

## The fix

Ownership rather than a better heuristic — ask which grob *this layer* drew, the way the hexbin and contour processors already do.

`Ggplot2SmoothLayerProcessor$own_curve_grob()` resolves the layer, then takes one of two routes depending on which shape above it is in:

- **geom-named tree** (`geom_smooth`, `geom_density`): `find_layer_grob_tree()`, then the last polyline *inside that tree*. This is the original "band first, line second" rule, applied where it is actually true.
- **bare polyline** (`geom_function`): `find_layer_polyline_grob()`, which answers "which of these is mine" by draw-order position — the question it already answers for the line and contour processors.

`polyline_layer_position()` counts through a new `layer_draws_bare_polyline()`, so a `geom_function` joins the population it was always part of.

`find_layer_polyline_grob()` / `layer_polyline_grobs()` / `other_geom_grob_prefixes()` gained a `target` argument, defaulting to this layer, mirroring `find_layer_grob_tree()`'s — the smooth processor may describe a layer other than its own.

## Tests

`tests/testthat/test-curve-selector-ownership.R` is new, and the `#204` limit pinned in `test-function-curve.R` is rewritten from a pin into a guard, as its own comment promised.

Every pairing is asserted **in both orders**, because one order was already correct and hid the other. Each test resolves the selector against the rendered document and counts the points of the element it lands on — 25 for a curve sampled at `n = 25`, 12 for a line over 12 rows — rather than comparing selector strings, which would only show that two ids differ and not which is whose.

The band test needed geometry rather than counting. ggplot2 draws each band edge as its own sub-polyline sampled at the same `n` as the curve, so all three hold 25 points:

```
GRID.polyline.12.1.1  25 points   ← band edge
GRID.polyline.12.1.2  25 points   ← band edge, traced in reverse
GRID.polyline.15.1.1  25 points   ← fitted line
```

It asserts instead that the outlined curve runs strictly between the other two at every sample, ordering each by x first — the ribbon's outline is traced up one edge and back down the other, so comparing by index would line each sample up against the wrong one.

## Falsification

Each change reverted alone, with the mutation asserted to have applied:

| mutation | result |
|---|---|
| `polyline_layer_position()` ignores `GeomFunction` again | `test-function-curve.R` guard **red** |
| `generate_selectors()` back to the panel-wide `max()` | 2 ownership tests + the guard **red** (the "smooth after line" order stays green, which is the order-dependence the table above shows) |
| take the **first** polyline in the layer's tree, not the last | band test **red** |
| drop the bare-polyline branch from `own_curve_grob()` | `test-function-curve.R` **red** |

The band test did *not* discriminate the third mutation on its first draft — it counted points, and a band edge holds the same count as the curve. That is what the geometry assertion above replaced it with.

Suite: `[ FAIL 0 | WARN 8 | SKIP 99 | PASS 5959 ]`. `roxygen2::roxygenise()` run; `man/` committed. All touched lines within the 100-character `.lintr` cap.

---


jooyoungseo:
The `NEWS.md` question has an answer, and it is yes — added in 2449b20.

You were right to flag it as a question rather than assume either way, but the history settles it. `NEWS.md` is written per pull request for user-visible fixes, and the six most recent `fix(...)` merges each carry one:

```
2fc836c fix(base-r): say which way a horizontal grouped bar chart is drawn (#190)
859f27c fix(ggplot2): read a horizontal grouped bar chart's categories and values (#188)
41d359b fix(ggplot2): put a horizontal bar's measure in x and its category in y (#185)
f293c12 fix(ggplot2): keep a faceted point's position, and carry its name beside it (#180)
17d2047 fix(ggplot2): keep a chart interactive when a reference line is drawn on it (#177)
a18c240 fix(ggplot2): emit only the scatter samples ggplot2 drew (#171)
```

What misled the earlier review on #205 — and me — is that the recent `feat(...)` merges for `geom_dotplot` and `geom_function` did *not* add one. That is a gap in the feature side of the convention rather than evidence that fixes are batched at release. The entry is under the existing dev-version "Bug Fixes" heading, in the same voice as the ones around it, and it says plainly that nothing about the announcement changed in any of these cases — which is why no assertion about the reading could see the defect.

On the sandbox caveat: agreed that CI is the authority there. `R-CMD-check` is green on all four platforms, and the suite result in the description came from a local `testthat::test_dir()` run — `[ FAIL 0 | WARN 8 | SKIP 99 | PASS 5959 ]`.

---

PR #207, opened 2026-08-21
fix(factories): report the processors a factory actually dispatches to

### LINKED ISSUE #200: get_available_processors() reports nothing available, in both factories

jooyoungseo:
Closes #200.

## Reproduced

```
is_processor_available('Ggplot2BarLayerProcessor'): FALSE
length(get_available_processors()): 0
baseR length: 0
class(Ggplot2BarLayerProcessor): R6ClassGenerator
```

`is_processor_available()` asked `exists(name, mode = "function")`, and a processor is an R6 **generator** rather than a function, so the predicate rejected every entry and the `sapply()` filtered them all out. Both factories reported an empty registry for every processor they ship.

## The choice #200 left open

The issue offered "fix it" or "delete it" and said the call wasn't its to make. This is neither exactly, and I think it dissolves the tension rather than picking a side.

The objection to fixing was that *"a registry that no dispatch path consults is a second source of truth for the `switch` that will drift again"* — and it had drifted, three times over. The objection to deleting was that the methods might be API for a downstream that reflects over the package.

So: the methods stay, and there is no second list. `get_available_processors()` reads its answer off `create_processor()` at runtime:

```r
dispatched_processor_classes(Ggplot2ProcessorFactory, "Ggplot2")
```

Drift is not guarded against; it is structurally impossible, because the dispatch and the registry are now the same thing.

`deparse(body(...))` rather than reading `R/`, for the reason the issue itself gives: an installed package has no `R/` to read — the sources are gone by then and only the parsed function survives.

## After

```
is_processor_available('Ggplot2BarLayerProcessor'): TRUE
length(get_available_processors()): 20
baseR length: 14
```

The hand-written list held 16 ggplot2 names. The four it was missing — `Contour`, `Dotplot`, `Gantt`, `Hexbin` — are exactly the drift the issue predicted, plus one that landed after it was filed.

## Tests

The two tests the issue names asserted only that the result **is** a character vector, which `character(0)` satisfies. One of them said so in its own comment:

```r
# Should return character vector (may be empty if exists() doesn't find R6 classes)
```

That is the defect written down as though it were a tolerance, which is why nothing caught it. Each is now two tests: the registry is non-empty and holds a named processor, and it is the same **set** the dispatch names — asserted with `expect_setequal` against `dispatched_processor_classes()`, so it checks agreement rather than that someone remembered to update a second copy.

## Falsification

| mutation | result |
|---|---|
| the predicate goes back to `exists(name, mode = "function")` | 7 assertions red across both factories |
| the ggplot2 registry goes back to a hand-written two-name list | the ggplot2 drift guard red |

Suite: `[ FAIL 0 | WARN 8 | SKIP 99 | PASS 5969 ]`. `roxygen2::roxygenise()` run and `man/` committed. All touched lines within the 100-character `.lintr` cap.

---


jooyoungseo:
Thanks — nothing to change, but the `exists()` scoping observation was worth measuring rather than reasoning about, so I did.

Both directions of the concern:

```r
assign("Ggplot2BarLayerProcessor", 42, envir = globalenv())
f$is_processor_available("Ggplot2BarLayerProcessor")   # TRUE
length(f$get_available_processors())                   # 20

assign("Ggplot2NotAProcessor", 42, envir = globalenv())
f$is_processor_available("Ggplot2NotAProcessor")       # FALSE
```

A user object shadowing a real processor's name does not hide it: `exists()`/`get()` are called from inside a package function, so the search starts at the namespace and finds the generator before it ever reaches `globalenv()`. And a global-only non-generator is rejected, which is what the `inherits(..., "R6ClassGenerator")` half is for. So it is right in both directions rather than only stricter than what it replaced.

On the prose comments — leaving them, agreed, and for the reason you give: this is the third time this class of bug has been fixed in this package, and the third one is the one where writing down *why* the shape recurs is worth more than the lines cost.

On the unverified section: `Rscript` was available here. `[ FAIL 0 | WARN 8 | SKIP 99 | PASS 5969 ]`, `roxygen2::roxygenise()` run and `man/` committed, and CI's R CMD check across the four platforms is the authority either way.

---

PR #208, opened 2026-08-21
fix(iframe): hand focus back to the page when a reader tabs out of a chart

jooyoungseo:
## Description

A chart on a Quarto `revealjs` slide swallows every key and gives the reader no way back to the deck. Tab reaches the chart; from inside it, Space does not advance the slide, the arrows do not go back, and Shift + Tab leaves the document altogether for the browser's own UI.

Charts are embedded in an iframe and keyboard events do not cross a frame boundary, so while the reader is inside a chart the page around it hears nothing. That is right while they are reading — MAIDR binds the arrows, Space and Page Up/Down itself — and it leaves Shift + Tab as the way back.

Usually the browser handles that on its own. It cannot when nothing reachable precedes the frame in the page, and a Quarto deck is exactly that case: it renders no controls of its own, so a chart is the first thing on the page.

The chart cannot fix this from its side. r-maidr embeds it through a `data:` URL, whose opaque origin puts the host document's DOM out of its reach, so it asks by `postMessage` instead — and this is the half that answers.

## Changes Made

**`R/svg_utils.R`** — the once-per-document parent-side script each chart already installs now handles two messages rather than one, so `maidr_iframe_resize_script()` becomes `maidr_iframe_host_script()`:

- `maidr-iframe-height` resizes the frame to its content. Now checks that the height it was handed is a usable number before applying it — anything on the page can post a message, and an absent height was writing `"undefinedpx"`. The widget binding has always checked this; the two copies now agree.
- `maidr:frame-focus-escape`, new, moves focus out of the frame and into this document. Focus goes where the browser would have sent it when the page has a reachable tab stop before the frame, and to the element holding the frame — the slide, in a deck — when it does not.

Reachability is checked rather than assumed: reveal.js leaves the slides on either side of the current one rendered, so a chart on the previous slide is a tab stop in document order even though it is marked hidden. Sending the reader there would land them on an off-screen chart instead of the slide.

**`inst/htmlwidgets/maidr.js`** — the same listener, because the widget binding renders the iframe HTML through `innerHTML`, which never runs a script element appended to it. That is why this file has always carried its own copy of the height handling. Both files now carry a comment pointing at the other, since nothing checks that the two stay in step.

**`NEWS.md`**, **`man/`** (roxygen), and six cases in **`tests/testthat/test-svg-utils.R`** pinning the script's contract: both message types, the once-per-document guard, the prefer-the-previous-stop rule, the unreachable-stop exclusion, and the height check.

## Related

- [`xability/maidr#1111`](https://github.com/xability/maidr/pull/1111) is the engine half: it decides when the reader would be stranded and sends the message. Both are needed for the fix to work here. Merged as `c92e821`.
- The reveal.js behaviour that makes off-slide charts tabbable at all is upstream and already filed as [hakimel/reveal.js#1587](https://github.com/hakimel/reveal.js/issues/1587), open since 2016. It reproduces in a stock Quarto deck containing nothing but links, with no MAIDR involved, and is not addressed here.

## Verification

`testthat::test_local()` runs clean — no failures, no errors, only the pre-existing ggplot2 4.0 deprecation warnings. The generated script parses under `node --check`, and the `TABBABLE` selector it builds was compared character for character after the line-length rewrap.

Checked in Chromium (Playwright) against a real three-slide Quarto 1.10.18 `revealjs` deck with a ggplot2 chart on each slide, on every slide:

```
Tab        → chart          role=application, "This is a maidr plot of…"
→          → chart          "x is a, y is 3"
Shift+Tab  → section#slide-1-bar        ← back on the page
Space      → slide-2                    ← the deck responds again
```

Before this change every step after the first left the reader stuck in the chart. A plain Quarto HTML document was checked too, to confirm its tab order is untouched.

## Notes

Both this and the engine half are needed, and the engine ships in the maidr bundle — r-maidr loads 4.4.0 today, so this reaches users with the next engine release.

`main` has moved to `d781a2f` since this opened. The merge was run locally: `NEWS.md` and `R/svg_utils.R` both auto-merge, so this needs no rebase.


jooyoungseo:
Thanks — and thanks for reconstructing the escaped string by hand; that is the part of this change most likely to be wrong in a way nothing catches.

**Line length.** Fixed in `5b66b43`, split where the adjacent clauses already split. Worth doing even though no workflow runs `lintr`: it was the only line over the limit that this PR introduced, and the other three in the file predate it. I checked the built string rather than trusting the wrap — extracting `maidr_iframe_host_script()`, stripping the script tags and evaluating `TABBABLE` gives back the selector character for character, and the whole snippet still parses under `node --check`.

**The duplication.** Deliberate, and it predates this change: `inst/htmlwidgets/maidr.js` has always carried its own copy of the height listener for exactly the same reason — the widget binding sets `el.innerHTML = x.iframe_content`, and a script element assigned that way never executes, so nothing appended to the iframe HTML runs on that path. The new focus listener follows the copy that was already there rather than inventing a third arrangement. Factoring both out would mean generating the shared JS from R into a file the widget binding also loads, which is a bigger change than this fix wants to carry; if the two ever drift it will be worth doing for both listeners at once, not just this one.

**Re-verified after the wrap**: `testthat::test_local()` clean — no failures, no errors — and in Chromium against a real Quarto 1.10.18 `revealjs` deck with a ggplot2 chart per slide, Shift + Tab still lands on the current slide and Space still advances the deck.

---

jooyoungseo:
Three follow-ups pushed (`574a79e`, `b14f01c`, `da9ac88`).

**`docs-drift` was failing, and it was right.** I rewrote the roxygen block after generating the `.Rd` from it and never regenerated, so the committed page still described the earlier design — focus always landing on the frame's container, no mention of preferring the tab stop before it or of checking reachability. Regenerated with roxygen2 8.1.0, the version `DESCRIPTION` pins; the page now matches the source it is generated from.

**Finding 3 — closed after all.** You're right that it predates this PR, but sharing `frameOf()` between the two branches is what put the two copies side by side in one diff, and leaving them different invites the next reader to wonder which one is deliberate. The parent-side script now applies the same `typeof … !== "number" || … < 50` check the widget binding has always had. Worth having on its own terms: an absent height was writing `"undefinedpx"`, and any script on the page could resize the frame to an arbitrary number.

**Finding 2 — took the smaller half of your suggestion.** Both files now carry a comment pointing at the other and saying why the duplication exists and that nothing checks the two agree. Factoring them into one generated file is the real fix, but it is a larger change than this one should carry, and it should cover both listeners rather than only the new one.

Re-verified after all three: the generated script still parses under `node --check` and the `TABBABLE` selector is unchanged character for character; `testthat::test_local()` clean, no failures or errors; and in Chromium against a real Quarto 1.10.18 `revealjs` deck, Shift + Tab still lands on the current slide and Space still advances the deck.

---

jooyoungseo:
Fair catch on the missing test — that one is on me by this repo's own rule, not just as polish: I added behaviour and left nothing holding it in place, while the rest of the script's contract is pinned exactly that way. `23bcb31` adds the sixth case, in the same `grepl` style, asserting both halves of the guard (the type check and the floor).

Also pushed `9361882`: an em dash had slipped into the comment `da9ac88` added, and it was the only non-ASCII character in `inst/htmlwidgets/maidr.js`. Nothing fails on it, but the file was ASCII before and its prose already spells a dash `--`.

`testthat::test_local()` still clean with the new case.

Two notes for whoever merges:

- The engine half, [`xability/maidr#1111`](https://github.com/xability/maidr/pull/1111), is merged (`c92e821`). Neither half does anything alone — the chart has to ask before this listener can answer — and r-maidr loads the 4.4.0 bundle today, so both reach users together with the next engine release.
- `main` has moved to `d781a2f` since this opened. I ran the merge locally: `NEWS.md` and `R/svg_utils.R` both auto-merge, no conflict, so this needs no rebase.

---

PR #213, opened 2026-08-21
fix(iframe): check that the page actually took focus before calling it done

jooyoungseo:
## Description

The focus handoff added in #208 assumed that calling `focus()` on an element moved focus there. It does not always: an element with no rendered box refuses silently, and `display: contents` is the case that bites.

Shiny wraps **every** output in a `display: contents` div, and a chart's frame sits directly inside one with no sectioning element above it. So the handoff picked that wrapper, called `focus()`, got nothing, and returned as though it had succeeded — leaving the reader inside the chart with nothing to show for the keypress. That is the very failure the handoff exists to prevent, reproduced in a different host.

## Changes Made

Both copies of the parent-side listener — `maidr_iframe_host_script()` in **`R/svg_utils.R`** and the widget binding's own copy in **`inst/htmlwidgets/maidr.js`** — gain a `takeFocus()` helper, and every candidate now goes through it:

- it asks the element to take focus and reads the outcome back (`document.activeElement === el`) rather than trusting the call, walking on up the ancestor chain until one actually takes it, the sectioning element still being tried first;
- it asks an element **as it stands** before giving it anything, so a tab stop the page already owns is never pulled into `tabindex="-1"` to hold focus it could already hold;
- and a `tabindex` it did add is **removed again** when the element still refuses, so a wrapper that cannot hold focus is not left carrying an attribute claiming it can.

The tab stop found by `stopBefore()` goes through the same check as the container. It has to: `stopBefore()` returns whatever matched the tabbable selector, and a `display: contents` element carrying a `tabindex` matches it and passes `reachable()` while still being unable to hold focus, so trusting that one strands the reader exactly as trusting the container did.

The roxygen block described only the pre-fix behaviour, so it now says what the function does, and `man/` is regenerated from it. Three cases in **`tests/testthat/test-svg-utils.R`** pin the contract, including the ordering of the `focus()` call against the `setAttribute()` so a flip would be caught.

## Related

- [`xability/maidr#1121`](https://github.com/xability/maidr/pull/1121) is the engine half, which had the same assumption and the same fix. Merged as `67f2a85`.
- Follow-up to #208, which introduced the handoff. Shiny was not among the hosts checked before that merged — this is the gap.

## Verification

`testthat::test_local()` clean — no failures, no errors. The generated script still parses under `node --check`, and `node --check` passes on the widget binding.

Driven in Chromium against a real py-shiny app, sampling the host document after one Shift + Tab out of the chart:

```
before                                    after
t=  50ms  top=iframe#…  frameHasFocus=1   t=  50ms  top=div.container-fluid  frameHasFocus=0
t=2000ms  top=iframe#…  frameHasFocus=1   t=2000ms  top=div.container-fluid  frameHasFocus=0

tabindex written on: div.shiny-html-output    tabindex written on: div.container-fluid
                     (display:contents,                            (the one that took it)
                      never took focus)
```

Walking that page's ancestor chain, measured rather than assumed:

```
div#first_chart.shiny-html-output   display=contents   takesFocus=false
div.container-fluid                 display=block      takesFocus=true
body                                display=block      takesFocus=true
```

Re-checked afterwards, and again after the second commit, that the hosts #208 was verified against are unchanged: a three-slide Quarto 1.10.18 `revealjs` deck with a ggplot2 chart on each slide still lands on the current slide with Space advancing on every slide, and a JupyterLab notebook and an nbconvert export still keep their native tab order with nothing written into them.

## Note

The tests here match the generated script's text rather than exercising a real DOM, as every other test in this file does — there is no JS test harness in this repo, and adding one is a larger change than this should carry. Browser runs cover the behaviour instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,13 @@
 
 ## Bug Fixes
 
+* The chart could still strand a reader in a Shiny app. Handing focus back to
+  the page assumed that asking an element to take it worked, and an element
+  with no rendered box refuses silently -- Shiny wraps every output in a
+  `display: contents` div, and a chart's frame sits directly inside one. The
+  page now reads the outcome back and walks on up the ancestors until one
+  actually takes focus, leaving no `tabindex` behind on the ones that refuse.
+
 * A chart in a Quarto `revealjs` slide swallowed every key, with no way back to
   the deck. Charts are embedded in an iframe, and keyboard events do not cross
   a frame boundary, so while a reader is inside a chart the slide around it
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,10 @@
   `display: contents` div, and a chart's frame sits directly inside one. The
   page now reads the outcome back and walks on up the ancestors until one
   actually takes focus, leaving no `tabindex` behind on the ones that refuse.
+  The tab stop before the frame is checked the same way, since one of those can
+  be a `display: contents` element too, and every candidate is asked as it
+  stands before being given a `tabindex` --- so a tab stop the page already
+  owns is never taken out of its own tab order.
 
 * A chart in a Quarto `revealjs` slide swallowed every key, with no way back to
   the deck. Charts are embedded in an iframe, and keyboard events do not cross
--- a/R/svg_utils.R
+++ b/R/svg_utils.R
@@ -1493,6 +1493,19 @@ maidr_iframe_host_script <- function() {
     "var s = window.getComputedStyle(el);",
     "return s.display !== \"none\" && s.visibility !== \"hidden\";",
     "}",
+    # Asking an element to take focus is not enough. `focus()` on an element
+    # with no rendered box -- a `display: contents` wrapper, which is what
+    # Shiny puts around every output -- is a silent no-op, so the outcome has
+    # to be read back. A tabindex this added is removed again when the element
+    # refuses, rather than leaving it claiming it can hold focus.
+    "function takeFocus(el) {",
+    "var added = !el.hasAttribute(\"tabindex\");",
+    "if (added) el.setAttribute(\"tabindex\", \"-1\");",
+    "el.focus();",
+    "if (document.activeElement === el) return true;",
+    "if (added) el.removeAttribute(\"tabindex\");",
+    "return false;",
+    "}",
     "function stopBefore(frame) {",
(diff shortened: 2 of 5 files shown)


jooyoungseo:
All three addressed in `906c775`.

**Finding 2 was the real one, and thank you for it.** You were right that it is the same class of bug in the sibling branch: `stopBefore()` returns whatever matched `TABBABLE`, `reachable()` only rules out `display: none` / `visibility: hidden` / hidden ancestors, and a `display: contents` element carrying a `tabindex` passes both while still being unable to hold focus. Trusting that one strands the reader exactly as trusting the container did. It now goes through `takeFocus()`.

That change needed a second half. `takeFocus()` used to add `tabindex="-1"` to anything lacking it, which is harmless for the container walk but not here — `stopBefore()` hands back tab stops the page already owns, and writing `tabindex="-1"` onto a link or a button would take it out of the page's own tab order in order to make it hold focus it could already hold. So every candidate is now asked as it stands, and the attribute only follows a refusal.

Worth noting for the record: the same "ask first" idea was raised on the engine half ([`xability/maidr#1121`](https://github.com/xability/maidr/pull/1121)) and I backed it out there, because on that path a natively tabbable ancestor precedes the frame in document order and so stops the service from firing at all — unreachable, and no test could pin it. Here it is reachable, precisely because `stopBefore()` deals in tab stops rather than ancestors.

**Finding 1** — the roxygen block only described the pre-fix behaviour. It now says what the function does (outcome read back, ancestors walked, asked-before-written), and `man/maidr_iframe_host_script.Rd` is regenerated from it with roxygen2 8.1.0.

**Finding 3** — a fair limitation, and I have left it as it is. There is no JS test harness in this repo, and adding one for this would be a larger change than this PR should carry; the new cases pin the contract the same way the rest of the file does, including the ordering of the `focus()` call against the `setAttribute()` so a flip would be caught. Real behaviour is covered by browser runs instead, and those were repeated after this commit.

**Re-verified after the change:** `testthat::test_local()` clean, no failures or errors; the generated script still parses under `node --check`, as does the widget binding; and the three-slide Quarto `revealjs` deck still lands on the current slide with Space advancing, checked on every slide.

On the sandbox note in your review — `Rscript`, `node` and the browser runs were all done locally here, so those results are from real runs rather than static reading.

---

Issue #214, opened 2026-08-21
Base R `contour()` ships a layer typed `"unknown"`, which the core refuses — so the chart is neither read nor a picture


jooyoungseo:
Found by sweeping the base R functions the adapter classifies, the way #211 came out of sweeping `geom_*`.

## Measured

```r
contour(matrix(1:12, nrow = 3))
maidr::save_html(plot = NULL, file = f)
```

renders **interactively** — no static-image fallback — and the payload carries:

```
types=[unknown]
```

`unknown` is not a trace type. `src/model/factory.ts` in the core ends its dispatch with

```ts
default:
  throw new Error(`Invalid trace type: ${layer.type}`);
```

so the figure never binds. The reader gets an interactive-looking shell that answers no key, **and** no picture either, because the fallback that would have drawn one was not taken. That is worse than both of the outcomes the package already has for a chart it cannot read.

Compare the neighbours in the same sweep, which behave correctly:

| call | result |
| --- | --- |
| `contour(m)` | **interactive, `type: "unknown"`** — never binds |
| `dotchart(...)` | static image, "Plot contains unsupported elements" |
| `stripchart(...)` | static image, same warning |
| `mosaicplot(...)` | static image, same warning |
| `image(m)` | interactive, `heat` |
| `plot(x, y)` | interactive, `point` |

## Why it happens

Two lines of `R/base_r_processor_factory.R` contradict each other. `get_supported_types()` claims it:

```r
"heat",
"contour",     # <- listed as supported by the Base R system
"candlestick",
```

and `create_processor()` gives it the generic one:

```r
"heat" = BaseRHeatmapLayerProcessor$new(layer_info),
"contour" = BaseRUnknownLayerProcessor$new(layer_info),   # <- the fallback processor
"candlestick" = BaseRCandlestickLayerProcessor$new(layer_info),
```

`R/base_r_adapter.R:224` maps the call to the type (`"contour" = "contour"`), so the classification is complete and only the processor is missing — and because `"contour"` is in `get_supported_types()`, the unsupported-elements path that saves `dotchart` never runs.

The explicit `"contour"` line is also redundant with the `default` branch two lines below it, which returns the same processor. Its only effect is to make the omission look deliberate.

## Proposed

[...]

Issue #211, opened 2026-08-21
A layer that draws nothing still costs the whole chart its interactivity — `geom_blank()` and `geom_label()`


jooyoungseo:
Found by sweeping every `geom_*` ggplot2 exports through `Ggplot2Adapter$detect_layer_type()`, rather than by hitting one of them.

## The two that should not cost anything

`detect_layer_type` returning `"unknown"` is what makes `has_unsupported_layers()` true, which drops the **whole plot** to a static image. #176 measured that cost for a reference line and #197 for an annotation; both were fixed by returning `"skip"` instead. Two more layers pay it today, and neither carries an observation to lose.

Measured with `save_html()` on ggplot2 3.4.4, same three-bar chart in each row:

| plot | result | bytes |
| --- | --- | --- |
| `geom_col()` | interactive SVG | 39,116 |
| `geom_col() + geom_text(aes(label = v))` | interactive SVG | 41,339 |
| `geom_col() + geom_label(aes(label = v))` | **base64 image** | 17,220 |
| `geom_col() + geom_blank()` | **base64 image** | 13,380 |

- **`geom_blank()` draws nothing at all.** It exists to force a scale limit — `geom_blank(aes(y = 0))` to include zero, `geom_blank(data = ...)` to align facets — so it is added to charts that are otherwise entirely readable, and it takes every one of them down. There is no reading question here: the layer has no marks.
- **`geom_label()` is `geom_text()` with a rounded rectangle behind it.** The adapter already returns `"skip"` for `GeomText`. `GeomLabel` is a **sibling** of `GeomText`, not a subclass — `GeomLabel < Geom` — so the exact `geom_class == "GeomText"` test misses it, and the same annotation costs the chart everything depending only on which of the two the author reached for.

Both are one name added to an existing skip branch.

## The rest of the sweep

For the record, since the question "which layers land on unknown" has not been written down anywhere. Every `geom_*` ggplot2 exports, classified with no plot context:

| geom | today | note |
| --- | --- | --- |
| `geom_blank` | unknown | **above** — draws nothing |
| `geom_label` | unknown | **above** — `geom_text` is skipped |
| `geom_rect` | unknown | #197 — the deciding rule is open |
| `geom_segment`, `geom_curve` | decided per plot | reads `gantt` when the spans lane up, `unknown` otherwise; the sweep sees `unknown` only because it passes no plot |
[...]

PR #217, opened 2026-08-21
fix(base-r): record the plot calls that stopped the save instead of falling back

### LINKED ISSUE #216: Seven base R plot calls are not recorded at all, so `save_html()` stops with "No Base R plots detected" for a plot that is on the device
Split out of the sweep in #214, which recorded these without filing them.

## Measured

Each of the seven below draws on the device, and then `save_html()` **stops**:

```r
maidr:::clear_all_device_storage()
grDevices::pdf(NULL)
qqnorm(c(1, 2, 3, 4, 5, 6, 7, 8))
maidr::save_html(plot = NULL, file = f)
#> Error: No Base R plots detected. Please create a plot first
#>        (e.g., barplot(), plot()).
```

| call | today |
| --- | --- |
| `persp(z = m)` | **error** — "No Base R plots detected" |
| `sunflowerplot(x, y)` | **error** |
| `fourfoldplot(m)` | **error** |
| `spineplot(f ~ x)` | **error** |
| `cdplot(f ~ x)` | **error** |
| `qqnorm(x)` | **error** |
| `filled.contour(m)` | **error** |
| `dotchart(v)` | static image, "Plot contains unsupported elements" |
| `mosaicplot(m)` | static image, same warning |
| `stripchart(v)` | static image, same warning |

The message is not merely unhelpful, it is **false**: the caller has drawn a plot, it is on the device, and they are told there isn't one. `dotchart` and `mosaicplot` are unread too, and they degrade correctly — the difference is only that those names are in `.base_r_function_classes$HIGH` and these seven are not.

## Why it happens

`classify_function()` returns `"UNKNOWN"` for a name that is in none of the three lists, so the wrapper is never installed and the call is never logged. `can_handle()` then sees a device with no recorded calls, and `save_html()` raises the message quoted above — which is the right message for `plot.new()` and the wrong one here.

Nothing else is missing. Once a name is recorded, `detect_layer_type()` has no branch for it, its `switch` falls through to the `"unknown"` default, `unsupported_layer_flags()` sees `"unknown"`, and the static-image path runs. That is the same route `dotchart` takes.

## Proposed

Add the seven to `HIGH` (and the matching `@export` stubs, since that is how base R wrappers reach the search path). Being listed there is the **lower** of the two claims the package can make about a call — "recorded, therefore at worst a picture" — not a claim that it is read.

## One thing that has to come with it

`qqnorm(x, plot.it = FALSE)` is the idiomatic way to *compute* theoretical quantiles; it draws nothing. Recorded, it puts a call on a blank device, and the save then fails a different way:

```
Error: Failed to create fallback image
```

The wrapper already declines to record `hist(x, plot = FALSE)` and `boxplot(x, plot = FALSE)` for exactly this reason. `plot.it` is the same request under the spelling `qqnorm()` and `qqplot()` use, so the guard needs the second name.

Measured after both changes:

| call | result |
| --- | --- |
| `qqnorm(x)` | static image + "unsupported elements" |
| `qqnorm(x, plot.it = FALSE)` | not recorded; "No Base R plots detected" — correct, nothing was drawn |
| `cdplot(f ~ x, plot = FALSE)` | not recorded (`cdplot.default` has a real `plot` argument) |
| `barplot(v)` after any of the seven | still interactive, `type: "bar"` |

The decoded fallback image is the chart, not a blank canvas — a `spineplot()` drawn through a formula comes back at 1050×750 and 26,500 bytes with both axis labels intact, against 859 bytes for an empty device of the same size. That matters because the replay has to reconstruct language arguments in the caller's frame.

## Two things deliberately not in this

- **`sunflowerplot(..., add = TRUE)` will now cost its chart the interactive reading.** It is HIGH, so it opens a plot group, types `"unknown"`, and the figure falls back. Before, the overlay was invisible to maidr and the underlying `plot()` read interactively — which means the reader was told a complete chart while a layer of real observations went unannounced. Falling back is the honest outcome and the one #176 and #211 settled on; it is called out here because it is a change to a case that previously appeared to work.
- **`qqnorm` could be read rather than pictured.** It draws sample quantiles against theoretical ones, which is a `point` layer, and `qqnorm.default` hands back exactly those two vectors. It is not attempted here for the reason `contour` was not in #214: the reading is a separate decision with an unmeasured selector story (`qqnorm.default` calls `plot()` from `stats`' namespace, so no maidr-visible point layer is recorded today). Worth its own issue against #137.

Also noticed and **not** changed: the computation-only guard tests the argument by name without checking the function has such a formal, so `spineplot(f ~ x, plot = FALSE)` — where `plot` is not an argument of `spineplot` at all, and R itself warns `"plot" is not a graphical parameter` — is declined even though the chart is drawn. That predates this issue for every already-recorded name (`plot()`, `barplot()` and the rest have no `plot` formal either), the input is one R already objects to, and tightening the guard would change behaviour well outside these seven.

## Related

- #214 — the sweep this came out of, and the same "unread costs more than it should" family
- #137 — the base R half of the coverage roadmap

jooyoungseo:
Closes #216.

## What was wrong

Eight base R calls were in none of `.base_r_function_classes`' three lists, so `classify_function()` returned `"UNKNOWN"`, no wrapper was installed, and nothing was logged. `save_html()` then found a device with no recorded calls and **stopped**:

```r
grDevices::pdf(NULL)
qqnorm(c(1, 2, 3, 4, 5, 6, 7, 8))
maidr::save_html(plot = NULL, file = f)
#> Error: No Base R plots detected. Please create a plot first
#>        (e.g., barplot(), plot()).
```

The message is not merely unhelpful, it is false: the plot is on the device. Measured on all eight before the change — `persp`, `sunflowerplot`, `fourfoldplot`, `spineplot`, `cdplot`, `qqnorm`, `qqplot`, `filled.contour` — every one raised it. `dotchart`, `stripchart` and `mosaicplot` are equally unread and degrade correctly, and the only difference is that those names are in `HIGH`.

(#216 lists seven. `qqplot` is the eighth, found in review of this PR: `stats::qqplot` was no more classified than `stats::qqnorm` was, and it reproduced the identical error.)

## What changed

The eight names are added to `HIGH`, with the matching `@export` stubs since that is how a base R wrapper reaches the search path. Nothing else was needed: `detect_layer_type()` has no branch for any of them, so its `switch` falls through to the `"unknown"` default, `unsupported_layer_flags()` sees it, and the static-image path runs — the route `dotchart` already takes.

Being listed in `HIGH` is the **lower** of the two claims the package can make about a call — recorded, therefore at worst a picture — not a claim that the chart is read. The comment beside the block says so, because the list is the natural place to look when adding a reading later and adding one does not mean touching it.

| call | before | after |
| --- | --- | --- |
| `persp(z = m)` | error | static image + "unsupported elements" |
| `sunflowerplot(x, y)` | error | static image + warning |
| `fourfoldplot(m)` | error | static image + warning |
| `spineplot(f ~ x)` | error | static image + warning |
| `cdplot(f ~ x)` | error | static image + warning |
| `qqnorm(x)` | error | static image + warning |
| `qqplot(x, y)` | error | static image + warning |
| `filled.contour(m)` | error | static image + warning |
| `barplot(v)`, `plot(x, y)`, `hist(x)` | interactive | interactive |

## The half that had to come with it

`qqnorm(x, plot.it = FALSE)` is the idiomatic way to *compute* theoretical quantiles, and it draws nothing. Once `qqnorm` is recorded, that call put an entry on a blank device and the save failed a different way — `Error: Failed to create fallback image`, because the fallback re-executes the recorded calls and there was no drawing to capture.

The wrapper already declined `hist(x, plot = FALSE)` and `boxplot(x, plot = FALSE)` for exactly this reason. `plot.it` is the same request under the spelling `qqnorm()` and `qqplot()` use, so the guard gained the second name. Measured after: both `plot.it = FALSE` calls are unrecorded and the save reports "No Base R plots detected" — which is correct, nothing was drawn — while the drawing spellings fall back to a picture.

The two are asserted separately, because the guard reads the argument off the **recorded** call and so protects a second function only once that function is recorded at all. `cdplot.default` has a real `plot` argument, unlike most of the eight, so the pre-existing half of the guard already covered it; that is asserted rather than assumed.

## One case that changes for the worse, deliberately

`sunflowerplot(..., add = TRUE)` now costs its chart the interactive reading. It is `HIGH`, so it opens a plot group, types `"unknown"`, and the figure falls back. Before, the overlay was invisible to maidr and the underlying `plot()` read interactively — which means a reader was handed a chart presented as complete while a layer of real observations went unannounced. #176 and #211 settled that trade the same way; it is called out because it is a change to a case that previously appeared to work.

## Tests

`tests/testthat/test-base-r-unrecorded-calls.R`, 60 assertions. Every one of the eight is driven through a real `save_html()` on an off-screen device.

The one worth naming is **"the picture is the chart, not a blank canvas"**. The three assertions above it would all pass against a fallback that captured an empty page, and the replay has to rebuild `factor(...) ~ x` in the caller's frame to draw a `spineplot`. So the embedded image is decoded and measured: 1050×750 from the PNG header, and over 10 KB against 859 bytes for an empty device of the same size.

Falsified — each mutation applied alone against the fixed baseline:

| mutation | result |
| --- | --- |
| all eight removed from `HIGH` | 21 failures |
| `filled.contour` alone removed | 4 failures |
| `qqplot` alone removed | 4 failures |
| `plot.it` half of the guard removed | "a call that computes without drawing is still not recorded" fails |
| the guard returns `invisible(NULL)` instead of the result | "qqnorm still hands back the quantiles it computed" fails |
| the replay skips the recorded `spineplot` | "the picture is the chart" fails on all four assertions — and the save fails outright, which is the same "Failed to create fallback image" the `plot.it` guard exists to prevent |

Full suite: `[ FAIL 0 | WARN 8 | SKIP 99 | PASS 6124 ]`. `NAMESPACE` and `man/base-r-wrappers.Rd` regenerated with roxygen2 8.1.0, the version `Config/roxygen2/version` pins. `NEWS.md` gains an entry naming all eight, since a reader who hit the error has no way to know which of the names they reached for was the missing one.

## Recorded, not fixed here

- **`qqnorm` and `qqplot` could be read rather than pictured.** They draw sample quantiles against theoretical ones, which is a `point` layer, and `qqnorm.default` returns exactly those two vectors. Not attempted for the reason `contour` was not attempted in #214: it is a separate reading decision with an unmeasured selector story, since `qqnorm.default` calls `plot()` from `stats`' namespace and no maidr-visible point layer is recorded.
- **The computation-only guard tests an argument name without checking the function has such a formal.** `spineplot(f ~ x, plot = FALSE)` is declined even though the chart is drawn — `spineplot` has no `plot` argument at all and R itself warns `"plot" is not a graphical parameter`. That predates this change for every already-recorded name (`plot()` and `barplot()` have no `plot` formal either), and tightening it would change behaviour well outside these eight.


jooyoungseo:
Both findings acted on in 85b8524.

## `qqplot()` — you were right, and it is now the eighth name

Measured before the change, exactly as you predicted:

```r
grDevices::pdf(NULL)
qqplot(c(1, 2, 3, 4), c(2, 3, 4, 5))
maidr::save_html(plot = NULL, file = f)
#> Error: No Base R plots detected. Please create a plot first
#>        (e.g., barplot(), plot()).
```

`classify_function("qqplot")` returned `"UNKNOWN"`, no wrapper, no recorded call — the same defect, not a related one. Deferring it would have been the wrong call: the PR's own rationale for the `plot.it` guard names `qqplot` as the other function that spells it that way, so leaving it out meant the guard was written for a function the package still could not see.

Added to `HIGH` with its `@export` stub, `NAMESPACE` and `man/` regenerated, and `stats::qqplot` has a real `plot.it` formal so the guard covers it — asserted separately from `qqnorm`, because the guard reads the argument off the *recorded* call and therefore protects a second function only once that function is recorded at all.

The test file's fixtures, wording and the classification comment all say eight now. Falsified: removing `qqplot` alone from `HIGH` fails 4 assertions across three tests.

## `NEWS.md`

Added, under the existing `## Bug Fixes` heading. It names all eight rather than only the new one, since a reader who hit the error has no way to know which of the names they reached for was the missing one; it also records that being listed is not a claim the chart is read, and that the two `plot.it = FALSE` spellings are still declined.

## Re-run

`[ FAIL 0 | WARN 8 | SKIP 99 | PASS 6124 ]` — 60 assertions in `test-base-r-unrecorded-calls.R`, up from 54.

Nothing was found that changes the two items under "Recorded, not fixed here": reading `qqnorm` (and now `qqplot`) as the point layer it draws is still a separate decision with an unmeasured selector story, and the computation-only guard still tests an argument name without checking the function declares such a formal.

---

jooyoungseo:
Windows R-CMD-check went red on 4ca7ea5, and it was my test, not the change. Fixed in 0a38f27.

```
── Failure ('test-base-r-unrecorded-calls.R:149:3'): the picture is the chart, not a blank canvas ──
Expected `length(decoded)` > 10000.
Actual comparison: 9600.0 <= 10000.0
```

The threshold came from measuring the fallback PNG on Linux — 26,500 bytes against 859 for an empty device — and Windows encodes the identical `spineplot` to 9,600. Same drawing, different PNG encoder. So the assertion was not asking "did the replay draw the chart", it was asserting which platform ran it. macOS and both Ubuntu jobs passed, which is exactly how a constant like that hides.

It now renders **two references on the same machine**, through the same device `create_fallback_image()` uses (7×5 inches at 150 dpi), and requires the captured image to be nearer the chart than an empty page:

```r
blank <- png_bytes(function() graphics::plot.new())
reference <- png_bytes(unrecorded_calls$spineplot)
expect_gt(reference, blank)
...
expect_gt(length(decoded), (blank + reference) / 2)
```

No constant, and it asks the question the check was always meant to ask. The PNG-header dimension assertions are unchanged — those passed on Windows, so the device geometry is stable across platforms even when the encoding is not.

Re-falsified with the same mutation: making the replay skip the recorded `spineplot` still fails this test on all four of its assertions, plus the two that notice the save failing outright.

`[ FAIL 0 | WARN 0 | SKIP 0 | PASS 61 ]` locally.

---

PR #219, opened 2026-08-22
feat(base-r): read a contour as the curves it draws, not as a picture

### LINKED ISSUE #218: Read a base R `contour()` as the contour it draws, instead of falling back to a picture
The half #214 left open. That issue stopped `contour()` shipping a layer typed `"contour"` with no processor behind it — which made the figure bind interactively and then fail to construct — by typing it `"unknown"` so it degrades to a static image. `tests/testthat/test-base-r-contour-fallback.R` says so in its own header:

> Reading it properly is still open: `contour` *is* a trace type, the ggplot2 adapter emits it for `geom_contour()` (#198), and base R hands over the same x, y, z and levels.

I recorded one blocker at the time — that the gridGraphics grob naming for selectors was unresolved. **It is resolved.** Measured on R 4.3.3 / gridGraphics.

## The data

`grDevices::contourLines()` reproduces exactly what `contour()` draws, using the same defaults:

```r
contour.default(x = seq(0, 1, length.out = nrow(z)),
                y = seq(0, 1, length.out = ncol(z)),
                z, nlevels = 10, levels = pretty(zlim, nlevels),
                zlim = range(z, finite = TRUE))
```

So the levels are reproducible without guessing, and each curve arrives as `{level, x, y}` — the shape `ggplot2_contour_layer_processor.R` already emits, where a curve is a list of `{x, y, level}`.

## The selectors — the part that was open

gridGraphics gives **one `lines` grob per curve**, named and ordered:

```
graphics-plot-1-contour-1-1  [lines]
graphics-plot-1-contour-2-2  [lines]
…
graphics-plot-1-contour-8-8  [lines]
```

Eight grobs for the eight curves `contourLines()` returns. And the two orders correspond — checked vertex by vertex rather than assumed:

| grob | vertices | `contourLines[i]` | level | vertices |
| --- | --- | --- | --- | --- |
| `contour-1-1` | 2 | 1 | 2 | 2 |
| `contour-2-2` | 4 | 2 | 3 | 4 |
| `contour-3-3` | 4 | 3 | 4 | 4 |
| `contour-4-4` | 4 | 4 | 5 | 4 |
| `contour-5-5` | 4 | 5 | 6 | 4 |
| `contour-6-6` | 2 | 6 | 7 | 2 |
| `contour-7-7` | 2 | 7 | 8 | 2 |
| `contour-8-8` | 2 | 8 | 9 | 2 |

Same count, same order, same sequence of vertex counts. A `lines` grob renders as `<polyline>`, so `generate_robust_selector(gt, "lines", "polyline", …)` is the existing route.

## Proposed

A `BaseRContourLayerProcessor` alongside the other base R processors:

- **data** — `contourLines(x, y, z, levels)` with `contour()`'s own defaults, one curve per entry, each a list of `{x, y, level}`, matching the ggplot2 processor's payload exactly.
- **selectors** — one per curve, off the per-curve grobs.
- **adapter** — `"contour" = "contour"` again, and `create_processor()` gains the case that was missing when #214 found the contradiction.

One known limitation to write down rather than discover: gridGraphics warns `cannot emulate labels on contour lines`, so the inline level labels a sighted reader sees on the curves are not in the grob tree. That costs nothing here — the level travels on every point of its curve, which is where the frontend's contour trace reads it from.

## Related

- #214 — the fallback this replaces, and the test that records this as open
- #198 — the ggplot2 side, whose payload shape this matches
- #200 — the processor registry this must register with

jooyoungseo:
Closes #218 — the half #214 left open, and which `test-base-r-contour-fallback.R` has been recording as open in its own header.

## Before

`contour()` fell back to a static image. That was deliberate: the call had been shipping a layer typed `"unknown"` that the core's trace factory throws on, so the figure bound interactively and then failed to construct — an interactive shell answering no key, and no picture either. A picture with a warning was the better of two bad answers, not a good one.

## The two halves, both measured

**Data.** `grDevices::contourLines()` is the computation `contour()` itself does, and `contour.default`'s own defaults make it reproducible without guessing:

```r
contour.default(x = seq(0, 1, length.out = nrow(z)),
                y = seq(0, 1, length.out = ncol(z)),
                z, nlevels = 10, levels = pretty(zlim, nlevels),
                zlim = range(z, finite = TRUE))
```

The payload matches `ggplot2_contour_layer_processor` exactly — a list of curves, each a list of `{x, y, level}` — so the two adapters describe one chart alike.

**Selectors — the part recorded as unresolved.** gridGraphics writes one `lines` grob per curve, and the two orders correspond. Checked vertex count by vertex count rather than assumed:

| grob | vertices | `contourLines[i]` | level | vertices |
| --- | --- | --- | --- | --- |
| `contour-1-1` | 2 | 1 | 2 | 2 |
| `contour-2-2` | 4 | 2 | 3 | 4 |
| … | | | | |
| `contour-8-8` | 2 | 8 | 9 | 2 |

Verified end to end too: the first announced point comes back at `x=0.496, y=0`, which is where the `contour-1-1` grob starts — so the data and the selector address the same curve. All 8 ids appear in the rendered SVG, each holding one `<polyline>`.

## A bug this PR's own tests caught

The wrapper records a **partially named** call. Measured: `contour(c(10, 20, 30), c(100, 200, 300), z)` arrives as one unnamed argument plus `y =` and `z =`.

My first version looked for the matrix and took the unnamed arguments before it. It found `z` already named, so it never looked at the unnamed one, and announced the caller's grid on the 0-1 default — **every coordinate wrong, and nothing raised.** The test asserting the caller's x is used caught it.

Now resolved the way `contour.default` resolves it: named arguments claim their slots, unnamed ones fill what is left in order, and the `if (missing(z) && !is.list(x)) z <- x` rule runs afterwards — which is what makes `contour(m)` a contour *of* `m` rather than a chart with `m` on the x axis.

## Tests whose premise changed

Six existing tests asserted contour is *unsupported*. Updated, not deleted, because what outlived both fixes is the rule they were two answers to: **a call is either dispatchable or it falls back, and only `"unknown"` on `layer$type` reaches the fallback.** `dotchart` still stands for the unclaimed side; `test-base-r-contour-fallback.R` keeps its "never emits a type the core would refuse" assertion completely unchanged, since that holds either way.

## Verification

Full suite `6192 passed, 0 failed, 99 skipped`. Docs regenerated with the pinned roxygen2 8.1.0. NEWS updated.

Four mutations, each applied alone:

| mutation | fails |
| --- | --- |
| adapter back to `"unknown"` | 1 — the chart is a picture again |
| drop the `z <- x` rule | 4 — `contour(z)` resolves nothing |
| sort selectors lexicographically | 1 — the >10-curve case only |
| ignore an explicit `levels` | 1 — the explicit-levels case only |

## Known limitation, stated rather than discovered

gridGraphics warns `cannot emulate labels on contour lines`, so the inline level labels a sighted reader sees along the curves are not in the grob tree. That costs nothing here — the level travels on every point of its curve, which is where the frontend's contour trace reads it from.

---


jooyoungseo:
Pushed `4b9412a`. Both findings acted on, #1 measured before deciding.

## #1 — the one-vertex curve and the count guard

You're right about the shape of the risk: the payload filters, the grobs don't, and the guard drops **every** selector on mismatch — a whole layer's highlighting for one degenerate curve.

Measured whether it can arise. It cannot, and the reason is more interesting than "it doesn't happen":

| field | curves | min vertices | any 1-vertex |
| --- | --- | --- | --- |
| paraboloid, level exactly at the minimum | **0** | — | no |
| saddle through zero | 2 | 4 | no |
| single corner touching the level | **0** | — | no |
| one-cell spike | 3 | 5 | no |
| `volcano` at its own levels | 20 | 5 | no |
| 400 random 4×4 fields | — | — | **0 of 400** |

A level the surface merely *touches* yields no curve at all rather than a one-vertex one. And gridGraphics agrees — drawing `contour(g, g, bowl, levels = c(0, 0.5))` writes **one** grob, not two. So the two sides already correspond.

But relying on that holding forever is the fragile part, so I made the correspondence explicit rather than leaving it as an observation: `read_curves` now reports which curves it kept, each announced curve takes the grob **at its own index**, and the count check compares against every curve *drawn* rather than every curve announced. If the filter ever does fire, the layer loses one curve instead of all its highlighting.

Pinned two ways — a direct test that a middle-curve drop yields selectors 1 and 3 rather than 1 and 2, and the tangent measurement itself. Indexing by position instead of by `kept` fails the first.

## #2 — the list form

A real gap, not an intentional omission — I'd simply not reproduced that branch of `contour.default`. `contour(list(x =, y =, z =))` fell through to `!is.matrix(z)` and returned `NULL`, i.e. silently no data. Now unpacked the way `contour.default` unpacks it, with a test; removing the branch fails it.

## Verification

Full suite `6200 passed, 0 failed, 99 skipped` (up 8). Docs regenerated with the pinned roxygen2 8.1.0.

On not being able to run R here — noted, and worth saying your static read found the one thing my green suite could not: a guard whose failure mode is silent and whose trigger my fixtures never hit.

---

PR #223, opened 2026-08-22
feat(rug): read a rug as the observations it marks

### LINKED ISSUE #222: geom_rug() reads as nothing at all, though it draws one tick per observation
## The gap

`geom_rug()` falls through to `Ggplot2UnknownLayerProcessor`. Measured, via `Ggplot2PlotOrchestrator$new(p)$generate_maidr_data()`:

```
rug alone       layers: 1   type=unknown  n=0  selectors=
points + rug    layers: 2   type=point    n=20 selectors=g#geom_point\.points\.31\.1 > use
                            type=unknown  n=0  selectors=
rug + colour    layers: 1   type=unknown  n=0  selectors=
```

A chart whose only layer is a rug produces one empty layer and nothing to navigate. A rug beside a scatter contributes an empty layer that a reader lands on and finds nothing in.

`GeomRug` appears nowhere under `R/` — it is not in a dispatch branch, not in the skip list #197 added for annotations, not in `get_supported_types()`. It is unhandled rather than declined, which is the distinction #197 drew.

## Why it is data

A rug marks the **raw observations** — one short tick each — which is exactly what the density curve or histogram it usually sits under does not state. That is the reasoning py-maidr settled on in xability/py-maidr#250 and has read as a scatter of ticks ever since; xability/py-maidr#597 has just split a hue-grouped one into a layer per level. The same chart is currently readable in Python and silent in R, which is the parity argument #137 opens with:

> Which plotting system a user chose should not decide whether their chart is accessible.

## What the built data already carries

The read needs no reconstruction. `ggplot_build()` gives one row per observation with the position in the column naming the side it was drawn on:

```r
ggplot(df, aes(x = v)) + geom_rug()      ggplot(df, aes(y = v)) + geom_rug()
    x PANEL group colour                     y PANEL group colour
1 1.5     1    -1  black                  1 1.5     1    -1  black
2 2.5     1    -1  black                  2 2.5     1    -1  black
3 3.5     1    -1  black                  3 3.5     1    -1  black
4 7.0     1    -1  black                  4 7.0     1    -1  black
```

So the orientation is readable from which column is present, and the observations are the column itself.

## The reading

Follow py-maidr's, which is settled and tested: a **point** layer, one entry per tick, the observation on the axis it was drawn against and a constant on the other. `height` is one number for the whole layer, so the tick's *length* is decoration and only its position is data — the same argument `EventPlot` makes about an event plot's ticks. The across axis is labelled "Rug" rather than left as whatever the panel calls it, because every entry sits at the same place on it and a rug over a density has a real "density" label there that no tick measures.

## The part that needs deciding: selectors

This is not a copy of the py-maidr work, and the difference is worth stating before anyone starts.

`geom_rug` draws **one unnamed `segments` grob** holding every tick:

```
grill.gTree.11
GRID.gTree.2
    child: GRID.segments.1        <- all four ticks, one grob
panel.border..zeroGrob.3
```

Note the name. A point layer gets `geom_point.points.31`, which is what `generate_selectors` builds `g#geom_point\.points\.31\.1 > use` from. A rug gets `GRID.gTree.2` — a generic auto-generated name with no geom in it, and one that renumbers as other layers are added. Selecting on it directly would be brittle in exactly the way #204 was about.

Options, in the order I would try them:

1. **Position among the panel's children**, resolved the way the other unnamed grobs already are, if there is precedent in the codebase for that.
2. **One selector per segment** (`nth-of-type`) under whatever addresses the gTree, which is the shape py-maidr uses for a grouped rug.
3. **No selectors**, reading the layer without highlighting it. A rug announced with no visual highlight is still a large improvement on a rug announced as nothing, and #145 established that a layer without selectors is a coherent state rather than a broken one.

Worth measuring which before writing the processor, because it decides whether the layer can be highlighted at all.

## Scope

- A `Ggplot2RugLayerProcessor`, registered so the dispatch reaches it — the registry is derived from dispatch since #200, so the two cannot drift.
- Both orientations, from which column the built data carries.
- `colour =` grouping: seaborn's equivalent splits into a layer per level, and ggplot2's built data carries `colour` per row and `group` for free, so this may be cheaper here than it was there. Worth checking whether `group` is populated for a rug — it is `-1` for an ungrouped one above.
- The selector question settled by measurement first.

## Also swept, and not gaps

The same sweep over 27 geoms found everything else either read or deliberately declined:

- `geom_text` / `geom_label` produce **zero** layers, and that is intentional — `ggplot2_adapter.R:306` names both explicitly, as annotations, per #197.
- `geom_spoke` and `geom_polygon` land in `unknown`, and neither is a chart maidr has a trace for.
- `geom_ribbon`, `geom_crossbar`, `geom_pointrange`, `geom_linerange` and `geom_errorbarh` all read as `error_bar`; `geom_qq`, `geom_count` and `stat_summary` as `point`; `stat_ecdf` as `step`; `geom_bin2d`, `geom_tile` and `geom_raster` as `heat`; `geom_density` as `smooth`; `geom_freqpoly` as `line`.
- `geom_quantile` could not be measured here — `stat_quantile()` needs `quantreg`, which is not installed in this environment. Untested rather than passing.

jooyoungseo:
Closes #222.

## The gap

`GeomRug` appeared nowhere under `R/` — not in a dispatch branch, not in #197's annotation skip list, not in `get_supported_types()`. Unhandled rather than declined, which is the distinction #197 drew.

```
before                                   after
rug alone     1 layer  unknown  n=0      1 layer  point  n=4, 4 selectors
points + rug  2 layers point    n=20     3 layers point  n=20 / 4 / 4
              +        unknown  n=0
```

A rug-only chart produced one empty layer and nothing to navigate; a rug beside a scatter contributed an empty layer a reader lands on and finds nothing in. The same chart has been readable in py-maidr since xability/py-maidr#250 — the parity argument #137 opens with.

## The reading

Points, following py-maidr: `length` is one number for the whole layer, so a tick's *length* is decoration and only its position is data. The coordinate across the tick is a constant rather than the tick's own base, which is a fraction of the panel and would read as data at whatever scale the other axis happens to use.

The across axis is named **"Rug"** — the same string py-maidr uses, so the two bindings announce one chart the same way. Renamed even where the caller labelled it: a rug under a density curve has a real "density" label there, and every entry sits at 0 rather than at any density.

Nothing is reconstructed. `ggplot_build()` already gives one row per observation in `x`, `y`, or both.

## Two things measured rather than assumed

**A rug can mark both axes.** `sides` defaults to `"bl"`, so on a chart with both aesthetics mapped it marks the x observations *and* the y ones. Up to two layers — but not one per drawn grob, because `sides = "trbl"` draws the same x observations at top and bottom, and emitting that twice would have a reader navigate the same numbers under two names.

```
sides="b"      GRID.segments.1  : x (n=4)
sides="l"      GRID.segments.41 : y (n=4)
sides="bl"     GRID.segments.78 : x (n=4)   GRID.segments.79 : y (n=4)
sides="trbl"   .116: x   .117: x   .118: y   .119: y
```

**Which axes are marked needs both halves — and the first draft got it wrong.** Reading the built data's columns alone emitted a y layer for `geom_rug(sides = "b")` on `aes(v, w)`: a whole layer of observations the chart never marked, invented out of an aesthetic mapped for the scatter underneath. Reading `sides` alone fails the other way, since it says `"bl"` on a rug that has no `y`. A side counts only where both agree, which is exactly what the drawing does.

## Selectors

`geom_rug()` wraps its grobs in a `gTree` of its own, and that turned out to answer two questions at once:

```
rug + segment    GRID.segments.117            <- the segment layer, bare
                 GRID.gTree.120
                   GRID.segments.118          <- the rug, x
                   GRID.segments.119          <- the rug, y
```

One layer's grobs arrive as a block, so there is nothing to slice — and a rug is told from `geom_segment()`, whose grob is bare, without consulting anything else. Each tick is then `<grob>.1.<n>`, the form #194 measured for `geom_segment()` and `Ggplot2GanttLayerProcessor` already pins.

Which axis a grob stands on it answers itself: a tick standing on x is held constant in `y`, so `y0` is a single recycled value while `x0` carries one per observation. No reference to `sides` needed — the same property py-maidr's `read_rug` uses.

Where a rug draws one axis twice, the first grob for that axis wins: the two are copies of one observation, so highlighting one is partial and highlighting neither is worse — and #145 settled that a mismatched selector list is withdrawn wholesale rather than applied in part.

## Three existing tests retargeted

`test-ggplot2-reference-lines.R` used `geom_rug()` as its stand-in for "a geom maidr cannot read". Their intent — that an unsupported geom still falls back, and is still `unknown` rather than `skip` — is unchanged; the stand-in is now `geom_polygon()`, measured as `unknown`. A note at the first of the three records that the substitution happened and why, so the next reader meets it rather than deducing it from a failure.

## Verification

- **Eleven mutations, applied one at a time against a restored baseline, all killed** — dispatch removed, `sides` ignored, columns ignored, either clause dropped alone, the x/y constant swapped, the strip left unrenamed, the axis test inverted, the selector's `.1.<n>` shape swapped, always taking the first rug tree, the `sides` default changed.
- **Two survived at first, and neither was left alone.**
  - `axis %in% names(rows)` beside `any(!is.na(rows[[axis]]))`: a missing column indexes to NULL and `any(!is.na(NULL))` is already FALSE, so the two decide the same thing and no test could tell them apart. **Removed**, not pinned.
  - The `startsWith(name, "GRID.gTree")` check: measured across boxplot, violin, errorbar, crossbar, pointrange, linerange, step and a reference line, **none** draws a gTree whose children are all segments, so the class test alone decides today. **Kept as a guard and documented as unexercised**, with a test pinning that measurement — so the ggplot2 release that ends it turns a test red rather than leaving the check silently load-bearing.
  - The `sides` fallback survived a third round: `geom_rug()` always populates `sides`, so no chart reaches it, and `"b"` in place of `"bl"` passed every chart-level test while silently dropping the y half of every rug that omits the parameter. Pinned directly against `marked_axes`.
- `tests/testthat/test-ggplot2-rug.R` — 14 tests, including three that resolve every selector against the **rendered page**, because a selector naming nothing is the blind spot no payload assertion can see.
- Full suite: **no failures**, 99 skips (patchwork/shiny not installed here), 6268 passing before the last two tests were added.
- `man/` regenerated with the pinned roxygen2 8.1.0, per #143's drift check.

## One thing worth knowing for the next reader

Building the schema with a separate `Ggplot2PlotOrchestrator` call and then rendering the same plot gives **two different sets of grob ids** — grid's automatic names come from a global counter that `ggplotGrob()` advances. My first draft of the selector tests did exactly that and failed against ids that were correct for a gtable nobody rendered. The tests now read their layers out of the rendered page's own `maidr-data`, as `test-gantt-segment.R` does. Recorded in the test file, since the failure it produces looks like a bug in the processor and is not.

---


jooyoungseo:
Good catch, and it was an oversight rather than a decision — I wrote `layer_rows()` without knowing `LayerProcessor$get_layer_built_data()` existed. The behaviour came out right by luck, which is exactly the situation your point describes. Fixed in `d3167d3`.

## The divergence is load-bearing, and now measured

Built a `facet_grid()` with a genuinely empty cell — rows exist for `(a, x)` and `(b, y)` only, so panels 2 and 3 draw no ticks:

```
panel 1 -> layer_rows: 2   get_layer_built_data: 2
panel 2 -> layer_rows: 0   get_layer_built_data: 4
panel 3 -> layer_rows: 0   get_layer_built_data: 4
panel 4 -> layer_rows: 2   get_layer_built_data: 2
```

Exactly what you predicted: the inherited helper hands each empty panel **all four** observations — every tick drawn in the other two cells — and the panel would announce four observations it does not draw. A reader navigating that panel gets someone else's data.

## Both halves of your suggestion

I did not extend the base method with a strict flag, and it is worth saying why rather than just picking one: `get_layer_built_data()`'s fallback is the right default for the layers that use it — a series that produced no rows for a panel is usually a facet the caller did not scope, not an empty one — and a flag would put the burden of knowing which case they are in on every existing caller. The rug is the one layer so far where an empty panel is a real chart. So the comment, and a test.

- The docstring now carries the measurement above and says plainly that swapping the two is a one-line simplification that reintroduces the bug.
- `test-ggplot2-rug.R` pins it: the four panels' row counts (`2, 0, 0, 2`), and that `process()` returns `NULL` for an empty panel while a populated one yields a layer.

That also closes the facet gap you noted at the end. You were right that the repo keeps faceting tests in dedicated files — but this one is about the rug's own scoping rather than about faceting generally, so it belongs with the geom. If a `test-ggplot2-facet-*.R` sweep later wants a rug case too, that would be additional rather than a duplicate.

## Verification

- `test-ggplot2-rug.R`: 15 tests, all passing.
- Full suite: no failures.
- `man/` regenerated with the pinned roxygen2 8.1.0.

Thanks also for checking the `.lintr` 100-char limit and the `NAMESPACE`/`@keywords internal` interaction — both were things I had assumed rather than verified.

---

PR #224, opened 2026-08-22
feat(rug): give a rug the axis bounds its braille is built from

jooyoungseo:
Refs xability/maidr#1132, where this was measured. Companion to xability/py-maidr#605, which does the same for `seaborn.rugplot`.

## The gap

A point layer renders braille **only in grid mode**, and grid mode is built from `axes.{x,y}.{min,max,tickStep}`. The rug layer #223 added emitted labels alone. Measured against maidr's `ScatterTrace`:

```
axes as #223 emitted            braille { empty: true }
    x: { label: 'v' }, y: { label: 'Rug' }

with the bounds this adds       braille { values: [[2, 1, 0, 1]], min: 0, max: 2 }
    x: { label: 'v', min: 0, max: 10, tickStep: 2.5 }
    y: { label: 'Rug', min: 0, max: 1, tickStep: 1 }
```

So a rug was the one chart with **no braille surface reachable by any keystroke**. An ordinary `geom_point()` can enter grid mode and get one, because the point processor already emits these; the rug never did.

## Why it matters more here than for a scatter

`values: [[2, 1, 0, 1]]` is the observation count per cell along the axis the rug marks — two observations in the first quarter, one in the second, none in the third, one in the fourth. That is the clustering a rug exists to show.

And it is the one thing a rug's **audio cannot carry**: every tick sits at the same place on the axis pitch is mapped from, so every tick plays the same tone. Position reaches the reader through pan and text, one tick at a time; the shape of the distribution reached them nowhere.

## What is emitted

| chart | x | y |
|---|---|---|
| `aes(x = v)` | `v`, 0.6–9.4, step 2.5 | `Rug`, 0–1, step 1 |
| `aes(y = v)` | `Rug`, 0–1, step 1 | `v`, 0.6–9.4, step 2.5 |
| `aes(v, w)` — two layers | `v` bounds / `Rug` | `Rug` / `w`, 6–94, step 25 |
| `colour =` | bounds | bounds |
| `scale_x_log10()` | label only | label only |

The observation axis takes the chart's own bounds. The axis across the ticks is supplied whole as `0` to `1` in one step — a rug is one row deep by construction and no reading of the panel could say so. One row is the right shape: measured, `tickStep` 0.5 gives `[[2, 1, 0, 1], [0, 0, 0, 0]]`, a row of the surface spent saying nothing.

**Neither axis gets bounds without the other.** A grid needs both, so half of one is a surface the frontend cannot build — and bounds on the strip alone would say the rug is one row deep in a chart with no rows.

## One thing moved

`Ggplot2PointLayerProcessor$extract_axis_grid_info` reads exactly the answer the rug needs, and uses no instance state. Rather than read the grid rule a second way — which is how two readings drift, the thing xability/py-maidr#599 was about — it is lifted to `axis_grid_info()` in `R/axes_utils.R`, beside `build_axis_config()` and the other shared axis helpers.

The point processor keeps its method as a one-line delegate, so its own callers and its existing test are untouched. Its `#158` reasoning about transformed axes moves with the body and is what declines the `scale_x_log10()` case above.

## Verification

- `test-ggplot2-rug.R`: 20 tests, 5 new — the bounds on each orientation, both axes of a two-sided rug carrying their *own* bounds (asserted to differ, so a copy-paste of one into the other fails), the log decline taking the strip with it, and the ordinary reading pinned unchanged.
- `test-ggplot2-point-layer-processor.R` passes untouched, which is the check that the delegation is faithful.
- Full suite: **no failures**.
- `man/` regenerated with the pinned roxygen2 8.1.0, adding `axis_grid_info.Rd`.

## Not in scope

This does not make a rug's *audio* informative. xability/maidr#1132 lists four options for that, all of them announcement-design decisions, and none is taken here. It means the information is not unavailable, only unavailable in one modality.

---


jooyoungseo:
Facet gap closed in `001d184` — and it was worth measuring rather than waving through, because `scales = "free_x"` makes the consequence concrete.

## Per-panel bounds

Two panels whose observations are two orders of magnitude apart:

```
panel 1 -> x: min=0.9  max=3.1   tickStep=0.5
panel 2 -> x: min=90   max=310   tickStep=50
```

So a `panel_id` dropped on the way through `axis_grid_info()` would hand one panel the other's grid, and a reader would feel a surface the panel does not draw. The test asserts the two ranges do not overlap (`first$x$max < second$x$min`) and that the tick steps differ — either of which fails if the id is lost — while the strip stays one row deep in both, since that does not vary by panel.

You were right that the point processor's own tests cover `panel_id` threading in general; what they do not cover is the rug's *new* call site, and free scales is where a mistake there stops being invisible.

## The `NEWS.md` note

Correct that #223 did not add one either. I have left it alone rather than adding an entry for this PR only — a rug arriving with no entry and its braille arriving with one would read oddly, and batching both is the maintainer's call. Happy to add either or both if you would prefer.

## Verification

- `test-ggplot2-rug.R`: 21 tests, all passing.
- Full suite: no failures.
- CI is the source of truth for the R runs, as you note — Ubuntu release and oldrel, docs-drift, pkgdown and test-coverage were all green on the previous commit, and this one adds a test only.

Thanks for verifying the extraction was byte-for-byte rather than taking the description's word for it. That was the thing most likely to have drifted, and it is the check I would have wanted done.

---

Issue #225, opened 2026-08-22
`geom_polygon()` and `geom_spoke()` still cost a readable chart everything


jooyoungseo:
Found by sweeping every `Geom` ggplot2 exports against what `Ggplot2Adapter$determine_layer_type()` dispatches. Two are left, and both take the whole plot down.

## Measured

ggplot2 3.4.4, the same thirty-point scatter in every row, `save_html()`:

```
geom_point()                                interactive SVG    51,395 bytes
geom_point() + geom_polygon()               base64 image       27,001 bytes
geom_point() + geom_spoke()                 base64 image       48,249 bytes
geom_point() + annotate("rect")             interactive SVG    52,205 bytes
geom_point() + geom_hline()                 interactive SVG    52,172 bytes
```

The scatter's thirty points are readable on its own and gone the moment either layer is added — the same shape #176, #211 and #197 each measured, on the two geoms nothing has claimed yet. `annotate("rect")` and the reference lines are in the last two rows as the controls: those were fixed, and these were not.

## Everything else is covered

For the record, so this can be closed as the last of the sweep rather than one of an unknown number. Every `Geom` in ggplot2 3.4.4, against the adapter:

| outcome | geoms |
|---|---|
| **read** | Abline\*, Area, Bar, Bin2d, Boxplot, Col, Contour, ContourFilled, Crossbar, Curve, Density2d, Density2dFilled, Dotplot, Errorbar, Errorbarh, Function, Hex, Hline\*, Label\*, Line, Linerange, Path, Point, Pointrange, Raster, Rect (via `annotate`), Ribbon, Rug, Segment, Smooth, Step, Text\*, Tile, Violin, Vline\* |
| **skipped** (\* above) | Abline, Blank, Hline, Label, Text, Vline |
| **left `unknown`** | **Polygon**, **Spoke** |
| out of scope | AnnotationMap, CustomAnn, Logticks, Map, RasterAnn, Sf |

Two that look missing and are not: `GeomDensity` reads as `smooth`, and grouped by `colour=` or `fill=` it emits one `smooth` layer holding **two series**, each named by `z` — the multi-series convention, not a fold. `GeomQuantile` needs `quantreg`, which is not installed here, so it is untested rather than broken.

## Which is not the same question for the two of them

[...]

PR #226, opened 2026-08-23
feat(ggplot2): read geom_spoke through the rule geom_segment already has

jooyoungseo:
Half of #225. `geom_polygon()`, the other half, is #197's open question and is deliberately untouched.

## What was wrong

`GeomSpoke` is a direct `Geom` subclass, so it matched no dispatch branch, reached the unknown processor, and made `has_unsupported_layers()` true — which drops the **whole plot** to a static image.

Measured on ggplot2 3.4.4 with `save_html()`, the same thirty-point scatter in each row:

```
geom_point()                    interactive SVG   51,395 bytes
geom_point() + geom_spoke()     base64 image      48,249 bytes
```

The scatter's thirty points were readable on their own and gone the moment a spoke was drawn beside them.

## Why this decides nothing new

A spoke is `geom_segment()` reparameterised — an angle and a radius rather than an endpoint — and `GeomSpoke$setup_data()` turns the pair into the same `xend`/`yend` columns the segment branch already reads. ggplot2 has done the work before the adapter is asked anything:

```
built columns: angle, radius, x, y, PANEL, group, xend, yend, colour, ...
           x        y       xend     yend angle radius
1 -0.6264538 21.51178 -0.3631790 21.65561   0.5    0.3
```

So it is dispatched into the existing branch rather than given a rule of its own, and the same `segments_span_lanes()` question decides it:

```
geom_spoke(angle = 0.5, radius = 0.3)   spans lanes FALSE   -> unknown
geom_spoke(angle = 0,   radius = r)     spans lanes TRUE    -> gantt
```

An angled spoke is a vector field, not a schedule. It stays `"unknown"` exactly as the segment spelling of the same chart would, so **nothing that was declined before is claimed now** — `test_that("an angled spoke is still not claimed")` pins that. A flat one is a gantt written the other way round: a lane per `y`, running from `x` to `x + radius`.

## The reading, measured

```
flat spoke (lanes)            interactive   43,167 bytes   layers: gantt(n=3)
angled spoke                  base64 image  25,481 bytes
point + flat spoke            interactive   45,174 bytes   layers: point(n=3), gantt(n=3)
geom_segment lanes (control)  interactive   43,475 bytes   layers: gantt(n=3)
```

`test_that("a spoke reads exactly as the segment spelling of the same chart")` asserts the spoke's intervals against the **segment chart's** rather than against written-down numbers, so a later change to how a gantt is emitted moves both sides together and the test keeps asserting what it means to. The literal reading is written out once besides, in `test_that("the spans come off the radius, not off a second coordinate")` — `1 0-3`, `2 3-8`, `3 8-11` — so the file states what a reader is actually told.

Selectors are checked against the page they were built from, since a spoke's grobs are named by grid's own counter rather than after the geom.

## Verification

- `tests/testthat/test-gantt-spoke.R` → 11 assertions, all passing (new file)
- Full suite: **6322 passed, 0 failed, 0 errors**, 99 skipped (patchwork and shiny are not installed here)

Mutations, each applied alone against a restored baseline — both killed:

| mutation | result |
|---|---|
| `GeomSpoke` removed from the branch | 6 failed |
| spoke claimed as `gantt` unconditionally | 1 failed (the angled case) |

---

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,18 @@
 
 ## New Features
 
+* `geom_spoke()` is now read as the gantt chart it draws, where it draws one.
+  A spoke is `geom_segment()` reparameterised -- an angle and a radius rather
+  than an endpoint -- and ggplot2 turns the pair into the same `xend`/`yend`
+  the segment reading already uses, so the two spellings of one mark now reach
+  one rule and answer it alike. A flat spoke is a lane per `y` running from `x`
+  to `x + radius`; an angled one is a vector field rather than a schedule and
+  is still declined, exactly as the segment spelling of the same chart would
+  be. What this fixes is what a decline used to cost: an unclaimed layer makes
+  the whole plot fall back to a static image, so a thirty-point scatter with a
+  spoke layer beside it went from an interactive SVG to a base64 image and lost
+  every one of its points.
+
 * A Base R `contour()` is now read as the contour it draws, instead of falling
   back to a static image. The curves come from `grDevices::contourLines()`,
   which is the computation `contour()` itself does and takes the same defaults
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,15 @@
   spoke layer beside it went from an interactive SVG to a base64 image and lost
   every one of its points.
 
+  A gantt layer now counts its position among the layers drawing the same
+  **grob class** rather than among layers of the same geom, which is the
+  population the grob list it indexes into was gathered by. The two were the
+  same set until a spoke joined `geom_segment()` in drawing a `segments` grob:
+  a chart with one layer of each gave both `position == 1`, so they resolved to
+  the same grob and the second highlighted the first's intervals while
+  announcing its own. `geom_curve()` never collided, because it draws a `curve`
+  grob instead.
+
 * A Base R `contour()` is now read as the contour it draws, instead of falling
   back to a static image. The curves come from `grDevices::contourLines()`,
   which is the computation `contour()` itself does and takes the same defaults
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Both right, and both fixed in 740463a. The selector one is a real regression this branch introduced — thank you for catching it.

**The collision, measured.** A `geom_segment()` layer of two lanes and a `geom_spoke()` layer of three, on one plot:

```
before   gantt (2 lanes)  GRID.segments.1.1.1  GRID.segments.1.1.2
         gantt (3 lanes)  GRID.segments.1.1.1  GRID.segments.1.1.2  GRID.segments.1.1.3

after    gantt (2 lanes)  GRID.segments.1.1.1  GRID.segments.1.1.2
         gantt (3 lanes)  GRID.segments.2.1.1  GRID.segments.2.1.2  GRID.segments.2.1.3
```

Exactly your diagnosis: two layers, one grob each, `position == 1` apiece.

The fix is your suggestion, framed so the two sides cannot drift again: `segments_grob_class(geom)` states which grob a segment-family geom draws, and **both** the `collect()` gather and the position count now ask it. Counting on the same thing the list was gathered by is the invariant that was broken; naming it once is what keeps it. `test_that("a segment layer and a spoke layer address their own grobs")` pins it by asserting the two selector sets are disjoint, and restoring the geom-class count turns it red on its own.

**The ancestry** was mine to get wrong. Measured:

```
GeomSpoke  < GeomSegment < Geom < ggproto < gg
GeomCurve  < GeomSegment < Geom < ggproto < gg
GeomRug    < Geom < ggproto < gg
```

So `GeomSpoke` is a `GeomSegment` subclass, not a direct `Geom` one — and it still needed naming in the branch because the dispatch matches `class(geom)[1]` rather than asking `inherits()`, which is what the comment and the test file now say.

Full suite after both: **6323 passed, 0 failed, 0 errors**, 99 skipped. `man/Ggplot2GanttLayerProcessor.Rd` regenerated for the new method; `NAMESPACE` unchanged.

---

PR #228, opened 2026-08-23
fix(ggplot2): stop a layer that drew nothing from costing the whole chart

### LINKED ISSUE #227: A layer that drew nothing still costs the whole chart its interactivity
An unclaimed layer makes `has_unsupported_layers()` true and drops the whole plot to a base64 image. That is the right answer when the layer put a mark on the page that maidr cannot read — the reader would otherwise be told a chart is complete while something is drawn on it that nobody described.

It is the wrong answer when the layer drew **nothing at all**. There is no mark, so there is nothing the reader is missing, and the chart loses everything to protect them from an absence.

Measured with `save_html()` on ggplot2 3.4.4, thirty points in each row:

```
geom_point()                                   interactive SVG    50,406 bytes
geom_point() + geom_point(data = d[0, ])       interactive SVG    51,313 bytes
geom_point() + geom_polygon()                  base64 image       31,848 bytes
geom_point() + geom_polygon(data = d[0, ])     base64 image       27,368 bytes
geom_point() + geom_quantile()                 base64 image       27,368 bytes
```

Rows two and four are the same chart in every way a reader could tell — one layer of thirty points and one layer of nothing — and one of them is interactive because its empty layer happened to be of a *kind* the adapter names.

`geom_polygon()` on row three is the case the fallback exists for and must keep falling back: it draws a filled shape that nothing announces.

## The case this actually turns up in

`geom_quantile()` on row five was not contrived. It is what happens when a **Suggests package is missing**:

```
Warning: Computation failed in `stat_quantile()`
Caused by error in `compute_group()`:
! The package "quantreg" is required for `stat_quantile()`
```

`ggplot_build()` reports `30, 0` rows. ggplot2 warns, draws the rest of the chart and carries on. r-maidr turns the entire figure into a picture, with no second warning saying so and nothing connecting the two. The user sees a chart that looks right and is silently no longer accessible.

Any stat that filters everything out lands the same way, as does an unclaimed layer whose `data =` frame is empty in one facet arrangement and not another.

## Suggested shape

Where the adapter's answer would be `"unknown"`, ask whether the layer drew anything first, and return `"skip"` when it did not. `ggplot_build()`'s data frame for that layer is what knows — zero rows is zero marks — and `segments_span_lanes()` already reaches it the same way, through `find_layer_index()`.

This decides nothing new about which geoms are readable. `geom_polygon()` with data in it stays `"unknown"` and keeps costing its chart exactly what it costs today; only a layer with nothing in it stops doing so.

The "everything skipped" trap #176 named is already guarded and stays guarded: `has_unsupported_layers()` also returns true when *every* layer is `"skip"`, so a plot made only of empty unclaimed layers still falls back to an image rather than announcing itself as an interactive chart with nothing in it.

## Not in scope

`geom_quantile()` deserves a reading of its own, separately: `GeomQuantile` is a `GeomPath` subclass, so `class(geom)[1]` dispatch misses it exactly the way it missed `GeomFunction` (#202) and `GeomSpoke` (#225), and a quantile-regression fit is a `smooth` rather than a `line` for the reason `StatFunction` is. That wants `quantreg` installed to verify the reading end to end, so it is worth its own issue rather than being folded in here — and it is *not* what makes row five fall back today, since the layer draws nothing at all without that package.

jooyoungseo:
Closes #227.

An unclaimed layer makes `has_unsupported_layers()` true and drops the plot to a base64 image. That is right when the layer put a mark on the page nothing describes — a reader told the chart is complete would be told wrong. It is not right when the layer drew **nothing at all**: there is no mark, so there is nothing the reader is missing, and the chart pays everything to protect them from an absence.

Measured with `save_html()` on ggplot2 3.4.4, thirty points in each row:

| chart | before | after |
|---|---|---|
| `geom_point()` | interactive, 50,406 B | unchanged |
| `geom_point() + geom_point(data = d[0, ])` | interactive, 51,313 B | unchanged |
| `geom_point() + geom_polygon(data = d[0, ])` | **base64 image, 27,368 B** | **interactive** |
| `geom_point() + geom_quantile()` | **base64 image, 27,368 B** | **interactive** |
| `geom_point() + geom_polygon()` | base64 image, 31,848 B | unchanged |

Rows two and three were the same chart in every way a reader could tell — thirty points and a layer of nothing — and only one of them was accessible, because its empty layer happened to be of a *kind* the adapter names. Row five is the case the fallback exists for, and it keeps falling back.

## The case this turns up in

Row four was not contrived. It is what a missing **Suggests** package does:

```
Warning: Computation failed in `stat_quantile()`
Caused by error in `compute_group()`:
! The package "quantreg" is required for `stat_quantile()`
```

`ggplot_build()` reports `30, 0` rows. ggplot2 warns, draws the rest and carries on; r-maidr turned the whole figure into a picture, with no second warning saying the chart had stopped being accessible and nothing connecting the two.

## What changed

Both places the adapter answers `"unknown"` — the fall-through and the `geom_segment()` decline (#194) — now route through one `unread_layer_type()` that asks the built data first and returns `"skip"` when the layer drew no rows. Routing both matters: fixing only the fall-through would leave an empty `geom_segment()` layer still costing its chart everything, and the two declines mean the same thing.

Emptiness is read off `ggplot_build()` for the reason `segments_span_lanes()` reads it there: a mapping expression cannot say how many rows survived a stat, and by build time ggplot2 has resolved it.

## What did not change

- A `geom_polygon()` **with data in it** is still `"unknown"` and still costs its chart exactly what it cost before. This decides nothing new about which geoms are readable.
- A plot made **only** of empty unclaimed layers still falls back, because `has_unsupported_layers()` is also true when every layer is `"skip"` — the trap #176 named, that "nothing unsupported" must not quietly come to mean "nothing at all".
- A plot that cannot be built at all declines rather than skipping: a build that raised has said nothing about what any layer drew.

## Not in scope

`geom_quantile()` deserves a reading of its own. `GeomQuantile` is a `GeomPath` subclass, so `class(geom)[1]` dispatch misses it the way it missed `GeomFunction` (#202) and `GeomSpoke` (#225), and a quantile-regression fit is a `smooth` rather than a `line` for the reason `StatFunction` is. Verifying that reading end to end needs **quantreg** installed, so it belongs in its own issue — and it is not what makes the chart fall back today, since without that package the layer draws nothing at all.

## Verification

`testthat::test_dir("tests/testthat")` — **6337 passed, 0 failed, 0 errors**, up from 6323. `man/Ggplot2Adapter.Rd` regenerated with roxygen2.

14 new assertions. Mutation sweep, each change applied alone against a restored baseline:

| mutation | outcome |
|---|---|
| never skip an empty layer | killed (6) |
| always skip, drawn or not | killed (3) |
| segment branch declines plainly again | killed (1) |
| a build failure skips | killed (1) |
| fall-through declines plainly again | killed (5) |
| a layer not in this plot skips | killed (1) |
| `isTRUE(...)` → `!isFALSE(...)` | **survived** |

The survivor is reported rather than papered over. The two spellings differ only when `nrow()` returns `NULL`, which an in-range index into a built plot's data cannot produce — so it is a defensive *spelling*, not a branch, and the comment beside it says exactly that. The first sweep also had a `!is.data.frame(rows)` branch guarding the same thing; that one *was* an unreachable branch, and it was removed rather than left as untestable code with a comment claiming a distinction no chart can reach.

---

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,29 @@
 # maidr (development version)
 
+## Bug Fixes
+
+* A layer that drew **nothing** no longer costs the whole chart its
+  interactivity. An unclaimed layer makes the plot fall back to a static
+  image, which is right when the layer put a mark on the page that nothing
+  describes -- a reader told the chart was complete would be told wrong -- and
+  wrong when it drew nothing at all, because then there is no mark and the
+  chart pays everything to protect the reader from an absence. Measured on
+  thirty points, `geom_point() + geom_polygon(data = frame[0, ])` was a 27,368
+  byte base64 image where `geom_point() + geom_point(data = frame[0, ])` was a
+  51,313 byte interactive SVG: the same chart in every way a reader could
+  tell, and only one of them accessible, because one empty layer happened to
+  be of a *kind* the adapter names.
+
+  The case this turns up in is a missing **Suggests** package.
+  `geom_quantile()` without **quantreg** warns, computes no rows and draws
+  nothing; ggplot2 carries on and the figure silently stopped being
+  accessible, with no second warning connecting the two.
+
+  Nothing here changes which geoms are readable. A `geom_polygon()` with data
+  in it is still declined and still costs its chart exactly what it cost
+  before, and a plot made only of empty unclaimed layers still falls back
+  rather than announcing itself as an interactive chart with nothing in it.
+
 ## New Features
 
 * `geom_spoke()` is now read as the gantt chart it draws, where it draws one.
--- a/R/ggplot2_adapter.R
+++ b/R/ggplot2_adapter.R
@@ -169,7 +169,10 @@ Ggplot2Adapter <- R6::R6Class(
       # written the other way round, a lane per y running from x to x + r,
       # and was costing its chart every bit of interactivity (#225).
       if (geom_class %in% c("GeomSegment", "GeomCurve", "GeomSpoke")) {
-        return(if (self$segments_span_lanes(layer, plot_object)) "gantt" else "unknown")
+        if (self$segments_span_lanes(layer, plot_object)) {
+          return("gantt")
+        }
+        return(self$unread_layer_type(layer, plot_object))
       }
 
       if (geom_class %in% c("GeomLine", "GeomPath", "GeomMA")) {
@@ -378,7 +381,7 @@ Ggplot2Adapter <- R6::R6Class(
         return("skip")
       }
 
-      "unknown"
+      self$unread_layer_type(layer, plot_object)
     },
 
     #' @description Check if a bar layer is drawn as pie wedges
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Thanks — the performance nit is real and worth putting a number on rather than leaving as a judgement, so I measured it.

```
                                       ggplot_build     save_html
30 points + 30 segments                    28.6 ms       616.3 ms
5000 points + 5000 segments                36.6 ms       774.7 ms
```

So the extra build is **~4.7% of a full render** at 5000 rows, and it barely scales — 167× the data moved it 28.6 → 36.6 ms, so it is almost entirely fixed cost. It is also confined to one branch: it happens only when `segments_span_lanes()` returns `FALSE`, and of the two ways that happens, the *drawn* non-lane segment still returns `"unknown"` and the chart renders as a static image, which is the cheaper path anyway. The interactive render pays it only for a segment layer that drew nothing.

I am leaving it, deliberately, and the reason is the one the review itself points at: `ribbon_is_area()`, `segments_span_lanes()`, `draws_single_ring()` and now `unread_layer_type()` each build independently. Threading a build through one of the four would leave the pattern half-changed and put a cache — with no invalidation story for a value type that callers can rebuild at will — in the middle of a classifier. If `ggplot_build` cost is ever profiled as a real bottleneck, the fix wants to be one build per `detect_layer_type()` call shared by every branch, which is a change to the file rather than to this branch.

Two other points worth confirming rather than leaving implicit:

**The inverted `quantreg` skip is deliberate**, and the review read it right. `skip_if(requireNamespace("quantreg", quietly = TRUE))` means the test runs precisely where the bug reproduces and skips where it cannot, so it needs no `Suggests` entry and does not silently pass by depending on quantreg's absence going unnoticed. It is documented as such in the test's own comment.

**On the "6337 passed" figure being taken on my word** — that is fair, and I would rather it were checkable. For the record it was `testthat::test_dir("tests/testthat", reporter = "silent")` on this branch, 0 failed / 0 errors / 99 skipped, up from 6323 on `main`; the mutation sweep is reproducible by applying each row of the PR table alone to `R/ggplot2_adapter.R` and re-running `tests/testthat/test-empty-unknown-layer.R`. CI runs the same suite, so its result is the independent check.

---

jooyoungseo:
CI was red on both ubuntu jobs. Diagnosed and fixed in `2eb2719` — the cause was mine, and it is worth writing down because the failure was invisible from the test output.

The suite reported `[ FAIL 0 | WARN 31 | SKIP 5 | PASS 7294 ]`, so the tests were not the problem. Pulling the `00check.log` out of the run's artifact gave the real answer:

```
  'library' or 'require' call not declared from: 'quantreg'
* checking tests ... [317s/315s] OK
Status: 1 WARNING, 1 NOTE
```

`check-r-package@v2` errors on warnings, so one line failed the whole job while every test passed. `R CMD check` reads my `requireNamespace("quantreg", quietly = TRUE)` guard as the tests depending on a package `DESCRIPTION` does not declare. (The NOTE is pre-existing — three `checkRd` items in `Ggplot2BarLayerProcessor.Rd`, `Ggplot2StepLayerProcessor.Rd` and `LayerProcessor.Rd`, none of them files this branch touches.)

Adding `quantreg` to `Suggests` would have been the wrong fix, and this is the part worth recording: the review above read the inverted skip as deliberate and it was — but declaring the package defeats it. quantreg would then be installed in CI, `stat_quantile()` would compute rows, and the guard would skip the test on the only machine that runs it. The case would never arise anywhere.

So the layer now gets a stat of its own that returns no rows:

```r
nothing_at_all <- ggplot2::ggproto(
  "StatNothingAtAll", ggplot2::Stat,
  compute_group = function(data, scales) data[0, , drop = FALSE]
)
```

Measured: 20 rows for the points, 0 for that layer, type `"skip"`, chart interactive. That is the same state the classifier sees, reached by *computation* rather than by an empty input — which is the whole difference between this test and the empty-`data =` one beside it, and the reason it is not a near-duplicate. The quantreg measurement stays in the comment, where it says what the case looks like in the wild without making the package a dependency of anything.

Re-verified after the change: suite **0 failed / 6337 passed**, mutation sweep unchanged (all six killed), and `R CMD check` now reports `checking for unstated dependencies in 'tests' ... OK`.

---

Issue #230, opened 2026-08-23
`stat_function(geom = "point")` stops the render: a stat-keyed `smooth` type the smooth processor cannot resolve


jooyoungseo:
Found while scoping #229, and reported rather than folded into it.

```r
ggplot(d, aes(x, y)) + xlim(0, 5) + stat_function(fun = sin, geom = "point")
```

```
save_html(p, f)
#> Error: No smooth curve layers found in plot
```

Not a fallback to a static image — an **error out of `save_html()`**. The user's script stops. `stat_function(fun = sin)` with its default `geom = "line"` renders interactively, so which geom the author passed decides whether the call returns at all.

## Why

`detect_layer_type()` types the layer `smooth` on the **stat**:

```r
if (geom_class == "GeomFunction" || stat_class == "StatFunction") {
  return("smooth")
}
```

`Ggplot2SmoothLayerProcessor$resolve_target_layer()` then decides what to read on the **geom**:

```r
is_smooth_like <- inherits(own_layer$geom, "GeomSmooth") ||
  inherits(own_layer$geom, "GeomLine") ||
  inherits(own_layer$geom, "GeomDensity") ||
  inherits(own_layer$geom, "GeomFunction") ||
  inherits(own_layer$geom, "GeomArea")
```

`GeomPoint` is in neither list, so the layer's own index is rejected, the fallback search over the whole plot finds no smooth-like layer, and it `stop()`s.

The two lists answer the same question — *is this a computed curve?* — from different sides, and a stat-keyed type can name a geom the geom-keyed list does not. `StatDensity` beside `GeomSmooth` has the same shape: `stat_density(geom = "point")` is the equivalent spelling.

## Why #229 does not add to it

#229 routes `GeomQuantile` through the same `smooth` branch. The symmetric change would have been `stat_class == "StatQuantile"` beside `StatDensity`, which would have made `stat_quantile(geom = "point")` a second instance of exactly this crash. It is keyed on the geom alone for that reason, with the trade recorded in the code.

## Shape of a fix

Two candidates, and the choice is a real one:

1. **Widen the processor** to accept any geom a `smooth` type can arrive with. Keeps every chart readable, but a `GeomPoint` layer read as a curve announces fitted points as a series — which may be the right reading, since they *are* fitted values, but it needs deciding rather than falling out.
[...]

PR #231, opened 2026-08-23
fix(ggplot2): read geom_quantile as the fitted curve it draws

### LINKED ISSUE #229: `geom_quantile()` is a `GeomPath` subclass the dispatch misses, and it costs its chart everything
`detect_layer_type()` matches `class(geom)[1]`, and the chain is:

```
GeomQuantile < GeomPath < Geom < ggproto < gg
```

`GeomPath` reads as `line`; the subclass matches no branch, reaches the unknown processor, and takes the whole plot down. This is exactly the shape `GeomFunction` had in #202 and `GeomSpoke` had in #225 — a readable geom declined for being a subclass of one that is read.

Measured on ggplot2 3.4.4, thirty points with a `GeomQuantile` layer that draws:

```
geom_point()                             interactive SVG    50,409 bytes
geom_point() + a GeomQuantile layer      base64 image       44,724 bytes
geom_point() + geom_smooth(se = FALSE)   interactive SVG    57,823 bytes
```

The scatter reads on its own, and reads as nothing the moment a quantile fit is drawn beside it — while the *mean* fit beside the same points is read fine.

## `smooth`, not `line`

A quantile regression line is a **model fit over** the data, not a series **of** it, which is the distinction `StatFunction` and `StatDensity` already settle in that branch:

```r
if (geom_class == "GeomFunction" || stat_class == "StatFunction") {
  return("smooth")
}
if (geom_class == "GeomSmooth" || stat_class == "StatDensity") {
  return("smooth")
}
```

`stat_quantile()` fits `rq`/`rqss` and evaluates it at renderer-chosen positions, exactly as `stat_smooth()` does for the conditional mean. Reading it as `line` would announce a fitted curve as observations — the same wrong-question failure xability/maidr#814 catalogues, where "the parent trace reads the data correctly and answers the wrong question."

So this decides nothing new: it routes a second spelling of "a fitted curve" through the rule the first already has.

## Verifying it without quantreg

Worth recording, because it is what makes this testable in CI at all.

`geom_quantile()` needs the **quantreg** package, which is a `Suggests` of ggplot2 and not declared by this package. Naming it in a test — even as `requireNamespace("quantreg", quietly = TRUE)` inside a `skip_if` — makes `R CMD check` report `'library' or 'require' call not declared from: 'quantreg'`, and `check-r-package` errors on warnings. That is the failure #228 hit and fixed by dropping the reference.

Both halves are reachable without it:

- **The dispatch** needs no data. `detect_layer_type()` reads `class(layer$geom)[1]`, so `ggplot2::geom_quantile()` can be built and classified whether or not the stat can run.
- **The reading** needs a layer that draws. `layer(geom = GeomQuantile, stat = "identity", ...)` is a real `GeomQuantile` layer over data the test supplies — measured above at 30 built rows and a base64 fallback.

Neither mentions quantreg, so neither trips the check.

## Related

- #202 — `GeomFunction`, the same miss, same branch.
- #225 — `GeomSpoke`, the same miss, routed through the segment rule.
- #227 — why a layer that draws *nothing* no longer costs the chart anything. `geom_quantile()` without quantreg computes no rows, so on `main` today it is already harmless; this issue is about the case where the fit **does** draw.

jooyoungseo:
Closes #229.

`detect_layer_type()` matches `class(geom)[1]`, and the chain is:

```
GeomQuantile < GeomPath < Geom
```

`GeomPath` reads as `line`; the subclass matched no branch, reached the unknown processor and took the whole plot down — the third geom missed this way, after `GeomFunction` (#202) and `GeomSpoke` (#225).

Measured with `save_html()` on thirty points, with a quantile layer that draws:

```
geom_point()                             interactive SVG   50,409 bytes
geom_point() + a GeomQuantile layer      base64 image      44,724 bytes
geom_point() + geom_smooth(se = FALSE)   interactive SVG   57,823 bytes
```

The scatter read on its own and read as nothing the moment a quantile fit was drawn beside it, while the *mean* fit beside the same points read fine.

## `smooth`, not `line`

For the reason `StatFunction` is: `stat_quantile()` fits `rq`/`rqss` and evaluates it at renderer-chosen positions exactly as `stat_smooth()` does for the conditional mean, so the curve is a model **over** the data rather than a series **of** it. Reading it as a line would announce a fit as observations.

So this decides nothing new — it routes a second spelling of "a fitted curve" through the rule the first already has. `geom_path()` keeps its own reading, pinned.

## Two things it deliberately does not do

**`StatQuantile` is not added beside `StatDensity`.** A stat-keyed `smooth` type can name a geom the smooth processor's own list does not, and the render then **stops outright** rather than declining. Measured on `main`:

```r
ggplot(d, aes(x, y)) + xlim(0, 5) + stat_function(fun = sin, geom = "point")
#> Error: No smooth curve layers found in plot
```

Matching the symmetry would have made `stat_quantile(geom = "point")` a second instance of it. Reported as #230 and pinned here by a test asserting `stat_quantile(geom = "point")` still reads as `point`, so the symmetric addition is a decision rather than a reflex.

**`GeomQuantile` is not added to that processor's fallback search.** It only runs when a layer's *own* geom was rejected, which a quantile layer's never is — so it would reach nothing except #230's case, and answering that by pairing a layer with someone else's curve would announce one fit as another's, which is worse than the error. The two lists now differ on purpose, with the reason written where they sit.

## Testing without quantreg

`geom_quantile()` needs **quantreg**, which this package does not declare, and naming it — even inside a `skip_if` — reintroduces the `'library' or 'require' call not declared from: 'quantreg'` warning that turned CI red on #228. Nothing in the new file mentions it, and `checking for unstated dependencies in 'tests'` reports OK.

Both halves are reachable anyway:

- **the dispatch** reads `class(layer$geom)[1]` and needs no data, so `geom_quantile()` itself is the call under test;
- **the reading** needs a layer that draws, which `layer(geom = GeomQuantile, stat = "identity", ...)` gives over data the test supplies.

## Verification

Suite: **0 failed, 6348 passed** (up from 6337). `R CMD check`'s unstated-dependency and code checks report OK; no `man/` drift.

11 new assertions, including the #226-shaped hazard — a chart holding both a mean fit and a quantile fit, asserting their selectors are disjoint and their data differ, since both draw `polyline` grobs and that was the exact population that collided there.

Mutation sweep, each applied alone against a restored baseline:

| mutation | outcome |
|---|---|
| drop `GeomQuantile` from the dispatch branch | killed (7) |
| drop it from the processor's own-layer list | killed (2) |
| add `StatQuantile` to the dispatch | killed (1) |
| drop it from the processor's fallback list | **survived** |

The survivor is why the fallback entry is not in this PR. It was in the first draft, and removing it changed nothing — the own-layer check returns early for every quantile layer, so the entry was unreachable except in #230's case, which this deliberately leaves alone. Rather than keep dead code, it is gone and the two lists' difference is documented.

---

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,12 @@
   made `stat_quantile(geom = "point")` a second instance of an existing
   crash, which is reported separately rather than matched.
 
+  A quantile layer that drew *nothing* keeps the answer it had before:
+  skipped, not claimed. `geom_quantile()` without **quantreg** computes no
+  rows, and it is the case that rule was written for, so claiming it on the
+  geom regardless would have put an empty layer in the schema for a chart
+  that had none.
+
 * A layer that drew **nothing** no longer costs the whole chart its
   interactivity. An unclaimed layer makes the plot fall back to a static
   image, which is right when the layer put a mark on the page that nothing
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,25 @@
 
 ## Bug Fixes
 
+* `geom_quantile()` is now read as the fitted curve it draws, instead of
+  costing its chart every bit of interactivity. `GeomQuantile` is a
+  `GeomPath` subclass and dispatch matches the first class name, so it
+  matched no branch and reached the unknown processor -- the third geom
+  missed this way, after `geom_function()` and `geom_spoke()`. Measured on
+  thirty points, a scatter went from a 50,409 byte interactive SVG to a
+  44,724 byte base64 image the moment a quantile fit was drawn beside it,
+  while the *mean* fit beside the same points read fine.
+
+  It reads as a `smooth` rather than a `line` for the reason `stat_function()`
+  does: `stat_quantile()` fits `rq`/`rqss` and evaluates it at
+  renderer-chosen positions exactly as `stat_smooth()` does for the
+  conditional mean, so the curve is a model over the data rather than a
+  series of it.
+
+  Keyed on the geom alone. The symmetric `StatQuantile` addition would have
+  made `stat_quantile(geom = "point")` a second instance of an existing
+  crash, which is reported separately rather than matched.
+
 * A layer that drew **nothing** no longer costs the whole chart its
   interactivity. An unclaimed layer makes the plot fall back to a static
   image, which is right when the layer put a mark on the page that nothing
--- a/R/ggplot2_adapter.R
+++ b/R/ggplot2_adapter.R
@@ -188,7 +188,33 @@ Ggplot2Adapter <- R6::R6Class(
       if (geom_class == "GeomFunction" || stat_class == "StatFunction") {
         return("smooth")
       }
-      if (geom_class == "GeomSmooth" || stat_class == "StatDensity") {
+      # `GeomQuantile` is a `GeomPath` too, so it reached the line branch's
+      # name check, matched nothing, and took its chart down with it -- the
+      # third geom to be missed for being a subclass of a read one, after
+      # `GeomFunction` (#202) and `GeomSpoke` (#225). Measured on thirty
+      # points with a quantile layer that draws:
+      #
+      #     geom_point()                            interactive   50,409 bytes
+      #     geom_point() + a GeomQuantile layer     base64 image  44,724 bytes
+      #     geom_point() + geom_smooth(se = FALSE)  interactive   57,823 bytes
+      #
+      # `smooth` rather than `line` for the reason `StatFunction` is:
+      # `stat_quantile()` fits `rq`/`rqss` and evaluates it at
+      # renderer-chosen positions, exactly as `stat_smooth()` does for the
+      # conditional mean, so the curve is a model over the data rather than
+      # a series of it. Reading it as a line would announce a fit as
+      # observations (#229).
+      # Keyed on the geom alone, deliberately. `StatQuantile` would have been
+      # the symmetric addition beside `StatDensity`, and it is a trap: a stat
+      # check claims `stat_quantile(geom = "point")`, whose `GeomPoint` the
+      # smooth processor does not recognise, so `resolve_target_layer()`
+      # falls through and stops the render outright. Measured on `main`, the
+      # `StatFunction` check above already does this --
+      # `stat_function(fun = sin, geom = "point")` raises "No smooth curve
+      # layers found in plot" out of `save_html()` (#230). Reported rather
+      # than matched.
+      if (geom_class %in% c("GeomSmooth", "GeomQuantile") ||
+            stat_class == "StatDensity") {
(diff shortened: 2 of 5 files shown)


jooyoungseo:
Good catch — that is a real regression this PR introduced, and it is fixed in `d8a4ae9`.

Reproduced exactly as described:

```
point + geom_quantile() [no quantreg]   point(10) smooth(1)
```

A `smooth` whose `data` has length **1** — one series, holding nothing. Worse than a missing layer, because the reader is offered a series to walk into.

The framing is worth sharpening in one respect, because it changes where the fix belongs. The ghost is not new to `GeomQuantile`; it is what **every** recognised type already does:

```
geom_point() + geom_smooth(data = d[0, ], se = FALSE)   point(10) smooth(1)
geom_point() + geom_point(data = d[0, ])                point(10) point(0)
```

Both measured on `main`, untouched by this PR. What #231 did was take `geom_quantile()` — the one chart named in `unread_layer_type()`'s own docs as the case #227 was found through — *out* of the rule and into that pre-existing gap. So the regression is real and specific, while the underlying gap is general.

**The fix.** The emptiness question is now its own method, `layer_drew_nothing()`, rather than a step inside `unread_layer_type()` — because two branches ask it, and a rule two callers ask is a rule rather than a fall-through. The quantile branch consults it before claiming:

```r
if (geom_class == "GeomQuantile" &&
      self$layer_drew_nothing(layer, plot_object)) {
  return("skip")
}
return("smooth")
```

```
point + geom_quantile() [no quantreg]   point(10)
point + EMPTY point                     point(10) point(0)     <- unchanged
point + EMPTY smooth                    point(10) smooth(1)    <- unchanged
```

**Why only `GeomQuantile`.** Asking it for every layer is the right general answer, and it costs one `ggplot_build()` per layer on every chart — measured at ≈ 37 ms against a ≈ 775 ms render, so about 5% each, where today the build only happens on the decline path. That trade wants deciding on its own, so it is filed as #232 with the measurements and two candidate shapes (thread one built plot through the classification pass, or drop empty layers in the orchestrator, which already holds the build).

**One test changed its premise rather than its expectation.** `"a quantile fit is classified as the smooth it is"` asked `geom_quantile()` directly for `"smooth"` — on a machine where the stat cannot run, which is now correctly `"skip"`. It now asserts the property that does not depend on quantreg being installed: the answer is never `"unknown"`, which is the one that costs the chart. The drawn case is pinned exactly, beside it.

Two new mutations, both killed: removing the guard (2 failures), and making `layer_drew_nothing()` always answer `FALSE` (2 failures).

Suite: **0 failed, 6351 passed**.

On the minor point — yes, `GeomQuantile`'s absence from the fallback `smooth_layers` search is deliberate, and it now says so where it sits.

---

PR #233, opened 2026-08-23
fix(ggplot2): stop a stat-claimed layer from raising when its geom is unreadable

### LINKED ISSUE #230: `stat_function(geom = "point")` stops the render: a stat-keyed `smooth` type the smooth processor cannot resolve

jooyoungseo:
Closes #230. Reported from #231, where a `StatQuantile` check would have been a second instance of this crash.

`detect_layer_type()` claims a layer as `"smooth"` partly on its **stat** — `StatFunction` and `StatDensity` both do it — and a stat can name a **geom** the smooth processor cannot read. `resolve_target_layer()` then rejected the layer's own index, found nothing in its fallback search, and `stop()`ped:

```r
stat_function(fun = sin, geom = "point")   # Error: No smooth curve layers found in plot
stat_function(fun = sin, geom = "step")    # Error: No smooth curve layers found in plot
stat_function(fun = sin)                   # interactive
```

Not a fallback to a picture — an **error out of `save_html()`**, so the caller's script stopped. Which geom the author happened to pass decided whether the call returned at all.

A decline is a reading decision; an exception is a broken call.

## What changed

The list the processor works from was written inline in `resolve_target_layer()`, and the classifier had no way to ask it. It is now stated once as **`smooth_reads_geom()`**, and both sides consult it, so the two cannot disagree.

Measured across every geom `stat_function()` accepts:

| geom | before | after |
|---|---|---|
| `GeomFunction` | `smooth` interactive | `smooth` interactive |
| `GeomLine` | `line` interactive | `line` interactive |
| `GeomPath` | `line` interactive | `line` interactive |
| `GeomPoint` | `smooth` **RAISED** | `point` interactive |
| `GeomStep` | `smooth` **RAISED** | `unknown` base64 image |
| `GeomArea` | `smooth` interactive | `smooth` interactive |
| `GeomSmooth` | `smooth` interactive | `smooth` interactive |

The point row falls through to the branch for the geom it was actually drawn with and reads as the scatter on the page. That loses "this is a fit, not observations" for that spelling — worth less than a chart that renders, and the alternative is teaching the processor to read a `GeomPoint`, which is a different change.

The step row reaches the unknown processor and the chart falls back to an image. That is not a regression introduced here: it is the step branch's own documented answer for a step drawn on some other computed stat ("a step layer drawn on some other computed stat keeps returning unknown"). This restores it from a crash.

## Why `StatQuantile` is still absent

#231 left it out because a stat check would have made `stat_quantile(geom = "point")` a second instance of exactly this crash, and reported it here rather than matching it. With the guard in place it is no longer *needed*: what a stat check buys is the spellings where the geom says nothing, and those are now turned away when the processor cannot read them.

## Verification

`smooth_reads_geom()` uses `inherits()`, not `class(geom)[1]` — deliberately, and unlike the dispatch. The dispatch asks *what did the author draw*; this asks *can the processor read this artist*, which a subclass of a readable geom can. That asymmetry is documented at the function.

Full suite: **6378 passed, 0 failed, 0 errors, 99 skipped**. `roxygenise()` adds only `man/smooth_reads_geom.Rd`.

27 new tests in `test-smooth-geom-agreement.R`, asserting the agreement itself rather than a list of geoms: for every geom, if the classifier says `"smooth"` then `smooth_reads_geom()` is TRUE, and every spelling renders without raising.

Mutation sweep, each applied alone against a restored baseline:

| mutation | outcome |
|---|---|
| `S1` drop the guard from the `StatFunction` branch | killed (6 failed, 2 error) |
| `S2` drop the guard from the `StatDensity` branch | killed (1 failed) |
| `S3` `smooth_reads_geom` returns TRUE for everything | killed (8 failed) |
| `S4` `smooth_reads_geom` returns FALSE for everything | killed (3 failed) |
| `S5` processor keeps its own inline list | killed (1 error) |

`S5` is the one that matters: it is the drift this change exists to prevent, and it now fails.

---


jooyoungseo:
Both findings taken, in `e5b6736`. The first turned out to have a stronger answer than "derive the list".

**1 — the third, partial list.** You're right that it's a third copy and right that it's currently unreachable for `GeomQuantile`/`GeomArea`. Measured how unreachable, with a `cat()` at the top of the search on the full suite (**6378 passed**):

```
FALLBACK own=GeomBar all=GeomBar
FALLBACK own=GeomBar all=GeomBar
FALLBACK own=GeomBar all=GeomBar
```

Three entries, all three from `test-ggplot2-smooth-layer-processor.R:241`, which constructs the processor by hand against `create_test_ggplot_bar()` specifically to assert the `stop()`. **No render reaches it at all** — and this PR is what made that true: a processor is only built for a layer the classifier typed, and the classifier now types `"smooth"` only when `smooth_reads_geom()` accepts the geom, so the early return always fires first.

That changes the fix. Deriving the fallback from `smooth_reads_geom()` would only swap one unreachable branch for another, and it would newly claim `GeomQuantile` and `GeomArea` layers as *fallback targets* — which is what the previous round removed as unreachable. The real options are delete-and-keep-the-error, or leave it; filed as #234 with the measurement.

What I did here is replace the comment, which this PR had made stale — it justified the divergence by "except the case #230 describes, where the render stops," and #230 is what this PR fixes. It now states what was measured, and the constraint that matters more than the list: the search must never *succeed* for a layer the processor could not read, because pairing one layer's data with a sibling's curve announces one fit as another's, which is worse than the error.

**2 — double render.** Both tests now bind the first render and reuse it. `expect_no_error(html <- ...)` keeps the no-error claim and the value.

Full suite after both: **6378 passed, 0 failed, 0 errors**. `roxygenise()` clean — the comment is a plain `#` inside the R6 method, so `man/` is untouched.

On "not verified": CI ran the four-platform `R CMD check` green, including `ubuntu-latest (oldrel-1)` and the "Generated docs match their sources" drift check, so the numbers hold on machines that aren't mine.

---

PR #236, opened 2026-08-23
fix(ggplot2): stop a recognised layer that drew nothing from reaching the schema

### LINKED ISSUE #232: A layer of a *recognised* type that drew nothing still becomes an empty layer in the schema
#227 stopped an **unclaimed** layer that drew nothing from costing the chart its interactivity: `detect_layer_type()` now asks `layer_drew_nothing()` before answering `"unknown"`, and returns `"skip"` instead.

A layer of a **recognised** type gets no such question. It is typed on its geom, a processor is built for it, and it reaches the schema holding nothing.

Measured on ggplot2 3.4.4, ten points, `n` being the length of that layer's `data`:

```
geom_point() + geom_point(data = d[0, ])                     point(10) point(0)
geom_point() + geom_smooth(data = d[0, ], se = FALSE)        point(10) smooth(1)
geom_point() + geom_polygon(data = d[0, ])                   point(10)            <- #227
```

The third is the one #227 fixed. The first two are the same absence read two other ways, and neither is excluded.

The `smooth` case is the worse of the two: `data` has length **1**, so the reader is offered a *series* to walk into, and the series is empty. `point(0)` is a layer with no points in it. Both are the shape #421 named on the Python side — "a layer to walk into and find nothing in".

## Where it turns up

The same places #227 lists: a `data =` filtered to nothing, a stat that dropped every row, a facet arrangement in which one layer's data is empty, or a **Suggests** package that is absent so the stat could not run.

## Shape of a fix

`layer_drew_nothing()` already exists as its own method, exactly so a second caller can ask it — #231 uses it to keep `geom_quantile()` from claiming a curve that was never drawn. The general form is to ask it once for every layer, before the type is answered at all.

The cost is why #231 did not do that. `layer_drew_nothing()` calls `ggplot_build()`, and asking it at the top of `detect_layer_type()` makes **every** chart pay one build per layer. Measured on a 5,000-point two-layer chart: `ggplot_build()` ≈ 37 ms against a `save_html()` of ≈ 775 ms, so roughly 5% per layer. Today the build only happens on the decline path and on the one geom #231 added.

So this wants either:

1. **A shared build per classification pass.** The adapter calls `ggplot_build()` independently in `ribbon_is_area()`, `segments_span_lanes()`, `draws_single_ring()` and `layer_drew_nothing()`; at most one runs per layer today, but a top-level emptiness check would make it two. Threading one built plot through the pass removes the cost objection and is worth doing on its own terms.
2. **Or the orchestrator**, which already has the built plot in hand, dropping layers whose built data is empty before a processor is made for them. That keeps `detect_layer_type()` a pure question about kind, which is what it reads as.

Option 2 is probably the smaller change and the better separation — emptiness is not a fact about what kind of chart a layer is.

## Related

- #227 — the same rule, for unclaimed layers.
- #231 — `geom_quantile()`, and the review that found it had taken #227's own documented case back off that rule.
- xability/py-maidr#421 — the same reading defect from the Python side.

jooyoungseo:
Closes #232. The general case of #227, which handled only *unclaimed* layers.

A layer of a **recognised** type got no emptiness question: it was typed on its geom, a processor was built for it, and a reader was handed a layer to walk into with nothing in it. Measured on ten points, the second layer drawn from `d[0, ]`:

| second layer | before | after |
|---|---|---|
| `geom_point(data = d[0, ])` | `point(10) point(0)` | `point(10)` |
| `geom_col(data = d[0, ])` | `point(10) bar(0)` | `point(10)` |
| `geom_line(data = d[0, ])` | `point(10) line(1x0)` | `point(10)` |
| `geom_smooth(data = d[0, ])` | `point(10) smooth(1x0)` | `point(10)` |

`1x0` is one *series* holding zero points, and those two are the worse cases — a reader is offered something to walk into. The issue named the point and smooth spellings; **the line and bar ones turned up on the way** and are the same absence read two more ways.

A plot whose only layer is empty now falls back to an image.

## Where it is asked, and why that is the whole design

`layer_drew_nothing()` calls `ggplot_build()`. Asking it at the top of `detect_layer_type()` — which runs per layer — makes every chart pay one build **per layer**, and that is exactly why #231 applied it to one geom only.

So it is asked once per chart in the orchestrator, which is about to build the plot anyway. Measured on a 5,000-point two-layer chart, five runs each:

```
with the pass     save_html median 3.376 s
without           save_html median 3.221 s
one ggplot_build            0.140 s
```

~0.15 s, one build — and **the same 0.14 s however many layers the chart has**, where the per-layer alternative would multiply it.

That is also the right place on its own terms, which is the reason to prefer it even if the costs were equal. Emptiness is not a fact about what *kind* of chart a layer is: `detect_layer_type()` still answers `point` for an empty `geom_point()` (asserted), and the orchestrator is what decides there is nothing to make a processor for.

The tag is `skip`, not a fourth answer, so the #176 guard reads it for free — a chart announcing itself as interactive with nothing in it is worse than an image, because an image at least says what it is.

## The two existing checks stay, and that was tested rather than assumed

The obvious follow-on is that this makes `unread_layer_type()` and the `GeomQuantile` emptiness check redundant. Mutation-tested both against this change:

| mutation | outcome |
|---|---|
| drop the `GeomQuantile` emptiness check | **1 failed** — still live |
| `unread_layer_type()` always answers `"unknown"` | **2 failed** — still live |

Both survive because they answer `detect_layer_type()`'s own contract, which tests ask directly, where this answers what the orchestrator should build. Two questions, converging on `"skip"`.

## Verification

Full suite: **6391 passed, 0 failed, 0 errors, 99 skipped**, up from 6381. `roxygenise()` regenerates only `man/Ggplot2PlotOrchestrator.Rd`, for the new documented method.

10 new assertions across four tests: each of the four ghost shapes gone, real layers untouched (including a *real* smooth, which emits a single series too — so a rule counting `data` alone would have called it empty), the all-empty fallback, and the classifier's unchanged answer.

Mutations on the new pass, each alone against a restored baseline:

| mutation | outcome |
|---|---|
| the pass never runs | killed (5 failed) |
| the pass skips every layer | killed (18 failed, 1 error) |
| tags `"unknown"` instead of `"skip"` | killed (9 failed) |

---


jooyoungseo:
All three taken, in `209fbb2`. The first was a real hole and you were right about it.

**1 — patchwork leaves.** Confirmed by reading the path you named: `ggplot2_patchwork_utils.R:411` and `:582` call `adapter$detect_layer_type(layer, leaf_plot)` directly, and `process_patchwork_plot()` never calls `detect_layers()`. So a composed chart would have kept ghosting while every plain one stopped — a fix that closes the issue in name only.

Fixed by stating the rule once as **`layers_that_drew_nothing()`** and asking it in all three places that classify layers: the orchestrator's pass and both leaf sites. One `ggplot_build()` per plot or per leaf, before the loop, still never per layer.

I could not *run* it: patchwork is not installed in this environment. It is in `Suggests`, so the new test is guarded with `skip_if_not_installed("patchwork")` and runs on CI. What I could run is the facet twin beside it, which exercises the same helper through `process_faceted_plot()` — measured `point(10) | point(10)`, no ghost — so that path is now pinned as well as covered.

**2 — two builds per chart.** Accurate, and worth having written down: the benchmark measured the delta, not the total, and the second build is stacked on an existing one rather than standing alone. Left as-is — reusing the first build for the gtable pass depends on `plot_for_render` not having diverged from `private$.plot`, which reordering and augmentation both do, so the condition is the work rather than the plumbing.

**3 — helper duplication.** Done, and it was worse than two copies: `rendered()` and `fell_back()` existed in **three** files — #227's, #230's and this one's — and testthat sources every file in `tests/testthat/` into one environment, so the later definitions silently shadowed the earlier ones. Now `helper-render.R`, with the schema reader and the `type(n)` summariser alongside. The summariser reports one string per *cell*, which is what made the facet and patchwork assertions natural to write.

Full suite: **6391 passed, 0 failed, 0 errors, 100 skipped** (the extra skip is the patchwork test).

---

Issue #237, opened 2026-08-23
`plot(type = "n")` draws nothing and is announced as a line chart of its data


jooyoungseo:
Found by sweeping base R's plotting functions the way #1138 swept Highcharts: draw each one and read what the orchestrator produces, rather than reasoning about what looks covered.

## The defect

`type = "n"` means *set up the axes and plot no points or lines*. It is how you draw a custom chart — establish the coordinate system, then add shapes with `segments`, `polygon`, `rect`. Measured:

```
  plot(type='p')    point   n=10  x=[1,2,3,4,5,6] y=[3,5,2,8,4,6]
  plot(type='l')    line    n=10  x=[1,2,3,4,5,6] y=[3,5,2,8,4,6]
  plot(type='s')    step    n=10  x=[1,2,3,4,5,6] y=[3,5,2,8,4,6]
  plot(type='n')    line    n=10  x=[1,2,3,4,5,6] y=[3,5,2,8,4,6]     ← draws nothing
```

A sighted reader sees an empty panel. A MAIDR reader is handed **ten points to walk and sonify** — the values passed to `plot()`, which it deliberately did not draw.

Worse than being unread, for the reason #572 gives about `triplot` and #232 about an empty layer: nothing about it looks wrong. The data is real, the axes are real, and the announcement is fluent. The only thing false is the claim that any of it was drawn.

This is the base R counterpart of #232 — a recognised layer that drew nothing reaching the schema — with `type = "n"` as the mechanism instead of an empty data frame.

## The shapes themselves are handled correctly

Worth recording, because it was my first guess and it is wrong. `polygon`, `segments`, `arrows` and `rect` drawn over a chart are **properly ignored**:

```
  plot(x, y) + segments(...)     point   n=10       ← the scatter alone, correctly
```

Every one of the four "shape" cases in my first sweep was reporting the *`plot()` call's* layer, not the shape's. The shapes contribute nothing, which is right.

## The rest of the sweep

For the record, so the next pass does not re-derive it. Each call drawn on a `pdf(NULL)` device and read through `BaseRPlotOrchestrator`:

**Read correctly** — `barplot`→bar, `boxplot`→box, `hist`→hist, `plot`→point, `pie`→pie, `contour`→contour, `image`→heat, `heatmap`→heat, `matplot`→line, `curve`→line.

**Registers nothing (14).** Two of these have an exact home in the grammar already:

| call | reads as | note |
|---|---|---|
[...]

PR #247, opened 2026-08-23
feat(base-r): read mosaicplot as the contingency table it draws

### LINKED ISSUE #242: `dotchart()` falls back to a picture although MAIDR has a dot plot trace
One of the two calls #237 named as having an exact home in the grammar that the base R adapter does not reach.

## What happens today

`dotchart` is classified HIGH, so its calls are recorded, but the adapter has no `detect_layer_type` branch for it and the factory no processor. That is the documented decline: a HIGH name with no processor takes the static-image path with a "Plot contains unsupported elements" warning. Measured end to end through `save_html()`:

```
  dotchart      fellback=TRUE   layers=-
  mosaicplot    fellback=TRUE   layers=-
  barplot       fellback=FALSE  layers=bar
  plot          fellback=FALSE  layers=point
```

So it is a whole chart unread, not a chart read wrongly — the better of the two failures, and still a chart a reader gets nothing navigable from.

## It has a home

`TraceType.DOT` exists for exactly this shape (xability/maidr#792): *"A bar chart's reading with a different mark, and a bar is oriented — a Cleveland dot plot is conventionally drawn with its categories down the page, which is the case this exists to get right."* That is `dotchart()`'s description almost word for word.

## What it draws

Echoed through `ggplotify::as.grob` and read back:

```
  dotchart ->
    graphics-plot-1-mtext-left-1        <- the category labels, in the margin
    graphics-plot-1-abline-h-1          <- the guide lines
    graphics-plot-1-points-1            <- the dots
    graphics-plot-1-bottom-axis-line-1
    ...
```

Only the dots carry a value; the guide lines and margin labels are frame. And `-points-1` is the same grob a scatter draws into, so the selector `BaseRPointLayerProcessor` already builds addresses them unchanged — measured on the rendered chart, `g#graphics-plot-1-points-1\.1 > use` resolved to 4 elements for 4 dots.

## Two things have to be right beyond the values

- **`orientation = "horz"`.** `dotchart()` puts the categories on the vertical axis and the value along the horizontal one. The core reads a `horz` layer's magnitude from `x` and its category from `y`; without the key the layer defaults to vertical and reads the category name where the magnitude belongs — #184 and #480 in one.
- **Only the one-value-per-category form.** `dotchart()` also takes a matrix, or a `groups` factor, and then draws every group's dots into the *same* points grob with a header per group in the left margin. A flat `dot` layer has nowhere to carry the grouping, so reading it would hand the reader one run of dots with no way to tell which group each belongs to and the group names silently dropped — worse than the picture it falls back to today. That form stays declined.

## Still open after this

`mosaicplot` — the other call #237 named. `MOSAIC` exists in the grammar (xability/maidr#804) and the chart draws one `-polygon-N` grob per tile, so the shape is there; the cross-classification the trace wants has to be recovered from the table the call was handed, which is a separate piece of work.

jooyoungseo:
Closes #242 — the last named gap from the #237 sweep.

`mosaicplot` is classified HIGH so its calls are recorded, but the adapter
had no branch for it and the factory no processor. That is the documented
decline: the whole figure takes the static-image path (#176). Not a wrong
reading — a whole chart unread, and the core has had `TraceType.MOSAIC` for
exactly this shape.

## Why `mosaic` and not `stacked_bar`

A mosaic is a two-way table drawn as tiles, and what separates it from a
stack is that the **column widths encode data too**: each column's width is
that category's share of all observations.

A reader given only the segment heights has half the table — the conditional
proportions without the group sizes they were computed from — so a category
of six observations and one of six hundred read identically. That is the one
thing the chart is drawn to show.

## Nothing is inferred from the drawing

`mosaicplot()` is handed the table itself, so the recorded call carries every
number the trace wants. For `mosaicplot(HairEyeColor[, , 1])`:

```
       Eye
Hair    Brown Blue Hazel Green
  Black    32   11    10     3     <- column total 56
  Brown    53   50    25    15     <- 143
  Red      10   10     7     7     <- 34
  Blond     3   30     5     8     <- 46      total 279
```

emits, for the Black/Brown cell:

```json
{"x": "Black", "y": 0.5714, "z": "Brown", "width": 0.2007, "count": 32}
```

`0.5714` is 32/56, `0.2007` is 56/279. Four numbers per cell rather than one:
the tile's conditional proportion, its column's share carried on every cell
of the column (the grammar's unit is the point, and a flat list has nowhere
else to put it), the cell's own count — a mosaic is drawn *from* counts and
they are the numbers a reader quotes back — and the fill level.

The nesting is `data[[fill]][[category]]`, the shape the stacked bar
processor already produces, because `MosaicTrace` extends `SegmentedTrace`
and navigates category-then-series the same way.

## The judgement calls

**The two dimension names go where the grammar puts them, not where the chart
draws them.** `mosaicplot()` labels its y axis with the second dimension —
"Eye" for this table — but a segmented layer's `y` holds the *magnitude* and
its `z` the fill. So the second dimension is named on `z`, and `y` says what
its numbers are:

```json
{"x": {"label": "Hair"}, "y": {"label": "Proportion"}, "z": {"label": "Eye"}}
```

`ylab` follows the dimension it names rather than the axis it shares a letter
with, for the same reason.

**A three-way table is declined.** `mosaicplot()` splits recursively for
three dimensions and more, and a `mosaic` layer has one category axis and one
fill — so a deeper table has nowhere to put its later dimensions, and reading
it flat would announce a cross-classification the chart does not claim. It
keeps the static-image fallback.

**A plain matrix is read with the levels base R gives it.** `mosaicplot()`
coerces with `as.table()`, which supplies "A", "B", … and then draws those
letters beside the tiles — so they are the chart's own labels rather than an
invention.

**A column drawn with no tiles is announced but not addressed.** A
zero-width column makes `mosaicplot()` draw fewer polygons than the table has
cells — measured, 2 grobs for a 2×2 whose second column is all zeros. The
cells are still all emitted, so a reader arriving there hears that the column
is empty rather than finding it missing; but the layer takes no selectors at
all rather than mapping them positionally onto the wrong tiles.

## Selectors are found, not guessed

Measured on a rendered `mosaicplot(HairEyeColor[, , 1])`: gridGraphics draws
one `-polygon-N` grob per cell and nothing else as a polygon — 16 for a 4×4
table, with no frame among them. Their geometry gives the order:

```
grob         xmin     xmax     ymin     ymax
p1        50.0000 219.8065 442.4000 950.0000   <- column 1 (Black), fill 1
p2        50.0000 219.8065 249.0125 423.5000   <- column 1, fill 2
...
p5       237.8065 671.4194 620.7699 950.0000   <- column 2 (Brown), fill 1
```

`polygon-1..4` share the leftmost column's x extent, `-5..8` the next, and
within a column the grob number runs down the fill levels. So cell (fill *f*,
column *c*) is grob `(c - 1) * fills + f`.

The grobs are **found in the tree** rather than named by formula, which is
the discipline `base_r_pie_layer_processor.R` already applies. That walk is
now shared — `find_graphics_plot_grobs()` in `base_r_selector_utils.R` — and
the pie's own goes through it, so this removes a duplicate rather than adding
a third copy.

## The exemplar moves for the third time

`test-base-r-contour-fallback.R`, `test-base-r-unrecorded-calls.R` and
`test-panel-scoped-fallback.R` each need "a recorded call with no processor"
as a subject. `dotchart` played it until #242 gave it a processor, then
`mosaicplot` until this PR does. Each move meant editing the same sentence in
three files, so it now lives in `helper.R` as `draw_unread_base_r_chart()`
and the three call it. `assocplot` is the current one.

Worth noting what that helper's comment records: it must call `assocplot()`
**unqualified**. maidr patches the name on the search path rather than inside
the `graphics` namespace, so `graphics::assocplot()` reaches the original and
records nothing — the chart then falls back for the wrong reason and all
three tests would have passed without testing anything. Measured: 0 recorded
calls via the qualified form.

## Verification

`tests/testthat/test-base-r-mosaicplot.R`, 15 tests / 45 assertions.

Mutation sweep — each reverted alone against a restored baseline:

| Mutation | Failures |
|---|---|
| transpose the nesting to category-major | 8 |
| `width` from the row rather than the column total | 3 |
| `y` as the raw count, not the conditional proportion | 1 |
| divide anyway on an empty column | 2 |
| dimension names onto the wrong axes | 1 |
| transpose the tile index | 2 |
| accept a three-way table | 1 error |
| drop the short-grob-list guard | 1 error |
| all restored | 45/45 pass |

Full `testthat` suite: **6506 pass, 0 fail, 0 error, 100 skip**.
`man/` regenerated with `roxygen2::roxygenise()`.


jooyoungseo:
All four addressed in 5f6c100.

**1. `NEWS.md`** — added, following the `contour()` entry as the template.
Includes both limitations: a three-way table and the formula interface are
still pictures.

**2. `UNREAD_BASE_R_CALL`** — dropped. It was dead exactly as described.

**3. The selector snippet** — extracted as `polygon_cell_selector()` in
`base_r_selector_utils.R`, beside `find_graphics_plot_grobs()`. You're right
that leaving it was inconsistent with this PR's own stated goal; the pie now
goes through it too, so both duplications are gone rather than one.

**4. The formula interface** — measured, and it does decline:

```
mosaicplot(~ Hair + Eye, data = df)

fn: mosaicplot   arg names: [ |data ]
  arg1 class: formula   is.language: TRUE
  has data arg: TRUE
layers: 0
```

Intentional, but only properly so now that it has been measured — thank you
for flagging it. Filed as **#248** rather than folded in here, because it is
more than relaxing the guard:

- The formula has to be *evaluated* against the recorded `data`, which is a
  different kind of operation from "read the argument you were given" — and
  that is the sentence this processor's whole design note rests on.
- `mosaicplot.formula()` has a `Freq`-column convention that has to be
  matched rather than guessed; a data frame of raw rows and one of
  pre-counted cells reach it differently. Both were measured to decline
  today, so neither currently misreads.

The table is recoverable rather than invented —
`xtabs(Freq ~ Hair + Eye, data = df)` reproduces exactly what the direct call
is handed — so it is worth doing, just not inside this PR.

**On the sandbox note:** the suite does run here. Full `testthat` after these
changes: **6506 pass, 0 fail, 0 error, 100 skip**, and the mosaic file's own
45 assertions plus the pie's 81 both stay green through the
`polygon_cell_selector()` extraction — which is what makes it a safe
refactor rather than a hopeful one.

---

Issue #252, opened 2026-08-24
A qqline() over a Q-Q plot is recorded only so it can be declined — read it as the reference line it draws


jooyoungseo:
`qqnorm()` and `qqplot()` are now read as the quantile scatter they draw (#251). `qqline()` — which is how nearly every Q-Q plot in the wild is finished — is not, and the chart carrying one falls back to a static image.

## Why it is recorded at all

`stats::qqline()` ends in `abline(int, slope, ...)`, and that `abline` is reached from *inside* the `stats` namespace, where maidr's search-path wrapper never sees it. Measured: before it was listed, `qqnorm(x); qqline(x)` recorded exactly one call — the `qqnorm` — and the reference line left no trace at all.

That mattered as soon as `qqnorm` became readable. A chart drawing a scatter *and* a fitted reference line would have come out as a bare scatter with the line silently missing from it, which is worse than the picture it used to fall back to, because a picture at least says what it is. So `qqline` went into `LOW` with no `detect_layer_type()` branch: it types `unknown`, `has_unsupported_layers()` sees it, and the chart keeps falling back exactly as it did before. That is the lower of the two claims, not a reading.

## Why reading it looks cheap

Two things were measured and both point the same way.

**The grob name is the one the adapter already addresses.** `qqnorm(x); qqline(x)` and `plot(x, y); abline(0, 1)` export the *same* grob:

```
graphics-plot-1-abline-ab-1
```

so `BaseRLineLayerProcessor`'s existing abline selector reaches it unchanged.

**The endpoints come from `stats`' own primitives.** `qqline()`'s body is four lines, and the line it draws is the one through two points:

```r
qs <- stats::quantile(y, probs, names = FALSE, type = qtype, na.rm = TRUE)
xs <- distribution(probs)
```

with `probs` defaulting to `c(0.25, 0.75)` and `distribution` to `qnorm`. Reading those two anchors is a reading rather than a re-derivation; the extension to the plot's x range is what `extract_abline_data()` already does for an `abline`.

## What still has to be worked out

`extract_abline_data()` gets its x range from `get_x_range_from_group()`, which reads the group's HIGH call's first argument as the x data. On a `qqnorm` group that argument is the **sample**, not the theoretical quantiles, so the range would be wrong — the same class of mistake #251 exists to avoid. The reference line's range has to come from the drawn pairs, which the Q-Q processor already computes.

[...]

Issue #251, opened 2026-08-24
Eleven base R plot types are recorded but read as nothing, and four of them already have an exact home


jooyoungseo:
`.base_r_function_classes$HIGH` lists twenty-two entry points, and its own comment is explicit that being listed is "the *lower* of the two claims, not a promise of a reading". This is the measurement of which claim each one has.

Every call driven through `BaseRPlotOrchestrator$generate_maidr_data()` on a fresh device, counting the layers it produced. `barplot` and `plot` are the controls, and they show the harness works.

## Measured

```
barplot          bar              stripchart       (no layers)
hist             hist             qqnorm           (no layers)
boxplot          box              qqplot           (no layers)
plot(x, y)       point            filled.contour   (no layers)
pie              pie              sunflowerplot    (no layers)
mosaicplot       mosaic           assocplot        (no layers)
dotchart         dot              fourfoldplot     (no layers)
contour          contour          spineplot        (no layers)
image            heat             cdplot           (no layers)
curve            line             pairs            (no layers)
matplot          line             persp            (no layers)
```

Eleven read; eleven produce nothing, which means the whole figure takes the static-image path. That is the right answer for some of them and a gap for the rest.

## Four already have an exact home

Not new trace types — existing ones the base R adapter does not reach:

| call | reads as | why it fits |
|---|---|---|
| **`qqnorm`** | `point` | A quantile-quantile plot *is* `plot(x, y)` over computed quantiles, and `plot` already reads as `point` on this very list. The theoretical quantiles are the x and the sample's are the y; nothing needs deriving. |
| **`qqplot`** | `point` | The same chart with two samples instead of one and a normal. |
| **`stripchart`** | `point` per group | One row of points per category — the shape `sns.stripplot` is read as in xability/py-maidr#586, and the same one `dotchart` already has a processor for. |
| **`filled.contour`** | `contour` | `contour()` already reads, and this is the same computation with the bands filled. The one thing to get right is that it draws a **legend panel beside the plot**, so the layer must be scoped to the contour panel rather than to the figure. |

`qqnorm` and `qqplot` are the cheapest: they were added to `HIGH` in #216 purely so the figure would fall back rather than error, and the reading needs no new machinery at all.

## Four more with a plausible home, each wanting a judgement first

[...]

PR #253, opened 2026-08-24
feat(base-r): read qqnorm and qqplot as the quantile scatter they draw

jooyoungseo:
A base R Q-Q plot fell back to a static image. Both names are classified HIGH — #216 put them there so `save_html()` would not stop with "No Base R plots detected" on a chart plainly on the device — but the adapter had no branch for either and the factory no processor. Not a wrong reading; a whole chart unread, and it is a scatter.

Closes part of #251, which listed eleven base R entry points that draw but are not read. Two of them here.

> **Branch note.** The commit list shows `cecb1a6`, the pre-squash tip of this branch left behind when #250 was squash-merged. Its tree is byte-identical to main's `dbfc946` (`git diff cecb1a6 dbfc946` is empty), so it contributes nothing to merge. It is still in the history because force-pushing and deleting the stale remote branch are both blocked in this environment, so the only available push was a fast-forward on top of it. **The change under review is the second commit alone.**

## The coordinates are not the arguments

This is what separates a Q-Q plot from every other base R scatter, and it is why reusing `BaseRPointLayerProcessor` unchanged would have been *worse* than the fallback rather than better. Every base R processor reads a call's recorded arguments rather than the drawn grob, and a Q-Q plot's arguments are samples, not positions:

| call | argument | what the chart draws |
| --- | --- | --- |
| `qqnorm(y)` | one sample | that sample against theoretical quantiles the function computes |
| `qqplot(x, y)` | two samples, possibly different lengths | one interpolated pair per point of the **shorter** |

Read as an ordinary scatter, `qqnorm` would have announced the raw sample on an axis of standard deviations — the one reading a Q-Q plot most needs not to have, since the comparison between the two is the entire point of the chart — and `qqplot` would have announced eight points over a chart that draws five, naming values no mark on the page stands for.

## Nothing is re-derived

`stats` computes the pairs, and both functions hand them over without drawing. The recorded arguments are forwarded **whole** rather than picked apart, which is what makes the awkward cases free. Measured on eight values against five:

```
qqnorm(x, plot.it = FALSE)$x    0.1525 -0.8525 0.8525 …  the caller's order, not sorted
qqnorm(x, datax = TRUE, ...)    the same pair, swapped
qqplot(x, y, plot.it = FALSE)   both length 5; x interpolated to
                                1.20 3.05 4.00 5.00 6.90
```

`datax = TRUE` is the clearest of them: it swaps which axis holds the sample, and forwarding means nothing in maidr has to know the argument exists.

The pairs come back in the order the call drew them, which is the order the `points` grob lays its marks down in, so the selector list keeps pairing positionally. **Nothing is sorted** — sorting would highlight a different point from the one announced.

## Labels

`qqnorm()`'s "Normal Q-Q Plot", "Theoretical Quantiles" and "Sample Quantiles" are constants in its own signature and are drawn on the chart, so announcing them reports the figure. `qqplot()`'s defaults are `deparse1(substitute(x))` — the caller's own expression, gone by the time the wrapper has recorded evaluated values — so its axes are left for the renderer's generic, on the reasoning the point processor already states: a guessed noun is worse than none.

The axis grid comes from the **drawn pairs**, not the recorded arguments: on `qqnorm` one of the two axes is not a sample at all, so a grid taken from the arguments would announce a range of roughly 1 to 7 for an axis the chart draws from −2 to 2.

## Two shapes stay pictures, deliberately

- `qqplot(conf.level = )` draws a confidence band as well as the points, through a `polygon()` from inside `stats` that the wrapper never sees.
- `qqline()` draws the reference line nearly every Q-Q plot is finished with, likewise an `abline()` from inside `stats`. Measured: before this change, `qqnorm(x); qqline(x)` recorded **one** call and the line left no trace at all.

Either would have shipped a chart with a drawn mark silently missing from it, which is worse than the picture, because a picture at least says what it is. So `qqline` is now recorded (LOW, no `detect_layer_type()` branch) purely so it can be declined — the same answer as before the reading existed. Reading the line itself is #252.

## Selectors needed nothing

`qqnorm()` exports its marks under `graphics-plot-N-points-1` — measured — which is the grob the point processor already looks for.

## Falsification

Eleven mutations, all killed against a clean baseline of 35 assertions:

```
baseline                                       ...................................
pairs sorted                                   6 failed
datax not forwarded                            2 failed
plot.it not stripped                           1 failed
unevaluated guard dropped                      1 failed
qqplot given qqnorm's labels                   2 failed
qqnorm's own labels dropped                    4 failed
grid from the recorded args                    1 failed
conf.level band waved through                  1 failed
qqline not recorded                            1 failed
qqnorm's own title dropped                     1 failed
the recorded sample read, not the quantiles    4 failed
```

## Also here

- `tests/testthat/test-base-r-unrecorded-calls.R` is split rather than weakened. #216's claim — a recorded call never stops the save — is still asserted over all eight names; the fallback assertion now covers the six still unread, and a new test asserts the two that gained a reading are **read**, so a regression putting either back on the fallback shows up as a contradiction between two files rather than as one quietly relaxed expectation.
- `NEWS.md` entry, `NAMESPACE` and `man/` regenerated with roxygen2 8.1.0.

## Gate

Q-Q file 35 assertions, unrecorded-calls file 59, and the **full suite green** (`testthat::test_local`, exit 0 — remaining output is skips for absent Suggests packages and pre-existing ggplot2 warnings).

---


jooyoungseo:
Good catch on the `conf.level` question, and the answer is worth writing down rather than leaving in a review thread.

**Measured on R 4.3.3 in this environment:**

```
formals(stats::qqplot)$conf.level        NULL
eval(...)                                is.null → TRUE
qqplot(x, y, plot.it = FALSE)            names: x, y
qqplot(x, y, conf.level = .95, plot.it)  names: x, y, lwr, upr
```

and `stats::qqplot`'s body takes the early return before drawing anything but the points:

```r
if (is.null(conf.level)) {
    if (plot.it) plot(sx, sy, xlab = xlab, ylab = ylab, ...)
    return(invisible(list(x = sx, y = sy)))
}
```

So the default really does mean "off", and the check is correct as written today. On the R-version half: `conf.level` was added in R 4.3, so on the older R the `DESCRIPTION` still supports the formal is simply absent — the argument falls into `...`, is ignored, and my check declines a chart that draws no band. Over-declining, which errs towards the picture.

**But your framing of the defect is the right one and I acted on it.** The check reads "no band" off the caller's *silence*, and that is sound only while the default stays NULL — nothing in the repo said so, so a future R could have turned every plain `qqplot()` into a chart with a drawn region missing from its reading, quietly.

I first tried the obvious fix — a `qqplot_conf_level()` helper consulting `formals(stats::qqplot)` instead of assuming. Then I mutation-tested it, and **the mutation survived**:

```
--- M12 the library default not consulted     38 assertions, all passed
```

Of course it did: with a NULL default the helper and the plain lookup are observationally identical, so no test on this R can tell them apart. The helper was a branch that could not be exercised — exactly the kind of code the repo's own comments warn about elsewhere.

So I reverted it and **asserted the assumption instead**, which is falsifiable:

```r
testthat::expect_null(eval(formals(stats::qqplot)[["conf.level"]]))

plain  <- stats::qqplot(SAMPLE, OTHER, plot.it = FALSE)
banded <- stats::qqplot(SAMPLE, OTHER, conf.level = 0.95, plot.it = FALSE)
testthat::expect_setequal(names(plain), c("x", "y"))
testthat::expect_true(all(c("lwr", "upr") %in% names(banded)))
```

Asserted against `formals()` rather than written down as a constant, so an R release that changed the default fails this test rather than silently degrading a reading. Both halves bite:

```
--- asserted against a formal with a real default   1 failed
--- a plain call claimed to carry bounds            1 failed
```

The adapter comment now says where the assumption lives and why it is not a `formals()` lookup. Simpler code, and the thing that could go wrong is now something a test catches.

**On the duplicated `do.call`** — agreed and deliberate. `quantile_pairs()` runs once for the data and once for the grid, and the alternative is caching state on a processor that is constructed per layer per render, which buys a `qqnorm` call on samples that are already in memory. Left as is; noting it here so the next reader finds the reasoning rather than the question.

---

PR #259, opened 2026-08-24
feat(base-r): read cdplot as the 100% stacked area it draws

jooyoungseo:
Refs #251. Fifth of the eight calls from #216 to move past the static-image fallback, after `qqnorm`/`qqplot`, `filled.contour` and `spineplot`.

## What it draws, and what that is

A conditional density plot shows, for each value of a numeric `x`, how the levels of a factor divide up: stacked bands that fill the height, whose shares sum to 1 at every `x`. That is a normalized stacked area, which `Ggplot2AreaLayerProcessor` already emits for `position = "fill"` — so a `cdplot()` is read as `stacked_normalized_area`, with the same payload shape (one series per band, each a list of `{x, y, z}` where `y` is the band's **own** share) and one selector per band. The two adapters describe one chart the same way.

Before this, `cdplot()` had no branch in `detect_layer_type()`, the switch fell through to `"unknown"`, and the chart degraded to a picture.

## Where the numbers come from

`cdplot()` has a `plot` argument, so it can be asked for what it drew without drawing it again — the same line `BaseRSpineplotLayerProcessor` takes, and `graphics::cdplot` by the qualified name so the replay is not itself recorded as a second chart. It returns the **boundaries** between the bands, and stacking them the way `cdplot()` does (`rbind(0, boundaries, 1)`, band `i` from row `i` to row `i + 1`) gives the shares back.

Verified over **300 random charts**: every column's shares sum to 1, and no band comes out negative — the boundaries do not cross in the drawn range.

## Three things the picture does not show

**The level order is reversed.** `cdplot()` does `ylevels <- nlevels(y):1L` before drawing, so the factor's *first* level is the *top* band. A reading that followed `levels()` would name every band after the one drawn opposite it. The boundaries are named for all but the topmost level, so the top band is the one level left over — derived that way rather than by reproducing the `ylevels` reordering, which is `cdplot()`'s to define.

**The grid is trimmed.** The returned functions interpolate over `stats::density()`'s grid, which pads three bandwidths past the data. `cdplot()` throws that padding away before drawing — 372 of 512 points on a measured chart. Announcing the untrimmed grid would put readings either side of the data, where the chart has no marks. The grid itself is read off `environment(f)$x`, since `approxfun()` keeps its knots; recomputing `density()` would have to guess `bw`, `n`, `from`, `to` and `weights` back out of the recorded call.

**`subset` moves the whole chart.** `cdplot.formula` builds its model frame with it, so everything downstream is the subset's. Measured on `cdplot(b ~ a, data = d, subset = a > 45)`: the chart draws over 46.1–70.9 while the whole column runs from 25.5, and reading it without the subset put a fifth of the announced points left of the leftmost mark. It is applied by indexing the frame rather than by `model.frame(subset =)`, which substitutes its argument and evaluates the expression inside `data` — and the recorded call carries the already-evaluated vector, which is not a name `data` knows. Measured: that route returns an empty frame and every reading with it.

## Selectors

One `polygon` grob per band, in draw order, which is series order. The frame `cdplot()` draws around the plot is a polygon too, but gridGraphics names it `-box-1` rather than `-polygon-N`, so the anchored pattern does not collect it — there is a test for that. Withheld entirely when the counts disagree.

## Guards

The class doc states once what a returning `cdplot()` has already proved — `y` is a factor, `x` is finite, a formula names two variables, the grid overlaps the data, the factor has two or more levels — each measured against the error `cdplot()` raises instead. The NULL checks that remain are there so an unreachable shape cannot *throw* out of the pipeline rather than to re-check any of that, and where one check would only repeat another it is not written twice.

The one branch measured unreachable and kept anyway is documented as such at the point it sits.

## Tests

`tests/testthat/test-base-r-cdplot.R`, 25 cases: type, shares summing to one, no negative band, the reversed band order, a two-level factor, `ylevels` reordering, the trimmed grid, one grid shared by every band, the formula form reading the same chart, axis names from the model frame, a written label winning over them, the `y`/`z` split, title, selectors, the frame not collected as a band, a mismatched count withholding, `subset`, an `ylevels` naming a strict subset being declined, a level with an empty name still announced, a `plot` the caller wrote not displacing ours, and the unreadable/empty/absent cases.

`tests/testthat/test-base-r-unrecorded-calls.R` moves `cdplot` to the read list (now five of eight) and hands the "picture is the chart" case — which needs a formula shape and an unread call — to `sunflowerplot`.

## Verification

Full local suite: **127 files, no failures**. `roxygen2::roxygenise()` regenerated `man/` with no drift beyond the new topic. No line over the 100-column cap.

## Branch note

The first commit is a no-op merge. The remote branch tip still carried the pre-squash commits of #258, which merged as a squash, and force-push is not available here — so the tip is merged in with every conflict resolved to this branch's content. `git diff` against the feature commit reports no change; the merge only lets the push fast-forward, which is what keeps this diff to the new work.

---


jooyoungseo:
Both findings acted on in `d42194a`. Thanks — the first one was real and I had not thought to look for it.

## 1. The `subset` + NA crash — confirmed, and it was worse than "plausible"

Reproduced before touching anything, on 80 rows with three `NA`s in the predictor and a subset of length 80:

```
rows: 80  model.frame rows: 77
subset length: 80  TRUEs: 57
process -> RAISED: missing value where TRUE/FALSE needed
predictor has NA: TRUE   length: 57
```

Exactly the chain described: `na.omit` shrinks the frame first, the positional subset overruns it, R pads the overrun with `NA` rows rather than erroring, the `NA` reaches `min()` in `drawn_grid()`, and `if (sum(kept) < 2L)` raises **out of `process()`** — which `BaseRPlotOrchestrator$process_layers()` does not catch, so the figure goes down rather than this layer declining.

Fixed as suggested, and for the reason the suggestion gives rather than just to silence it: `model.frame()` applies `subset` *before* `na.action`, so `cdplot.formula` gets that order and this has to reproduce it — `na.pass`, subset, then `na.omit`. Checked against what `cdplot()` actually drew:

```
NA-free subset         series=3  announced [45.803, 77.160]  cdplot drew [45.795, 77.168]
subset carrying NAs    series=3  announced [45.803, 77.160]  cdplot drew [45.795, 77.168]
```

Three regression tests: a subset over data with missing rows; a subset that itself carries `NA`s (`score > 45` is `NA` wherever `score` is, which is how a caller actually writes it); and one asserting the subset picks *the rows `cdplot` picked* rather than merely three series — an off-by-three alignment would pass a count check and still announce a range the chart never drew.

Mutation-swept: removing `na.pass`, removing the `na.omit`, and dropping the subset application each fail the suite.

## 2. Memoising `variables()` — applied

You are right about the convention, and my reason for not having it was weak: I had removed a cache earlier in this branch because a mutation deleting it *survived* the tests. That is the correct outcome for something whose purpose is not behavioural, and not a reason to leave the work being done three times. It is back, with the reason written into the doc so the surviving mutation is explained rather than rediscovered.

Full local suite re-run after both: **127 files, no failures**.

---

PR #261, opened 2026-08-24
fix(base-r): stop recording stem(), which draws nothing

### LINKED ISSUE #260: `stem()` draws nothing, is recorded as a chart, and takes the real chart beside it down to a picture
Found by sweeping every `package:graphics` export through `classify_function()` and then through `BaseRAdapter$detect_layer_type()`, looking for names that are recorded but unread.

`stem` is listed in `HIGH` in `R/base_r_function_classification.R:28`, between `stripchart` and `pie`. It does not belong there: **`stem()` writes a stem-and-leaf display to the console.** It opens no device, draws no marks, and returns invisibly. It lives in `package:graphics` for documentation reasons, not because it plots.

## Measured

```
hist alone               error=none                                   static_image=FALSE
hist then stem           error=none                                   static_image=TRUE
stem alone               error=Failed to create fallback image        static_image=NA

one barplot records: 1 -> barplot
hist+stem  records: 2 -> hist, stem
```

Two distinct failures:

**1. `stem()` alone stops the save.** `"Failed to create fallback image"` — the device has a recorded call and nothing drawn, which is the exact shape #216 identified for `qqnorm(plot.it = FALSE)` and handled by declining to record it. Here the message is actively misleading: it says a fallback image could not be made, when the truth is that there is no plot. Had `stem` never been recorded, the caller would get `"No Base R plots detected. Please create a plot first"`, which is the accurate message and the one `plot.new()` already gets.

**2. `stem()` beside a real chart destroys the real chart's reading.** A histogram alone renders interactively; the same histogram followed by a `stem()` call degrades to a static image with `"Plot contains unsupported elements"`. The console output the user asked for costs them the accessible chart they also drew — and nothing on the page says why.

That second one is the reason this is a bug rather than tidiness. `hist(x); stem(x)` is not a contrived pairing: they are the two ways of looking at one distribution, and R's own documentation puts them side by side.

## The fix

Remove `"stem"` from the `HIGH` list. Nothing else. The comment block already in that file states the rule this follows:

> Being listed here is therefore the *lower* of the two claims, not a promise of a reading.

But the lower claim is still a claim — that a chart is on the device. `stem()` makes no such chart, so the right list for it is none, exactly as for `hist(x, plot = FALSE)` and `qqnorm(x, plot.it = FALSE)`. Those two are declined at the wrapper because the *argument* says nothing will be drawn; `stem()` needs no argument check because it is unconditional.

Worth a test on each half: that `stem()` alone reports "No Base R plots detected" rather than a failed fallback, and that `hist(); stem()` still renders the histogram interactively.

## Also found by the same sweep, filed here only as notes

Recorded-but-unread `graphics` entry points, beyond the four already on #251 (`persp`, `sunflowerplot`, `fourfoldplot`, `pairs`):

- `assocplot` — a Cohen-Friendly association plot, a mosaic variant drawn from Pearson residuals
- `coplot` — a conditioning plot, a trellis of scatter panels

And drawing functions `classify_function()` returns `UNKNOWN` for, so they are not recorded at all:

- `smoothScatter` — a 2-D density scatter
- `stars` — star/segment glyph plots, which MAIDR has `radar` for
- `plot.design` — factor means
- `bxp` — draws a box plot from precomputed `boxplot.stats` output, where `boxplot()` itself is read
- `rug` — the base R spelling of `geom_rug`, which the ggplot2 adapter reads

Those are chart-support gaps rather than defects, and none of them stops a save or degrades a neighbouring chart the way `stem` does. Recording them here so the sweep does not have to be redone.

jooyoungseo:
Closes #260.

## The defect

`stem()` writes a stem-and-leaf display to the console. It opens no device, draws no marks, and returns invisibly — it lives in `package:graphics` for documentation reasons rather than because it plots. It was nevertheless listed in `HIGH` in `R/base_r_function_classification.R`, between `stripchart` and `pie`, so every call was recorded as a chart.

Measured:

```
hist alone               error=none                                   static_image=FALSE
hist then stem           error=none                                   static_image=TRUE
stem alone               error=Failed to create fallback image        static_image=NA

one barplot records: 1 -> barplot
hist+stem  records: 2 -> hist, stem
```

**`stem()` alone stopped the save.** `"Failed to create fallback image"` — a recorded call over a blank device, the exact shape #216 found for `qqnorm(plot.it = FALSE)`. The message is actively misleading: it says the picture of a plot could not be made, when the truth is that there is no plot.

**`stem()` beside a real chart destroyed the real chart's reading.** A histogram alone renders interactively; the same histogram followed by a `stem()` call degraded to a static image with `"Plot contains unsupported elements"`. That is the half that costs a reader something — `hist(x); stem(x)` is not contrived, they are the two ways of looking at one distribution and R pairs them in its own documentation, and the console output the caller asked for took away the accessible chart they also drew, with nothing on the page to say why.

## The fix

Remove `"stem"` from the list. Nothing else. The comment block already in that file states the rule:

> Being listed here is therefore the *lower* of the two claims, not a promise of a reading.

But the lower claim is still a claim — that a chart is on the device — and `stem()` makes no such chart. Unlisted, `save_html()` reports `"No Base R plots detected. Please create a plot first"`, which is accurate and is what `plot.new()` already gets, and a real chart beside it is read on its own terms. The same call the wrapper makes for `hist(x, plot = FALSE)` and `qqnorm(x, plot.it = FALSE)`, except that those need an argument check and this one does not: `stem()` never draws.

After:

```
hist then stem           static_image=FALSE          (was TRUE)
stem alone               error=No Base R plots detected
hist+stem  records: 1 -> hist                        (was 2)
```

## Tests

Three, in `tests/testthat/test-base-r-unrecorded-calls.R` — the file about what a recorded call does to the save — plus the assertion in `test-function-classification.R` that listed `stem` as `HIGH`, which is the assertion this changes.

Restoring `"stem"` to the list fails four of them, so they are load-bearing rather than descriptive.

## How it was found

By sweeping every `package:graphics` export through `classify_function()` and then through `BaseRAdapter$detect_layer_type()`, looking for names recorded but unread. #260 records the rest of what that turned up — `assocplot` and `coplot` recorded but unread, and `smoothScatter`, `stars`, `plot.design`, `bxp` and `rug` not recorded at all. Those are chart-support gaps rather than defects: none of them stops a save or degrades a neighbouring chart, so none is in this change.

## Verification

Full local suite: **127 files, no failures** — run twice, once before the rebase onto the `cdplot` merge and once after.

## Branch note

The first commit is a no-op merge. The remote branch tip still carried the pre-squash commits of #259, which merged as a squash, and force-push is not available here — so the tip is merged in. `git diff` against the fix commit reports no change; the merge only lets the push fast-forward, which is what keeps this diff to the new work.

---


jooyoungseo:
Good catch, and it was worse than one missing entry. Added in `98d39fe`.

`NEWS.md` is hand-maintained here, and `spineplot()`, `filled.contour()` and `stripchart()` — all landed earlier today — each have an entry. So the convention is live, not historical, and I simply skipped it.

**Checking that also turned up a second gap: `cdplot()` has no entry either**, and #259 merged an hour ago. That one is mine from the previous pull request rather than a change in convention, so this commit adds both:

- **New Features** — `cdplot()` read as the 100% stacked area it draws, including the three things the picture does not show (the reversed level order, the trimmed density grid, and `subset` moving the whole chart).
- **Bug Fixes** — `stem()`, this pull request's change, with both halves of what recording it cost.

Thanks — the "several sibling fixes of this same shape" observation is exactly what made me go and look rather than just add the one you asked for.

On the verification note: the sandbox not being able to run the suite is worth stating as you did. For what it is worth from this side, the full suite was run twice on this branch — once before rebasing onto the `cdplot` merge and once after — 127 files, no failures, and the mutation of restoring `"stem"` to the list fails four of the new assertions, so they are load-bearing rather than descriptive.

---

Issue #262, opened 2026-08-24
Twelve more base R calls draw a chart and then get told "No Base R plots detected"


jooyoungseo:
#216 found eight functions that drew a chart and whose `save_html()` then reported that no plot existed. It fixed those eight. **The same defect is in twelve more**, all standard, documented R plotting calls, and the sweep that found them is repeatable.

## Measured

Each call driven with a **bare name**, as a caller writes it — a qualified `stats::acf(v)` bypasses the search-path patch and would have made this list wrong in both directions. Two questions per call: did anything reach the device's display list, and did the save nevertheless refuse?

```
acf                  drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
pacf                 drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
ccf                  drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
biplot               drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
interaction.plot     drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
cpgram               drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
monthplot            drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
spectrum             drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
lag.plot             drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
termplot             drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
stars                drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED
bxp                  drew=TRUE  refused=TRUE   <== DRAWS BUT REFUSED

assocplot            drew=TRUE  refused=FALSE
coplot               drew=TRUE  refused=FALSE
barplot control      drew=TRUE  refused=FALSE
```

The refusal is `"No Base R plots detected. Please create a plot first"` — told to a caller whose chart is on the device. That is #216's message, its cause and its remedy, twelve more times.

**A correction I made to my own sweep, worth recording so the next one avoids it.** My first pass used `stats::acf(v)`, `graphics::stars(m)` and so on, and reported `assocplot` and `coplot` as broken too. They are not: they are in `HIGH` and patched, and a qualified call simply does not go through the patch. Any sweep of this kind has to use bare names or it measures the wrong function.

## What goes through `plot()` is already fine

[...]

PR #265, opened 2026-08-24
feat(base-r): read bxp() as the box plot it draws

### LINKED ISSUE #264: bxp() draws a box plot and is described as a picture of one
Follow-up to #262, which listed `bxp` among twelve calls that draw without going through `plot()`. #263 put it in `HIGH`, which stopped `save_html()` reporting no plot; it now degrades to a static image. This is about the reading.

## What `bxp()` is

`bxp()` is `boxplot()`'s drawing half. `boxplot.default()` computes the five-number summaries and hands them to `bxp()`, which puts the boxes, whiskers, medians and outliers on the page. Calling it directly is how you draw boxes from summaries you already have — from `boxplot(plot = FALSE)`, from `boxplot.stats()`, or computed elsewhere entirely.

Measured today, with bare calls (a qualified `graphics::bxp()` bypasses the search-path patch — the caveat #262 records):

```
bxp(z)                fell back to a picture, warned "unsupported elements"
bxp(z, horizontal)    fell back to a picture, warned
bxp(z, xlab=, ylab=)  fell back to a picture, warned
boxplot(...)          type=box, orientation=vert   <- control
```

## Why this is the cheap one of the twelve

The marks are identical, not merely similar. Drawn off-screen and echoed through `gridGraphics`, `bxp(z)` and the `boxplot()` call that produced `z` emit **the same grob names in the same order** — `polygon-1`, `segments-1`, `points-1`, and so on. The only names that differ are the `xlab` and `ylab` grobs, which a bare `bxp()` genuinely does not draw.

So every selector `BaseRBoxplotLayerProcessor` already builds addresses a `bxp()` chart unchanged — including the part that is easy to get wrong, the index shift each box with no outliers puts on the boxes after it. The reading is the box plot's.

## The one difference

Where the summaries come from. `boxplot()` is handed the observations, so its processor replays `boxplot(plot = FALSE)` to recover them. `bxp()` is handed the summaries themselves, in its first argument — and replaying `boxplot()` on *that* reads the six-element list as six groups of numbers and summarises them, which is not the chart.

Two things `bxp()` shares with `boxplot()` should be shared including their limits: `horizontal = TRUE` means the same thing to both, and `at =` repositions boxes without reordering the drawing in either, so a non-monotonic `at` reads in drawing order here exactly as it already does for `boxplot()`.

## Argument spellings that have to work

`match_recorded_args()` names a recorded call's arguments but keeps the **author's** order, and leaves a positional dispatch argument unnamed. Measured:

```
bxp(Z)                          names=[]                slot 1 is z
bxp(z = Z)                      names=[z]               slot 1 is z
bxp(horizontal = TRUE, z = Z)   names=[horizontal, z]   slot 1 is not
```

so neither a named lookup nor a positional fallback covers the other on its own.

## Not in scope

The other eleven of #262 stay unread here. `stars` is the next candidate whose reading is a type MAIDR already has, and it is a harder call than this one: it draws one glyph per row as small multiples inside a single figure region, so what a `radar` layer would have to say about the layout is a separate question.

jooyoungseo:
Closes #264. Follow-up to #262, which listed `bxp` among twelve calls that draw without going through `plot()`; #263 put it in `HIGH`, which stopped the save failing and left it degrading to a picture. This is the reading.

## Measured, end to end, before and after

With **bare** calls — a qualified `graphics::bxp()` bypasses the search-path patch, the caveat #262 records:

```
bxp(z)               before: fell back, warned   after: type=box, orientation=vert
bxp(z, horizontal)   before: fell back, warned   after: type=box, orientation=horz
bxp(z, xlab=, ylab=) before: fell back, warned   after: type=box, labels used
boxplot(...)         unchanged in both                                  <- control
```

## The marks are identical, so the reading is shared rather than restated

`bxp()` is `boxplot()`'s drawing half: `boxplot.default()` computes the five-number summaries and hands them to `bxp()`, which puts the boxes, whiskers, medians and outliers on the page. Echoed through `gridGraphics`, `bxp(z)` and the `boxplot()` call that produced `z` emit **the same grob names in the same order** — `polygon-1`, `segments-1`, `points-1`, … The only names that differ are the `xlab` and `ylab` grobs, which a bare `bxp()` genuinely does not draw.

So every selector `BaseRBoxplotLayerProcessor` builds addresses a `bxp()` chart unchanged, including the part that is easy to get wrong: the index shift each box with no outliers puts on the boxes after it. On the same data the two readings come out **equal in both `data` and `selectors`**, and that equality is what the tests assert rather than restating the numbers.

## The one thing overridden

Where the summaries come from. `boxplot()` is handed the observations and replays `boxplot(plot = FALSE)`; `bxp()` is handed the summaries themselves. Replaying `boxplot()` on *that* reads the six-element list as six groups of numbers and summarises them.

- `read_stats()` is a new seam on `BaseRBoxplotLayerProcessor`, holding the replay it already did in two places.
- `BaseRBxpLayerProcessor` overrides only that, and inherits everything else.
- `detect_layer_type()` types `bxp` as `"box_stats"` — an internal routing name, like `spine`, `qq`, `strip` and `filled_contour`. The layer it emits is a `box` like any other, which a test pins: a routing name the core does not know would fail its trace factory.

Two things `bxp()` shares with `boxplot()` are shared including their limits: `horizontal = TRUE` means the same to both, and `at =` repositions boxes without reordering the drawing in either, so a non-monotonic `at` reads in drawing order here exactly as it already does for `boxplot()`.

## Argument spellings

`match_recorded_args()` names a recorded call's arguments but keeps the **author's** order, and leaves a positional dispatch argument unnamed. Measured:

```
bxp(Z)                          names=[]                slot 1 is z
bxp(z = Z)                      names=[z]               slot 1 is z
bxp(horizontal = TRUE, z = Z)   names=[horizontal, z]   slot 1 is not
```

Neither half of the lookup covers the other, and both are tested — including the whole reading through the out-of-order spelling, since nothing downstream may assume slot 1 either.

## Axis names

A bare `bxp()` keeps the generic pair. Unlike `boxplot(y ~ g)` it has no formula to take titles from and draws no axis-label grob at all, so naming the axes after anything in particular would announce text that is not on the page. `xlab` and `ylab` written on the call are used as usual.

## Mutations

Eight killed:

```
KILLED   override removed: subclass inherits the boxplot() replay
KILLED   named z ignored
KILLED   positional z ignored
KILLED   shape never validated
KILLED   five-row check dropped
KILLED   dispatch branch removed
KILLED   factory routes to the parent processor
KILLED   type dropped from the supported list
```

A ninth — **the parent no longer forcing `plot = FALSE`** — survived at first. That line predates this change and was untested, and dropping it puts a second chart on the device being read: measured, 0 display-list entries added with it, 47 without. It is now pinned rather than trusted, because moving an invariant is when it gets dropped.

## Verification

```
tests/testthat  == DONE, 0 failures (full suite, 127 files)
line length     no line over 100 in any changed file
man/            regenerated with roxygen2
NEWS.md         entry added under New Features
```

The last commit is a documented no-op merge: this branch's previous pull request was squash-merged, so the remote tip still carried its pre-squash commits and force-push is unavailable here. `git diff` between it and the commit before it is empty.

---


jooyoungseo:
The review is right, and I confirmed it by measurement before fixing. `8d20d0b`.

`bxp(horizontal = TRUE, Z)` is a call R accepts and draws — an unnamed argument binds to the first *remaining* formal, which is `z`. Driven through the recorder:

```
bxp(Z)                        names=[]               read_stats ok
bxp(z = Z)                    names=[z]              read_stats ok
bxp(horizontal = TRUE, z = Z) names=[horizontal, z]  read_stats ok
bxp(horizontal = TRUE, Z)     names=[horizontal, ""] read_stats NULL   <- wrong
```

and separately, `bxp(horizontal = TRUE, Z)` on a pdf device: **drew**.

I had the right *reason* for the fallback and the wrong rule for it. My own PR body quotes `match_recorded_args()`'s contract — "keeps the author's order and leaves a positional dispatch argument unnamed" — and then the code reads slot 1, which only follows if the dispatch argument was also written first. The three spellings I tested all happened to satisfy that, so the test agreed with the bug.

It fails quietly, which is the part worth recording: `is.list(TRUE)` is FALSE, so `read_stats()` declined and the layer shipped `box`-typed with `data = list()` — an empty interactive chart, with the caller's summaries sitting in the next slot. Not a raise, not a fallback to the picture.

Fixed with the pattern you pointed at: the fallback now finds the first **unnamed** argument, which is how `resolve_xy_args()` already resolves a positional argument and for the same reason. Two regression cases added — one on `read_stats()`, one driving the whole reading through the positional spelling, since nothing downstream may assume slot 1 either.

Verified the cases actually catch it. Against the old lookup:

```
-- 1. Failure ('test-base-r-bxp.R:194:3'): it takes the summaries from z, named or positional
-- 2. Failure ('test-base-r-bxp.R:217:3'): a call written out of order still reads its boxes
```

and green against the new one (`base-r-bxp: 37 passed`).

---

Issue #276, opened 2026-08-25
Read acf(), pacf() and ccf() as the correlogram they draw


jooyoungseo:
`acf()`, `pacf()` and `ccf()` are recorded — `save_html()` succeeds rather than reporting "No Base R plots detected", which #262 fixed — but nothing reads them, so the chart comes out as a picture.

## Measured

Through the public API on R 4.3.3, drawing each onto a `pdf(NULL)` device and calling `maidr::save_html()`:

```
== acf                == pacf               == ccf
   save: ok              save: ok              save: ok
   Warning: Plot contains unsupported elements. Rendering as static image
   instead of interactive MAIDR plot.
```

The same warning, and the same static image, for all three.

## What they draw, and what it reads as

A correlogram is one vertical segment per lag, drawn from the zero line up or down to the correlation at that lag, with a horizontal reference band for the significance threshold. That is a **spike chart**, and r-maidr already reads exactly that shape: `BaseRSpikeLayerProcessor` emits a `lollipop` layer for a baseline-anchored `segments()`/`vlines()` call, and `maidr/patch/spike` does the same in the Python binding (xability/py-maidr#664). matplotlib's `acorr()` and `xcorr()` — the same chart under different names — already read as `lollipop` there.

So this needs no core work and no new trace type: a `detect_layer_type()` branch and a processor that reads the lags and the correlations off the recorded call.

## Scope

- [ ] A `detect_layer_type()` branch for the three entry points
- [ ] A processor emitting one point per lag: the lag on the category axis, the correlation as the magnitude
- [ ] `ccf` is the one that differs — its lags run negative through positive, so the axis is signed rather than starting at zero
- [ ] Name the axes from what the call draws (`Lag` and `ACF`/`Partial ACF`/`CCF`) rather than leaving the `X`/`Y` fallback

The significance band is a reference line rather than data; whether it is worth emitting alongside is a separate question from reading the spikes, and this issue is the spikes.

## The rest of the twelve

#262 listed twelve calls that record but are not read. Measured the same way just now, five more are still unread and each falls out as a different shape:

| Call | Draws | Plausible reading |
|---|---|---|
| `interaction.plot` | one line per level, over a factor | a multi-line `line` layer |
| `monthplot` | a short horizontal segment per period, grouped by month | a `gantt`, or a spike per period |
| `lag.plot` | a grid of scatters, one per lag | the `pairs()` treatment (#272) |
| `stars` | one star/segment glyph per row | a radar per row, or a decline |
[...]

PR #280, opened 2026-08-25
feat(base-r): read monthplot as the seasonal subseries it draws

jooyoungseo:
Part of #262's tail — the sixth of the twelve recorded-but-unread base R
calls to gain a reading.

## What was wrong

`monthplot()` is recorded, so `save_html()` succeeds rather than reporting
"No Base R plots detected" — that was #262. But `detect_layer_type()` had no
branch for it, so it fell through to `unknown` and every seasonal subseries
plot came out as a static image with "Plot contains unsupported elements".

## What it draws

`stats::monthplot.default` runs one `lines()` call per cycle position:

```r
for (i in 1L:f) {
  sub <- phase == i
  lines((y[sub] - min(y)) * scale - 0.45 + i, x[sub], type = type, ...)
}
```

which is the shape `BaseRLineLayerProcessor` already reads for `matplot`:
one series each, every point carrying its series name as `z`.

The x coordinate on the page is a **slot offset** — each subseries squeezed
into its own 0.9-wide band so twelve fit on one axis — and it says nothing
about when a reading was taken. February's four readings sit at 1.55, 1.85,
2.15 and 2.45; a reader given those hears four Februaries that all happened
in year 2. What the offsets are computed *from* is `times`, and that is the
reading, so the processor recovers them from the recorded call the way the
correlograms recover their correlations (#276).

For a `ts`, `time()` encodes the cycle position in its fraction — February
of cycle 2 is `2.083333` — and that fraction is the phase, already announced
as the series name. `frequency` is defined so that one unit is one cycle,
which makes the floor exactly the cycle and not a rounding.

## Month names

A monthly series is named with `month.abb` rather than with `monthplot.ts`'s
own `c("J", "F", "M", "A", "M", "J", "J", "A", "S", "O", "N", "D")`.

Those letters do not name twelve things: `"M"` is March and May, `"J"` is
January, June and July. On the axis that costs a sighted reader nothing,
because the third tick is March by where it sits. **A series name has no
position to be read from**, so a reader who hears "M" cannot tell which "M"
it is. `month.abb` is base R's own spelling of the same twelve months and the
one it uses wherever a month has room to be named.

This applies only to labels this package derives. Labels the caller wrote are
passed through exactly as written, ambiguous or not, and a test pins that.

## Not read

The `base` argument draws one horizontal segment per position at
`base(x[phase == i])` — the position's mean by default — and those are not
emitted. There is no shape in the grammar for a per-series reference level,
and a series of its own would be wrong: the means run across the *cycle
positions* while every series here runs across the *cycles*, so one layer
would hold two x domains. Left to a maintainer with the grammar to change.

## Measured, not assumed

- `monthplot(ts(1:6, frequency = 12))` and friends were run and the drawn
  `lines()` calls counted. A third phase under two labels draws **two** lines,
  not three — `monthplot` loops `1L:f` where `f <- length(labels)` — so the
  undrawn phase is not read either.
- Fifty monthly readings give January and February five cycles and the other
  ten four, so the series are built point by point rather than through
  `extract_multiline_data()`, which takes a matrix and would pad the short ten
  with two points the chart never drew (the claim #170 makes about
  `geom_point`).
- A five-observation monthly series still falls back to a picture, and not
  because of this change: `monthplot` calls `lines()` twelve times regardless,
  seven of them on empty vectors, and gridSVG's own "We shouldn't be here!"
  assertion fires inside `grid.export()` — the same failure `symbols()` hits,
  already guarded by `build_interactive_svg()`.

## Verification

- 44 new cases in `tests/testthat/test-base-r-monthplot.R`.
- `monthplot` moved from the unread list to the read one in
  `test-base-r-unrecorded-calls.R`, with the upper claim asserted: it ships a
  `line` layer rather than a picture of one.
- **Mutation sweep: 15/15 caught**, after resolving four survivors — two by
  sharpening the tests (a caller who writes both `labels` and `phase`; a
  series with empty cycle positions) and one by removing a redundancy
  (`as.character()` on a whole double already writes `"2"`, so the
  integer-cast helper was doing nothing except risking `NA` past
  `.Machine$integer.max`).
- Full suite: **2241 files, 7047 passing, 0 failures, 0 errors**, 100 skips
  and the 10 pre-existing warnings.
- `roxygen2::roxygenise()` run, so `man/BaseRSubseriesLayerProcessor.Rd` is in
  the commit and the docs-drift check (#143) stays green.

---


jooyoungseo:
The bug is real — I ran it. Fixed in `7a1f7ad`, along with the minor point.
The mechanism differs from the one named, and the difference is what the fix
had to be built around, so it is worth recording.

### Measured

```
type=l  series=12  selectors=12   #graphics-plot-1-lines-1\.1 polyline
type=h  series=12  selectors=0
```

So: correct data, no highlighting, exactly as predicted.

### It is not `lines(type = "h")` — it is `segments()`

`stats::monthplot.default` branches rather than passing the `type` through:

```r
if (type != "h")
  lines((y[sub] - min(y)) * scale - 0.45 + i, x[sub], type = type, ...)
else segments((y[sub] - min(y)) * scale - 0.45 + i, means[i], ..., x[sub], ...)
```

so `"h"` never reaches `lines()`, and the grobs come out under
`-segments-`, not the `-spike-` name `BaseRSpikeLayerProcessor` overrides
for. That also bounds the problem: `monthplot`'s `type` is
`match.arg(c("l", "h"))`, so `"s"`, `"S"`, `"p"` and `"b"` are not
reachable — the function rejects them. Two cases, not six.

`find_lines_grobs` gained a `segments` branch, and the processor overrides
`selector_grob_type()` to pick it when the recorded `type` is `"h"`.

### The off-by-one that came with it

There are **thirteen** `-segments-` grobs for twelve series. `monthplot`
draws the `base` reference means in one `segments()` call *before* the loop,
so the first grob is the twelve means and the twelve after it are the
subseries. Taken as they came, every series would have been outlined on the
cycle position before it, and January on the means.

`base = NULL` would remove that grob, and cannot arrive here:

```
monthplot(M, type = "h", base = NULL)
#> Error: object 'means' not found
```

That is a real defect in `stats` — the `"h"` branch reads a `means` that is
only computed `if (!is.null(base))` — and a call that raised is never
recorded, so `f + 1` grobs is the only shape that reaches this code. The
count is checked against the series anyway and the selectors are dropped on
a mismatch rather than guessed at.

Now:

```
type=h  series=12  selectors=12
        first: #graphics-plot-1-segments-2\.1 polyline
        last:  #graphics-plot-1-segments-13\.1 polyline
```

### Also taken

The `ts`-with-written-`phase`-and-no-`labels` case you asked for is now
pinned (`monthplot(MONTHLY, phase = rep(c("odd", "even"), 24))` → two series
named after the phase, not after the months).

### Coverage

Four new cases: the spike variant's selectors exist and are `-segments-`;
the first one is `segments-2` rather than `segments-1`; and the `ts` + phase
combination. The mutation sweep grew from 15/15 to **19/19**, with four new
mutants covering exactly this — including "a spike chart is outlined one
cycle position behind" and "the base line's grob is handed over as a
subseries", both of which the old tests let through.

Full suite re-run: 2241 files, 0 failures, 0 errors.

---

Issue #272, opened 2026-08-25
`pairs()` draws a scatterplot matrix and is read as nothing


jooyoungseo:
The last of the four "exact home" readings #251 named. `pairs` is HIGH, so the call is recorded, but no processor reads it and the figure takes the static-image path.

```r
pairs(iris[1:3])
# subplot grid: 1 x 1, layers: 0
```

#251's note said "not one layer — a scatterplot matrix is n×n panels, which is a figure of subplots rather than a layer, and it is a much bigger piece than the other three." Measured, the panels turn out to be entirely determined; what is genuinely new is that **one recorded call has to produce a grid**, and nothing in the orchestrator can express that yet.

## What is recorded

One call, one argument, the data frame verbatim:

```
recorded calls: 1
  fn: pairs   class: HIGH
     <unnamed> : data.frame  dim= 150x3  names= Sepal.Length,Sepal.Width,Petal.Length
```

and **no layout call at all**:

```
layout calls: 0
low calls:    0
```

`pairs` sets its own `par(mfrow)` internally and restores it, so `detect_panel_configuration()` sees nothing and `is_multipanel_config()` is false. The grid cannot come from the recording; it has to come from the reading.

## What is drawn

`gridGraphics::grid.echo()` on `pairs()` over three columns writes **nine** `graphics-plot-N` groups — one per cell, diagonal included:

```
plot 1 : xlab, ylab, box, text                                  <- diagonal
plot 2 : xlab, ylab, box, left-axis-…,   points
plot 3 : xlab, ylab, box, bottom-axis-…, points
plot 4 : xlab, ylab, box, top-axis-…,    points
plot 5 : xlab, ylab, box, text                                  <- diagonal
plot 6 : xlab, ylab, box, points
plot 7 : xlab, ylab, box, right-axis-…,  points
plot 8 : xlab, ylab, box, points
plot 9 : xlab, ylab, box, bottom-axis-…, right-axis-…, text      <- diagonal
```

so a diagonal cell draws the variable's **name** and no marks, and every off-diagonal cell draws exactly one `points` grob. The selector is `g#graphics-plot-{k}-points-1\.1 > use`, the shape every other base R point layer already uses.

## Which cell is which

Instrumented with a `panel` that records what it was handed, on columns `a`, `b`, `c`:

```
panel draw order (first x, first y)
  (a, b)  (a, c)  (b, a)  (b, c)  (c, a)  (c, b)
```

[...]

PR #281, opened 2026-08-25
feat(base-r): read lag.plot as the grid of lagged scatters it draws

jooyoungseo:
Another of #262's twelve. `lag.plot()` is recorded — that is what #262 fixed — but `detect_layer_type()` had no branch for it, so it fell through to `unknown` and the chart came out as a static image with "Plot contains unsupported elements".

It is the only one of the twelve that draws a **grid**: one panel per series and lag. So it takes the `multi_panel` shape `pairs()` introduced (#272), for the same reason — the call sets its own `par(mfrow)` and restores it, so nothing reaches the device's layout calls and the grid has to come from the reading.

## What a panel pairs, measured

Traced through `graphics::plot.xy` on a real call. The panel is `plot(lag(X, k), X)`, and `lag()` shifts the time base **back**, so the pair at time `t` is

```
x = X[t + k]      the later reading, across
y = X[t]          the earlier one, up
```

This is the opposite of what "plotted against a lagged copy" suggests, and getting it backwards produces a chart that still looks entirely plausible — so it is asserted per pair rather than in aggregate. Zero and negative lags fall out of the same expression and were measured too: `set.lags = -1` puts the *earlier* reading across, `set.lags = 0` is the series against itself.

Panels are numbered in draw order — series outermost, lag innermost, filled row by row — confirmed against a `grid.echo()` export where a two-column matrix at `lags = 2` writes `graphics-plot-1` … `graphics-plot-4` for `(a,1) (a,2) (b,1) (b,2)`.

## Both mark shapes are outlined

`labels` decides the mark and `do.lines` only adds the joining line, so the two are read separately rather than through the default that ties them together. All four combinations measured:

| `labels` | `do.lines` | grobs in a panel |
|---|---|---|
| FALSE | FALSE | `points` |
| FALSE | TRUE | `points`, `lines` |
| TRUE | FALSE | `text` |
| TRUE | TRUE | `text`, `brokenline` |

`labels` defaults to `do.lines`, which defaults to `n <= 150` — so the **default** chart is the labelled one, and reading only the `points` case would have left most real calls without an outline.

I nearly shipped exactly that. A first probe had `gridSVG::grid.export()` failing on the text grob, which would have made "labelled panels read without an outline" look like a forced conclusion. Run through the package's own pipeline instead, the export succeeds and the labels come out as one addressable group per pair:

```html
<g id="graphics-plot-1-text-1.1">
  <g id="graphics-plot-1-text-1.1.1" transform="translate(254.32, …)">
  <g id="graphics-plot-1-text-1.1.2" transform="translate(221.70, …)">
```

— the same shape a `points` grob has, with `use` elements replaced by nested `g`s. So the selector differs only in the grob name and the child it reaches for. Both were then **resolved in a real browser** against a real `save_html()` export rather than asserted as strings: 11 and 10 elements for lags 1 and 2, matching the pairs the reading emits.

## Edge cases

- **A one-reading series is read, not declined.** `lag.plot(5, set.lags = 0)` pairs it with itself and draws a mark — measured. An earlier `nrow < 2` guard would have declined a chart that is on the page; the mutation sweep is what surfaced it.
- **A panel with no pair to make is left out of the grid.** That is what the drawing does too: `ts.intersect()` warns "non-intersecting series" and the panel comes out with its axes and no marks.
- **Nothing filters for `NA`,** because a series carrying one never reaches the drawing: `xlim` comes from `range(X)`, so one missing value makes the limits `NA` and `plot.window()` stops.
- **A non-numeric series is declined.** `lag.plot()` stops on it too — "NAs introduced by coercion", then "invalid 'xlim' value" — but the arguments of a call that stopped are recorded all the same, and coercing them would announce a grid of `NA`s that was never on the page.

## Verification

- `tests/testthat/test-base-r-lag-plot.R` — 49 assertions
- `test-base-r-unrecorded-calls.R` updated: `lag.plot` moves from the five-still-unread list to the seven-now-read list, in its **default** (labelled) spelling on purpose, so the harder half is the one exercised
- Selectors resolved in Chromium against a real export
- Full suite green
- Mutation sweep: see the comment below for the final count

---


jooyoungseo:
Mutation sweep: **26/26 caught.**

It took three rounds to get there, and the two rounds that failed are the reason the diff looks the way it does. Both survivors were real, and neither was a missing test:

**`nrow(matrix_form) < 2` survived** — removing it changed nothing any test noticed. Measuring why turned up a defect rather than a gap: `lag.plot(5, set.lags = 0)` pairs a single reading with itself and **draws a mark**, so the guard declined a chart that is on the page. Removed, with a case that asserts the point is read.

**`match.call(definition = stats::lag.plot, ...)` survived** — swapping in a different function's definition changed nothing, because `x` is the first formal with nothing ahead of it, so the matching bought exactly what a direct read buys. Replaced with a direct read of the written argument, and the two spellings that matter (`lag.plot(x = v)` and `lag.plot(lags = 1, v)`) each have a case now, along with a mutant apiece.

A third round then surfaced `ncol(...) < 1` as unreachable — the only argument `as.matrix()` turns into a column-less matrix is a data frame with no columns, and that is rejected twice already, by length and by `is.numeric`. Dropped as redundancy rather than left as a guard nothing can reach.

The `is.numeric` guard beside it stays, and is now covered by a direct case rather than left to the sweep to flag. It is not reachable through a *completed* chart either — measured, a character series makes `lag.plot()` warn "NAs introduced by coercion" and then stop in `plot.window()` with "invalid 'xlim' value" — but the arguments of a call that stopped are still recorded, and coercing them would announce a grid of `NA`s that was never drawn.

Sweep contents, for the record: the type not being recognised; the processor being swapped for the flat scatter one; the type built but not claimed by the registry; the pairing reversed; each half of the range check dropped; the panels placed column-major; a panel named after its series rather than its lag; the series name dropped; the outline dropped; a labelled panel pointed at a `points` grob; a labelled panel reaching for a `use` inside a text group; every panel answering to the first panel's marks; the `n <= 150` default; `labels` and `do.lines` each ignored; the caller's `layout` discarded; the default grid forced to one column; `set.lags` ignored; `lags` ignored; the vector name, the matrix names, and the unnamed-matrix case; both branches of the `x` lookup; and the non-numeric decline.

---

PR #284, opened 2026-08-27
docs(vignette): tell deck authors how to keep off-slide charts out of the tab order

jooyoungseo:
## Description

Adds a `## Quarto revealjs Slides` section to `vignettes/getting-started.Rmd`, after the sonification part and before the plot-type list.

A chart on a slide is already keyboard reachable on its own — Tab moves into it, the arrows explore it, Shift+Tab hands focus back to the slide, and Space advances the deck. That half is what #208 and #213 fixed, and it needs no configuration; the section says so, along with the fact that `maidr_on()` in a setup chunk is all a deck needs from this package.

What does need attention is a reveal.js behaviour that has nothing to do with MAIDR. reveal.js keeps the slides on either side of the current one rendered so transitions stay smooth, and marking them `hidden` does not take them out of the tab order, because reveal's own inline style overrides the attribute. On a deck with a chart on every slide, one Tab therefore lands on an off-screen slide's chart rather than the one in front of the reader.

Upstream this is hakimel/reveal.js#1587, open since 2016. The fix has since landed on reveal.js `master`, which marks every slide but the current one `inert`. It is not in a published release yet, and Quarto carries its own copy — Quarto 1.10 ships reveal.js 5.1.0 — so it will reach decks in some Quarto release after reveal.js cuts one, and nothing in the deck will need to change when it does.

## Changes Made

`vignettes/getting-started.Rmd`, 52 lines added, no other file touched:

- that a deck needs nothing special from the package beyond `maidr_on()`
- what the keyboard already does, and that it needs no configuration
- why one Tab can land on an off-screen slide's chart, with the upstream issue
- that reveal.js `master` fixes it, and what has to happen before that reaches Quarto
- until then, `quarto add mcanouil/quarto-revealjs-a11y` plus the YAML to enable it
- that 0.2.3 or newer is required: earlier versions parked off-slide elements at `tabindex="-1"` and could not find them again to restore them, which left the chart on the *current* slide unreachable too

No R code, no roxygen block and no generated file is touched, so nothing under `man/` changes.

## Verification

Checked in Chromium against a ten-slide Quarto `revealjs` deck with a ggplot2 chart on every slide, not reasoned about. With quarto-revealjs-a11y 0.2.3 installed, ten of the eleven slides carry `inert` and each slide gives one path in and out:

```
slide  2: chart after 2 Tab(s) -> iframe @slide-2*  | Shift+Tab -> skip-link | Space 2->3
slide  5: chart after 2 Tab(s) -> iframe @slide-5*  | Shift+Tab -> skip-link | Space 5->6
slide 10: chart after 2 Tab(s) -> iframe @slide-10* | Shift+Tab -> skip-link | Space 10->10 (last slide)
```

The 0.2.2 warning is from a measured regression, not caution: with 0.2.2 the chart on the current slide stayed at `tabindex="-1"` with the saved value still parked on it, and 42 consecutive Tab presses never reached it. Its restore query was `[tabindex]:not([tabindex="-1"])`, which cannot match anything it had itself set to `-1`. 0.2.3 replaced that mechanism with `inert` on the slide.

`rmarkdown::render()` on the whole vignette succeeds with the section in place.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Eadkcqpkod61zhhbfy8Wj8)_

### DIFF
--- a/vignettes/getting-started.Rmd
+++ b/vignettes/getting-started.Rmd
@@ -166,7 +166,7 @@ Plots can be heard through:
 - Volume changes
 - Different tones for different series
 
-## Quarto revealjs Slides
+## Quarto Reveal.js Slides
 
 A `revealjs` deck needs nothing special from this package: call
 `maidr_on()` once in a setup chunk, as in any other Quarto or R Markdown
@@ -179,7 +179,7 @@ slide, so **Space** advances the deck again. None of that needs configuring.
 What does need attention is a reveal.js behaviour that has nothing to do with
 MAIDR. reveal.js keeps the slides on either side of the current one rendered so
 that transitions stay smooth, and marking them `hidden` does not take them out
-of the tab order --- reveal's own inline style overrides the attribute. On a
+of the tab order — reveal's own inline style overrides the attribute. On a
 deck with a chart on every slide, a single **Tab** can therefore land on an
 off-screen slide's chart rather than the one in front of the reader. This is
 [hakimel/reveal.js#1587](https://github.com/hakimel/reveal.js/issues/1587),
@@ -187,7 +187,7 @@ open since 2016.
 
 The fix is now on reveal.js `master`, which marks every slide but the current
 one `inert`. It has not reached a published release yet, and Quarto carries its
-own copy of reveal.js --- Quarto 1.10 ships 5.1.0 --- so it will arrive in a
+own copy of reveal.js — Quarto 1.10 ships 5.1.0 — so it will arrive in a
 Quarto release some time after reveal.js cuts one. Nothing will need to change
 in your deck when it does.
 
@@ -216,7 +216,6 @@ With the extension enabled, each slide gives one **Tab** to its own chart and
 **Shift+Tab** back out. The extension also adds a skip link ahead of the
 slides, so Shift+Tab lands there rather than on the slide element itself;
 either way **Space** still moves to the next slide.
-
 
 ## Supported Plot Types
 
--- a/vignettes/getting-started.Rmd
+++ b/vignettes/getting-started.Rmd
@@ -166,6 +166,58 @@ Plots can be heard through:
 - Volume changes
 - Different tones for different series
 
+## Quarto revealjs Slides
+
+A `revealjs` deck needs nothing special from this package: call
+`maidr_on()` once in a setup chunk, as in any other Quarto or R Markdown
+document, and every plot the deck draws becomes an accessible MAIDR chart.
+
+A chart on a `revealjs` slide is keyboard reachable on its own: **Tab** moves
+into it, the arrow keys explore it, and **Shift+Tab** hands focus back to the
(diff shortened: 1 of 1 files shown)


jooyoungseo:
All three nits taken in `db8f169`, and the fact checked.

1. One blank line before `## Supported Plot Types` instead of two.
2. The section's two `---` are now the literal `—` the rest of the file uses.
3. Heading is now `## Quarto Reveal.js Slides`, Title Case like its neighbours. It was not deliberate — the format key stays `revealjs` in the YAML block either way, and the prose already spells the library `reveal.js`.

On the one point you could not verify: **Quarto 1.10 does ship reveal.js 5.1.0**, checked against an installed copy rather than the changelog. `quarto --version` reports 1.10.18, and the first lines of the reveal.js it carries, at `share/formats/revealjs/reveal/dist/reveal.js`, are:

```
/*!
* reveal.js 5.1.0
* https://revealjs.com
* MIT licensed
* Copyright (C) 2011-2024 Hakim El Hattab, https://hakim.se
*/
```

Grepping that bundle for `inert` returns nothing, which is the point the sentence is making. For completeness I also checked the published tarballs on npm: neither `reveal.js@5.2.1` nor `reveal.js@6.0.1` contains `inert` either, so the fix really is `master`-only so far.

`rmarkdown::render()` on the whole vignette still succeeds after the edits.

---

jooyoungseo:
`d459e68` reconciles the two rounds rather than trading one for the other: `## Quarto reveal.js Slides` — Title Case for the English words, as the previous round asked, and the product's own lower case for its name, as this round asks. The name is not heading-initial, so nothing forces a capital on it, and the heading now agrees with the body text. That is the last change I will make on casing; I have made the same edit in the sibling py-maidr PR so the two docs match.

`rmarkdown::render()` on the whole vignette still succeeds.

On the reveal.js version claim: I answered that above with the evidence — `quarto --version` 1.10.18, and the bundled `share/formats/revealjs/reveal/dist/reveal.js` opening with `reveal.js 5.1.0`, containing no `inert`. That check has been done; it does not need doing again before merge.

---

jooyoungseo:
**`macos-15-intel (release)` failed, and it is not this PR's.** It died in the `Download Quarto` setup step, before R was even asked to do anything:

```
curl: (6) Could not resolve host: quarto.org
##[error]Process completed with exit code 6.
```

DNS on that runner could not resolve `quarto.org` while fetching the Quarto installer. No R code ran, no vignette was built, no check body executed — and this PR touches one vignette's prose and nothing that could reach the runner's name resolution. The other three platforms (`ubuntu-latest release`, `ubuntu-latest oldrel-1`, `windows-latest release`) are running the same steps against the same commit.

Re-running that job once is the right move here, and I will: the API refused it just now with "This workflow is already running" because the sibling jobs are still in flight. I will re-run the failed job as soon as the run finishes, and if it fails a second time I will treat it as real and root-cause it rather than re-running again.

---

PR #287, opened 2026-08-27
docs: mark the roadmap's new layer types as experimental

jooyoungseo:
## What

Both factories claimed **14** layer types each when the coverage roadmap (#137) was filed. `Ggplot2ProcessorFactory` claims **24** now and `BaseRProcessorFactory` claims **35**. The 31 that arrived with #137 and the base R sweeps that followed (#251, #262) are not on the same footing as the ones that predate them — they landed over a few weeks, and **none has been through a user study**.

Worse, the README's "Supported plot types" tables never mentioned them at all, so a reader had **neither the list nor the caveat**. This adds an **Experimental Plot Types** section with both.

## The boundary is measured, not editorial

It is the diff of each factory's `get_supported_types()` against `8de0e98`, the last commit on `main` before #137 was filed. The stable side ends at `violin` / `candlestick`, which matches where the pre-roadmap work actually stopped.

Every experimental row names what draws it, read out of the adapters rather than recalled — `stars()` → `radar`, `assocplot()` → `residual`, `plot(type = "h")` → `lollipop`, `dotchart()` (ungrouped) → `dot`, and so on.

## What the experimental set is told not to promise

Under a `> [!WARNING]` alert, and in the same words as the core and the Python binding: under active development, unstable, no deprecation period, no support commitment, and **measured against the drawing rather than validated with a reader**. Measuring that a reading is *faithful to the chart* is a different claim from establishing that it is *useful*, and only the first has been done for these.

The section also says what would move a type out of it, which is feedback.

## The drift guard

`tests/testthat/test-plot-type-stability.R` fails if a supported type lands in **neither** set or in **both**.

One design note: the stable side is a **written-out baseline**, not a scrape. Those README tables name charts — "Bar charts", "Density/Smooth" — rather than layer types, so there is no type string in them to read. The baseline is the diff source the README names, written into the test so the check does not need a git history to run against, and the test comment says exactly that.

## Two things caught while writing it

- The prose assertions failed at first because the warning is a GitHub alert, so its lines carry a `> ` prefix that lands mid-phrase once the lines are joined. The test now strips the prefix before normalising whitespace, so a reflow is not what breaks it.
- Running the test with `library(maidr)` could not see `Ggplot2ProcessorFactory` — the R6 classes are internal. `testthat::test_local()` is the invocation that works, which is also what the package's own suite uses.

## Verified

```
Rscript -e 'testthat::test_local(filter = "plot-type-stability")'
  [ FAIL 0 | WARN 0 | SKIP 0 | PASS 17 ]

Rscript -e 'testthat::test_local(filter = "plot-type-stability|base-r-biplot|unrecorded")'
  [ FAIL 0 | WARN 0 | SKIP 0 | PASS 164 ]
```

## Companions

The same split, same boundary, same reasoning, in the core and the Python binding: xability/maidr#1208 and xability/py-maidr#687.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,84 @@ graphics with a one-time advisory.
 | Multi-panel layouts | `patchwork` package | `par(mfrow)`, `par(mfcol)` |
 | Multi-layered plots | Multiple `geom_*` layers | Sequential plot calls |
 
+### Experimental Plot Types
+
+> [!WARNING]
+> **These are prototypes. Treat them as prototypes.** They are under active
+> development, they are unstable, and **none of them has been through a user
+> study**. Field names, announcement wording and navigation may change without
+> a deprecation period, including in a patch release. If you are building
+> something that has to keep working, build it on the plot types above.
+
+Everything in the two tables above predates the plot coverage roadmap
+([#137](https://github.com/xability/r-maidr/issues/137)) and has been exercised
+by real readers over real charts. Everything below was added by that roadmap
+and the base R sweeps that followed it
+([#251](https://github.com/xability/r-maidr/issues/251),
+[#262](https://github.com/xability/r-maidr/issues/262)), most inside a few
+weeks.
+
+Each was measured against the chart it reads — that is what the issues and the
+tests record. But measuring that a reading is *faithful to the drawing* is a
+different claim from establishing that it is *useful to a reader*. Nobody has
+asked a blind or low-vision reader whether hearing `stars()` as a radar, or
+navigating a `termplot()` panel by panel, is the right way to read one. Until
+that happens these are proposals about how a chart could be read, not answers.
+
+Feedback is exactly what would move one of these into the tables above.
+
+#### ggplot2
+
+| Layer type | Drawn by |
+|-----------|----------|
+| `area` | `geom_area()`, `geom_ribbon(aes(ymin = 0, ...))` |
+| `stacked_area` | stacked `geom_area()` |
+| `stacked_normalized_area` | `geom_area(position = "fill")` |
+| `stacked_normalized_bar` | `geom_bar(position = "fill")` |
+| `contour` | `geom_contour()`, `geom_density_2d()` |
+| `error_bar` | `geom_errorbar()`, `geom_errorbarh()`, `geom_linerange()`, `geom_pointrange()`, `geom_crossbar()`, `geom_ribbon()` as a band |
+| `gantt` | `geom_segment()`, `geom_curve()` |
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Thanks — no findings, and the NEWS.md nit is one I agree with: this changes no reading behaviour, so an entry would claim more than the diff does.

Two notes on what you verified, since they were the parts I most wanted checked:

- The **per-adapter namespacing** you spotted is the right reading. `stacked_normalized_bar` appearing in both experimental tables is not a duplicate — the two factories are separate registries, and the test asserts each one against its own `get_supported_types()`.
- You were right not to trust the boundary claim you couldn't reach. `8de0e98` is not in the review sandbox, which is precisely why the stable side is a **written-out baseline** rather than a scrape: the check runs against the values themselves rather than against a git object that may not be present. The internal consistency you did verify is the part that catches real drift.

One thing carried over from the sibling reviews, for symmetry rather than because it was raised here: this test already pinned the *whole* stable set rather than a sample, which the core and Python bindings only spot-checked. Both have now been brought up to what this one does — the partition check alone is satisfied by any split that covers the registry, including a wrong one, so the boundary needs its own assertion. This PR needed no change for it.

Re-verified: `testthat::test_local(filter = "plot-type-stability")` → `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 17 ]`.

---

jooyoungseo:
CI was red on all four R CMD check platforms — my fault, and worth writing down because the local suite could not have caught it.

`testthat::test_local()` passes because it runs from the source tree, where `test_path("..", "..")` is the package root and `README.md` is right there. `R CMD check` copies the tests into `<pkg>.Rcheck/tests/` and runs them from there, so the same path resolves to the check directory and the README is not beside it. Every test in the file errored on the missing file.

This repo already has the shape for it: `skip_without_sources()` in `test-roxygen-orphans.R`, which handles the same situation — a source-consistency check that cannot run against an installed package. Followed that rather than inventing a second convention, so `skip_without_readme()` says so instead of erroring, and the check runs on `devtools::test()` and `testthat::test_local()`, which is the loop a new layer type is actually added in.

Verified against the layout that failed rather than assuming the fix works — a copy of `tests/` with no README beside it:

```
== Skipped =====
1. every ggplot2 layer type is classified as stable or experimental
   - Reason: README.md is not beside the tests under R CMD check
... 5 skips
```

Five skips naming the reason, where before there were five errors. `test_local()` still reports `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 17 ]`.

Pushed as `5cb0992`.

---

PR #288, opened 2026-08-28
feat(base-r): read a word cloud as its terms and their counts

jooyoungseo:
## What

The core has read a `word_cloud` trace since xability/maidr#796. Neither binding could emit one. This teaches r-maidr to.

`wordcloud::wordcloud()` lives in a Suggests package, so its entry point was not on the patched list. **Measured before this change:** a four-term call left the device with **zero** recorded plot calls, so `save_html()` reported "No Base R plots detected". The chart was invisible rather than misread, which makes this additive.

Wrapped on exactly the same footing as `vioplot()` — named in the `HIGH` list, reachable through the Suggests namespaces in `find_original_function`, and bound late through a `packageEvent("wordcloud", "onLoad")` hook so a call made after `library(wordcloud)` is still recorded. The `.onUnload` hook removal is mirrored too.

## The counts survive here, and that is the interesting part

`wordcloud()` takes `words` and `freq` **directly**, and both reach the recorded call intact:

```
recorded: fn = wordcloud  args: words, freq, min.freq, random.order
          words: machine, learning, data, model
          freq : 412, 300, 250, 120
```

So this reading announces occurrences, and the y axis says `Occurrences`.

The Python binding **cannot**. `wordcloud.WordCloud` divides every frequency by the largest and keeps only the ratio — measured, `{machine: 412, learning: 300}` comes back as `{machine: 1.0, learning: 0.728}`, and the raw counts are on no attribute of the object. So xability/py-maidr#688 announces a `Relative frequency` instead.

Same chart, two different honest readings, and the difference is written down on both sides rather than smoothed over.

## Filtering: replicated, and one part deliberately not

`min.freq` (default 3) and `max.words` (default `Inf`) decide **which** terms are drawn, so both are replicated — announcing a term the chart left out would name a word a sighted reader cannot find. Including `wordcloud()`'s own rule that `min.freq` drops to **zero** when it exceeds every frequency, which is what stops a cloud of rare terms coming out empty; copied rather than approximated, with a test.

**The page-fit drop is not modelled, and the reason is in the source.** `wordcloud()` also abandons a term that will not fit in the space left, warning `"<word> could not be fit on page"`. That depends on the device size and a random layout, so it cannot be decided from the arguments. A term dropped that way is still announced — the author passed it, and R told them out loud it was not drawn.

`max.words` ranks with `ties.method = "random"` upstream, so which of two equally frequent terms survives a cut between them is a coin flip *in the chart*. Ranked deterministically here; the reading can then differ from the drawing only by which of two interchangeable labels it names, never by a number.

## No selectors

Measured through the package's own export (`create_enhanced_svg()` over the orchestrator's gtable): the result carries **no `id` attributes at all**. There is no addressable element per term and no named grob to build a selector from — unlike the `text` grob that `lag.plot()` and `biplot()` write their labels as, which is what made those two highlightable. `wordcloud()` draws each term with a bare `text()` at a rotation, and nothing names those.

So the layer ships without a highlight, which the core supports: `WordCloudTrace` returns none rather than pairing the terms with whatever else resolved.

## Verified

```
testthat::test_local(filter = "base-r-wordcloud")   19 passed
testthat::test_local()                              [ FAIL 0 | WARN 10 | SKIP 100 | PASS 7312 ]
roxygen2::roxygenise()                              man/ regenerated, 2 new Rd
```

Experimental, and listed in the README table that #287's drift guard enforces — a supported type in neither the stable nor the experimental set fails that check, which is the guard doing its job on the first new type after it landed.

## Companion

xability/py-maidr#688.

---


jooyoungseo:
Both points acted on in `a058f2a`.

## The `freq == 0` edge case — measured, and there is no drift

You flagged that with `min.freq` reset to 0, a term whose count is `0` would pass `freq >= min_freq` and be announced, and asked whether upstream `wordcloud()` unconditionally skips zero-frequency terms. It does not. Its filter, in full:

```r
if (min.freq > max(freq))
    min.freq <- 0
...
ord <- rank(-freq, ties.method = "random")
words <- words[ord <= max.words]
...
words <- words[freq >= min.freq]
```

`freq >= min.freq` is the whole of it — there is no separate zero-skip anywhere in the function.

Rather than stop at reading the source, I traced `graphics::text` through a real call to see what actually reaches the page:

```
wordcloud(words = c("alpha","beta","zero"), freq = c(1,2,0), min.freq = 5)

UPSTREAM DREW: alpha, beta, zero
MAIDR READS:   beta, alpha, zero
```

All three are drawn, `zero` among them, and all three are announced. So announcing it **is** the match — the reading would have been wrong to skip it. Pinned as a test, because the correct behaviour here is the counter-intuitive one and the next person to look at that `>=` deserves to find the answer rather than re-derive it.

(The order differs because upstream defaults to `random.order = TRUE`; this layer emits heaviest-first, which the file documents and which the core re-sorts for navigation regardless.)

## The `nolint` suppression — unnecessary, removed

Confirmed exactly as you described. `.lintr` already carries:

```
object_name_linter = object_name_linter(styles = c("snake_case", "CamelCase"))
```

and with the suppression gone, all **72** `R6::R6Class(` definitions under `R/` now read the same bare way. It was noise. Dropped.

## On the one thing you could not check

You noted you had no R in the sandbox and so could not run the suite yourself. For the record, on this branch: `testthat::test_local(filter = "base-r-wordcloud")` passes 21 expectations, and the full suite is `[ FAIL 0 | WARN 10 | SKIP 100 | PASS 7312 ]`.

Your read on CI installing Suggests is right, and worth stating plainly since it was the one thing that would have made half these tests vacuous: `setup-r-dependencies@v2` with `needs: check` installs Suggests, so `wordcloud` is present and the `skip_if_not_installed("wordcloud")` guards do not fire in CI — they exist for a bare local checkout.

---

jooyoungseo:
Correcting something in my last comment: I reported the suite green on the strength of `test_local()`, and that was the wrong check. `test-coverage` was red, and chasing it found that **the feature did not work at all in an installed package** — the tests were passing on a path no user takes.

## What was actually broken

Interception for a Suggests package needs five touch points. I had four — HIGH classification, the two namespace lists, and the `packageEvent` hooks. The fifth is an **exported stub** in `base_r_wrapper_exports.R`.

`wrap_function("wordcloud")` assigns the recording wrapper into *maidr's own* namespace, so that a bare `wordcloud()` after `library(maidr)` resolves to the wrapper instead of upstream's function. But by the time `wordcloud` loads, maidr's namespace is sealed and that assignment silently no-ops. That is precisely why `vioplot` and `chartSeries` each carry a full recording wrapper as their export, with a comment saying so. I read that comment, mirrored the four mechanical touch points, and missed the one that does the work.

## Why the tests did not catch it

`load_all()` leaves the namespace **open**, so `wrap_function()` succeeds and the whole thing works from the source tree. `covr` and `R CMD check` install the package first, where it does not. So `test_local()` was structurally incapable of seeing this.

Reproduced with `R CMD INSTALL` + `test_dir(package = "maidr")`, the path CI takes:

```
before:  5 errors -- could not find function "wordcloud"
after:  23 pass
```

and the full suite against the installed package: **0 failures, 0 errors**.

Worth being plain about the user-facing consequence: as merged-ready a moment ago, `library(maidr); library(wordcloud); wordcloud(...)` would have drawn the chart and recorded nothing — silence, which is the exact failure this PR exists to remove. Good thing the coverage job runs against an install.

## Also in `c271db3`

The stub name-matches its arguments before recording. `wordcloud(words, freq)` is as natural to write positionally as by name, and the layer reads `args$words` / `args$freq`, so without `match_recorded_args()` the positional spelling would record a call with nothing in it.

And a test asserting the export exists. The drawing tests cannot see this in the mode a contributor runs locally, so the guard is the thing that makes the dev path honest:

```r
testthat::expect_true("wordcloud" %in% getNamespaceExports("maidr"))
```

I tried to demonstrate that guard failing by deleting the line from the installed `NAMESPACE`, and it did not fail — R reads the pre-built `Meta/nsInfo.rds` at load time, not the file. So the evidence for the fix is the reproduction above rather than a simulated failure of the guard itself.

---

jooyoungseo:
Recorded in `94faf53` — the reasoning is now a comment on the stub rather than something a reader has to reconstruct.

I checked the claim rather than taking it on trust, and it holds, with one correction to the aside:

```
wordcloud() is an S3 generic: FALSE
vioplot()   is an S3 generic: TRUE
```

So `dispatched_definition()`'s first-argument inspection — which is gated on `UseMethod` appearing in the target's body — never fires for `wordcloud()`. Matching after the NSE fallback is safe here, exactly as you said.

The correction: `vioplot`'s stub does not call `match_recorded_args()` at all (`grep` inside its body: 0 occurrences, vs 1 in the new one), so this is not a shared deviation — it is specific to this stub. Which makes the note more worth having, not less: `vioplot()` **is** a generic, so a future stub copied from this one for a generic target would hit precisely the case the template avoids. That is now written into the comment, with both measurements.

`R/base_r_wrapper_exports.R` re-parses cleanly and `roxygenise()` produces no Rd change — it is a comment inside a function body.

On the `pkgdown` red: that one is not this branch. It failed inside `pkgdown:::cran_link()` reaching for the CRAN sidebar link —

```
! SSL connect error [cloud.r-project.org]: Recv failure: Connection reset by peer
 11. └─pkgdown:::cran_link(pkg$package)
 12.   └─httr2::req_perform(req)
```

— a network reset on the runner, in a step that has nothing to do with the diff. Same job passed on the previous commit of this branch. I'll let the re-run on this push settle it rather than treat it as a code failure.

---

PR #289, opened 2026-08-28
docs(articles): add a word cloud example to the gallery

jooyoungseo:
## What this adds

A worked word cloud example in the article gallery (`vignettes/articles/examples.qmd`).

## Why

`word_cloud` merged in #288 with only a README table row naming it, so a reader could learn the type exists but not how to produce one.

Unlike the Python side, this is not a word-cloud-specific gap — **none** of the roadmap-added Base R types (vioplot, stripchart, spineplot, cdplot, biplot, stars, lag.plot) has a gallery section. This adds the one that was just asked for; the rest remain a known gap worth a separate pass if you want it.

## The part that matters more than the example

The same **attach-order trap** the quantmod section documents applies here, and it is the most likely way a user's cloud goes silently unread. Measured on a three-term cloud rather than assumed:

| How it is called | Recorded? |
|---|---|
| `library(maidr)` then `library(wordcloud)`, bare `wordcloud()` | **no** |
| `library(wordcloud)` then `library(maidr)`, bare `wordcloud()` | yes |
| `maidr::wordcloud()` | yes |

`library(wordcloud)` after `library(maidr)` puts `package:wordcloud` ahead on the search path, so a bare call reaches upstream directly — the chart draws, and `save_html()` then reports that no Base R plot was found. This is the visible consequence of the same mechanism behind the export stub in #288.

## Why the chunk is live rather than `eval: false`

The example calls `maidr::wordcloud()` explicitly, which works in either order **and** keeps the chunk safe to evaluate: attaching `wordcloud` mid-document would mask `maidr::wordcloud` for every later chunk, which is why the quantmod chunk is this file's one `eval: false`.

Suggests are available at render time — the `tidyquant` / `patchwork` chunk is evaluated with no guard and pkgdown passes — so `wordcloud` will be there too.

## Also stated

- The counts survive into the reading, so the axis honestly says **Occurrences**, unlike py-maidr's normalised ratio.
- The layer emits **no selectors**: `wordcloud()` draws each term with a bare `text()` at a rotation and nothing names those.
- `min.freq` / `max.words` are replicated so a term the chart left out is never announced, including `wordcloud()`'s own rule that `min.freq` drops to 0 when it exceeds every frequency.

## Verification

The chunk was run before being written down:

```
type      : word_cloud
axes      : Term / Occurrences
n points  : 6
terms     : accessibility, sonification, braille, screenreader, keyboard, contrast
counts    : 412, 300, 250, 190, 155, 120
selectors : none (by design)
```

The raw counts reaching the payload intact is what makes the "Occurrences" claim in the prose true rather than aspirational.

Vignette-only change; no R code, `NAMESPACE`, or `man/` touched.

---

### DIFF
--- a/vignettes/articles/examples.qmd
+++ b/vignettes/articles/examples.qmd
@@ -227,6 +227,78 @@ pie(shares,
 
 ---
 
+## Word Cloud
+
+A word cloud is the extreme case of a chart that carries real data while
+being readable only by eye: each term's weight is drawn as **glyph size**
+and written down nowhere on the page. Structurally it is a categorical
+label and a magnitude, so maidr reads it as a term and its number.
+
+> **Experimental.** `word_cloud` is one of the experimental layer types.
+> It has not been through a user study, and it may change without a
+> deprecation period. See the experimental table in the README.
+
+> **Note:** Word clouds are a Base R feature. ggplot2 has no word cloud
+> geom, and the packages that add one draw through their own devices
+> rather than through a layer maidr can read.
+
+> **Requires the {wordcloud} package.** Install with
+> `install.packages("wordcloud")`.
+
+#### Base R
+
+The counts you pass survive into the reading, so the weight axis honestly
+says **Occurrences**. (The Python binding cannot: `wordcloud.WordCloud`
+divides every frequency by the largest and keeps only the ratio, so
+py-maidr announces a *relative* frequency. Same chart, two different
+honest readings.)
+
+```{r}
+mentions <- c(
+  accessibility = 412, sonification = 300, braille = 250,
+  screenreader = 190, keyboard = 155, contrast = 120
+)
+
+maidr::wordcloud(
+  words = names(mentions),
+  freq = mentions,
+  min.freq = 1,
+  random.order = FALSE,
+  colors = c("#3498db", "#2ecc71", "#9b59b6", "#e67e22")
+)
+```
+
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Fixed in `12cdf01` — six colours for six terms.

You're right that it's cosmetic: `wordcloud()` recycles them itself and colour carries none of the reading. But a reader counting six terms against four colours has to stop and work out that it doesn't matter, and that's a cheaper thing to remove than to explain.

Re-verified by extracting the chunk **from the vignette itself** and running that, rather than re-running my earlier standalone copy:

```
VERIFIED: word_cloud | Term / Occurrences | 6 points | 412,300,250,190,155,120
```

On the thing you couldn't do: no problem that the sandbox blocked `Rscript` — your static cross-check against `private$terms()` and `.maidr_wordcloud_defaults` is the part I'd have wanted most, since it checks the prose against the implementation rather than against my own output. For the record, the attach-order table is the one claim that *only* a live run could establish, and it was measured in three fresh R sessions:

```
library(maidr); library(wordcloud); wordcloud(...)   NOT recorded
library(wordcloud); library(maidr); wordcloud(...)   recorded (word_cloud)
maidr::wordcloud(...)                                recorded (word_cloud)
```

That's the same search-path shadowing the quantmod section documents, and it's the reason the example is written as `maidr::wordcloud()` rather than with a `library()` call.

---

PR #290, opened 2026-08-28
fix(base-r): export the twelve wrappers that were never reachable

jooyoungseo:
## What this fixes

Twelve `HIGH`-classified calls drew their chart and recorded **nothing**:

`acf` · `pacf` · `ccf` · `cpgram` · `spectrum` · `monthplot` · `termplot` · `lag.plot` · `biplot` · `bxp` · `stars` · `interaction.plot`

Each of these has a layer processor, a README row saying it is supported, and its own passing test suite. And each was silent for a real user — the chart appeared, and `save_html()` reported that no Base R plot was found.

## Why

Interception works by **shadowing**. `.onLoad` runs `wrap_function()`, which installs the recording wrapper into maidr's *own* namespace. A user's bare `acf(x)` resolves through the search path, so it reaches that wrapper only when `package:maidr` carries the name — i.e. when maidr exports it. Unexported, these twelve resolved straight to `stats::`/`graphics::`, and maidr never saw the call.

This is the same mechanism behind the `wordcloud` export stub in #288, at twelve times the scale.

## Measured, not inferred

Against an **installed** package, across all 22 experimental Base R types, export status and being read correlate **22 out of 22**:

| | exported | reads |
|---|---|---|
| 10 types | ✅ | ✅ |
| 12 types | ❌ | ❌ (drew, zero layers) |

Then confirmed causally on one before changing twelve — adding an `acf` stub alone:

```
before:  acf → 0 layers
after:   acf → lollipop, 5 points, axes Lag / ACF
```

which is exactly what `test-base-r-correlogram.R` already expected.

After the fix, every one of the 22 reads:

```
biplot → point      bxp → box          acf/pacf/ccf → lollipop
cpgram → step       stars → radar      lag.plot → point
interaction.plot / spectrum / monthplot / termplot → line
```

## The `lag.plot` trap

Its export name has to be given explicitly. Left to infer, roxygen reads `lag.plot` as an S3 method for a `lag` generic on class `plot` and writes `S3method(lag, plot)` — which exports nothing a caller can reach. That was silently wrong until measured; there is now a comment saying so.

## The test that could have caught this

Drawing tests structurally cannot. `load_all()` leaves the namespace open **and** runs each test in an environment whose parent is that namespace, so a bare call reaches the wrapper there and only there — which is why twelve broken types had green suites.

`tests/testthat/test-base-r-interception-exports.R` asserts that every `HIGH`-classified function is exported. `getNamespaceExports()` reflects NAMESPACE in both modes, so the check reads the same from a source checkout and from an install. Verified it fails, naming the function, when one export is removed:

```
Failure: every chart-producing Base R call is exported
These HIGH-classified functions are not exported, so a bare call reaches
the original and is never recorded: acf. Add a stub to R/base_r_wrapper_exports.R.
```

It also checks each exported name still resolves to a function, so a typo in a delegation target cannot hide until the call runs.

## Verification

- All 39 `HIGH` functions are now exported; `setdiff(HIGH, exports)` is empty.
- 22/22 experimental Base R types read against an installed package.
- `roxygenise()` regenerated `NAMESPACE` and `man/base-r-wrappers.Rd`; the bogus `S3method(lag,plot)` line is gone.
- Full local suite was still running at the time of writing with **0 failures and 0 errors** so far; CI is the authority and I will confirm on this PR.

## Follow-up

The gallery still has no examples for these types — that work is what turned this up, and it is next, on top of this fix rather than under it.

---


jooyoungseo:
Both points acted on in `b5976dc`, and the first one was a real defect I'd half-fixed.

## The Rd usage line

You're right, and the interesting part is *why*: `@export lag.plot` and the Rd `\usage` generator are independent. The export tag fixed the NAMESPACE entry; roxygen's usage generator still inferred S3 formatting from the `generic.class` name pattern on its own, so the shipped man page described a dispatch that does not exist.

I fixed the half that caused the bug and missed the half that documents it. Confirmed and corrected:

```
before:  \method{lag}{plot}(...)
after:   lag.plot(...)
```

`export(lag.plot)` is intact and the guard still passes. The comment now states that neither tag implies the other, so the next person doesn't find this a third time.

## NEWS entry

Added under Bug Fixes. Your reasoning is the right one — these twelve were advertised as supported while silently doing nothing, which is precisely what someone reading NEWS wants to know changed. The entry states the mechanism, the 22/22 measurement, and why the tests couldn't see it.

## On the parts you verified by inspection

Worth recording, since you flagged that you couldn't run R: the full local suite finished at **0 failures, 0 errors**, and CI on this PR is running the same thing against an install plus four platforms.

Two of your inspection-only checks I can confirm from the other side:

- `stats::spectrum`, `stats::biplot` and `stats::monthplot` being generics — delegating to the generic does preserve dispatch, and it shows up in the readings: `biplot` comes through as `point` with the observation name on each point (`label=Alabama`), which is `biplot.prcomp`'s own output, not something a broken delegation would produce.
- The twelve mapping to the right underlying functions — each was measured individually after the fix: `acf`/`pacf`/`ccf` → lollipop, `cpgram` → step, `stars` → radar, `bxp` → box, `lag.plot` → point, and `interaction.plot`/`spectrum`/`monthplot`/`termplot` → line.

---

PR #291, opened 2026-08-28
docs(articles): add gallery examples for the experimental Base R charts

jooyoungseo:
## What

Adds gallery examples for the nineteen experimental Base R chart types that had none. The gallery covered word cloud; everything else the package reads through an experimental layer type had no worked example anywhere a user could see it.

New `## Experimental Base R Charts` section in `vignettes/articles/examples.qmd`, with 22 runnable chunks across 19 subsections: correlogram (`acf`/`pacf`/`ccf`), spectral density, cumulative periodogram, seasonal subseries, lag plot, partial effects, biplot, radar, interaction plot, box plot from summary statistics, strip chart, dot chart, lollipop, mosaic and spine plots, conditional density, association plot, filled contour, Q-Q plot, and 100% stacked bar.

## Why the prose leads with the layer type

The mapping is frequently not the one the chart's name suggests, so each section names the reading its chart actually gets:

| Call | Read as |
|---|---|
| `acf` / `pacf` / `ccf` | `lollipop` |
| `spectrum`, `monthplot`, `termplot`, `interaction.plot` | `line` |
| `cpgram` | `step` |
| `lag.plot`, `biplot` | `point` |
| `stars` | `radar` |
| `bxp` | `box` |
| `stripchart` | one `point` layer per group |
| `mosaicplot` / `spineplot` | `mosaic` |
| `cdplot` | `stacked_normalized_area` |
| `assocplot` | `heat` |
| `qqnorm` + `qqline` | `point` + `line` |

## Measured, not asserted

Every chunk was extracted from the file and executed against an installed build, and the emitted layer type, point count and axis labels recorded. The prose states what came back. That includes four places where the reading is incomplete — called out in the text rather than left for a reader to trip over:

- `cpgram()` labels only its x axis and takes no `ylab` to supply one.
- `monthplot()` has a `ylab` argument but no `xlab`, so x comes through unnamed.
- `stars()` has no x/y axes to name, so it carries no axis labels.
- `filled.contour()` given only `z` positions the grid on 0–1 and emits no axis labels at all — so the example passes real coordinates and gets `Easting (m)` / `Northing (m)`.

This is a docs-only change; no package code is touched.

## Depends on

The twelve wrapper exports from #290. Before that fix, twelve of these calls drew their chart and produced zero layers against an installed package, so examples for them would have documented something that did not work.

---

### DIFF
--- a/vignettes/articles/examples.qmd
+++ b/vignettes/articles/examples.qmd
@@ -1196,8 +1196,10 @@ cpgram(lh, main = "Luteinizing hormone: cumulative periodogram")
 so every January sits together and the seasonal shape is read directly.
 maidr reads it as a **line**.
 
-> **Note:** `monthplot()` has a `ylab` argument but no `xlab`, so the
-> reading carries the series name on y and nothing on x.
+> **Note:** `monthplot()` takes `ylab` but has no `xlab` argument, so the x
+> axis cannot be named and the reading carries nothing on x. The example
+> below passes `ylab` explicitly; left out, y falls back to the series name
+> — `nottem` here.
 
 ```{r}
 monthplot(nottem, ylab = "Temperature (F)", main = "Nottingham temperatures by month")
--- a/vignettes/articles/examples.qmd
+++ b/vignettes/articles/examples.qmd
@@ -1121,3 +1121,271 @@ p <- ggplot(mtcars, aes(x = wt, y = mpg)) +
 
 p
 ```
+
+---
+
+## Experimental Base R Charts
+
+Base R draws a long tail of statistical charts that have no ggplot2
+equivalent maidr reads — a correlogram, a periodogram, a biplot, a star
+plot. maidr reads each of them by mapping it onto a layer type it already
+knows, so a chart drawn for the eye becomes a series a reader can walk.
+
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Taking the nit — it is the same trap that got caught on py-maidr #689, where a paragraph explained a default that the chunk beneath it overrode, so a reader running the example never saw the thing being described.

Fixed in `8da24c7`. The note now leads with the x axis, which is the actual limitation (`monthplot()` has no `xlab` argument at all), and states the y fallback as the case it is rather than as what the example does:

> **Note:** `monthplot()` takes `ylab` but has no `xlab` argument, so the x axis cannot be named and the reading carries nothing on x. The example below passes `ylab` explicitly; left out, y falls back to the series name — `nottem` here.

Both halves are measured:

```
monthplot(nottem)                          line n=12  y=nottem
monthplot(nottem, ylab = "Temperature (F)") line n=12  y=Temperature (F)
```

Worth adding for the record, since it came out of the same measurement pass: `monthplot(nottem, xlab = "Month")` produces a layer with **zero** points — labelling the x axis discards the whole series. That is a package bug rather than a docs one, so it is filed separately as #292, and it is why this example passes only `ylab`.

---

PR #293, opened 2026-09-02
fix: repository review -- fallback recursion, phantom replays, and twelve readings that disagreed with the drawing

### LINKED ISSUE #292: monthplot(xlab = ...) draws the chart but empties the layer
## Summary

`monthplot(x, xlab = "...")` draws the chart but produces a layer with **zero data points**. The identical call without `xlab` produces the expected 12. The chart is silently reduced to nothing to read.

`xlab` is not a formal of `stats:::monthplot.default`, so it reaches the function through `...`:

```r
names(formals(stats:::monthplot.default))
#>  [1] "x"    "labels" "ylab" "times" "phase" "base" "axes" "type"
#>  [9] "box"  "add"    "col"  "lty"   "lwd"   "col.base" "lty.base"
#> [16] "lwd.base" "..."
```

`ylab` **is** a formal, and passing it is fine. So the two arguments a caller would reach for to label their axes behave completely differently: one works, the other empties the reading.

## Reproduce

```r
library(maidr)
probe <- function(label, draw) {
  pdf(NULL); d <- dev.cur(); maidr:::clear_device_storage(d)
  on.exit(invisible(dev.off()), add = TRUE)
  draw()
  l <- maidr:::BaseRPlotOrchestrator$new(d)$generate_maidr_data()$subplots[[1]][[1]]$layers
  cat(sprintf("%-12s type=%s n=%d\n", label, l[[1]]$type, length(l[[1]]$data)))
}

probe("bare",      function() monthplot(nottem))
probe("ylab only", function() monthplot(nottem, ylab = "Temperature (F)"))
probe("xlab only", function() monthplot(nottem, xlab = "Month"))
probe("main only", function() monthplot(nottem, main = "M"))
```

## Observed

```
bare         type=line n=12
ylab only    type=line n=12
xlab only    type=line n=0     <-- data gone
main only    type=line n=12
```

The layer is still emitted, still typed `line`, and still carries the x axis label that was passed — it just has no points in it. So nothing errors and nothing warns; the chart draws normally and the reading is empty.

## Expected

`n=12` in all four cases. Labelling an axis should not discard the series.

## Impact

This is the same failure shape as #262: the chart looks right, so the loss is invisible unless someone inspects the emitted layer. It is reachable by the most ordinary thing a user does to a chart — naming its axes — and `xlab` is the argument they would try first, since every other Base R plotting function takes it.

## Notes

Found while measuring readings for the experimental Base R gallery examples (#291). The gallery works around it by passing only `ylab`, and says in the text that x comes through unnamed — that note should be removed once this is fixed.

Worth checking whether other wrapped functions that take `...` rather than an explicit `xlab` share the behaviour; I only measured `monthplot`.

jooyoungseo:
## What this is

A whole-repository review, with the defects it turned up fixed where each fix
was small, verifiable and testable, and the rest written down below. Every
fix here was **reproduced first** against the current `main` (R 4.3.3,
ggplot2 4.0.3, roxygen2 8.1.0) and has a regression test that fails without
it.

The commits are one logical change each; the PR is meant to be squashed.

### Fallback rendering

- **A ggplot that maidr can read but cannot export got neither the chart nor
  the picture.** `create_fallback_image()` drew the fallback with a bare
  `print(plot)`, which dispatches to `maidr_print_ggplot()`, which builds the
  chart again, fails again, and asks for another picture -- one nested
  `png()` device per round. Measured: 38 build attempts, then
  `"Failed to create fallback image"`. The picture is now drawn with
  ggplot2's own method (`print_ggplot_natively()`), as `show()`'s native
  path is too.
- **The Base R fallback recorded its own replay.** `do.call("barplot", …)`
  resolved the name in maidr's namespace -- the recording wrapper -- so the
  picture's replay was logged against the throwaway `png()` device, whose
  number R reuses for the next device opened. Measured: after one fallback,
  `get_device_storage_summary()` held a second device with a phantom call.
  The replay now goes through `replay_plot_call()`, which also restores the
  `call_env` an NSE call is evaluated in.
- **`maidr_set_fallback(format = "svg")` was validated, stored and read by
  nothing.** `create_fallback_html()` and `create_inline_image()` now default
  to the option.

### Base R readings

- **`monthplot(x, xlab = "Month")` emptied the series (#292).** `args$x`
  partial-matched `xlab`. Read through `resolve_xy_args()` like every other
  base R reading; the gallery example names both axes now instead of saying
  x cannot be named. Closes #292.
- **Box plot outlier selectors were off by one per outlier-free box before
  them.** `bxp()` emits a `points` grob only for boxes that have outliers;
  the segment selectors already applied that shift, the points selector did
  not. Measured on `boxplot(list(a, b, c))` with `a` clean: emitted
  `points-4`/`points-6`, drawn `points-3`/`points-5`.
- **A recorded formula `subset` was ignored.** `recorded_formula_frame()`
  built the frame without it, so `stripchart(len ~ supp, data = tg,
  subset = tg$dose == 0.5)` announced 30 points per group against 10 drawn.
  A plain vector is applied; an unevaluated expression declines the frame.
- **`plot(<ts>)` was typed `point` with a selector on a grob that was never
  drawn.** `plot.ts` defaults to `type = "l"`; the adapter now knows.
- **`main = expression(alpha^2)` failed the whole save** in a dozen readers
  (`jsonlite` has no method for an expression). One shared
  `recorded_main_title()` replaces twelve unguarded copies.
- **`lines(numeric(0))` after a plot aborted the render** (`1:n` with
  `n = 0`).
- **Wrapped calls returned invisibly no matter what the original did**, so
  `par("mar")` and `hist(x, plot = FALSE)` printed nothing once maidr was
  attached. The generic wrapper now keeps the original's visibility via
  `withVisible()`; every drawing call is invisible in base R already, so
  nothing new auto-prints in knitr.
- A `NULL` layer title on the single-panel path serialised as `{}`; it now
  defaults to `""` like the multipanel path.

### ggplot2 readings

- **Reading a horizontal bar whose orientation is set on the layer rewrote
  the caller's plot.** A ggplot2 layer is a ggproto environment;
  `plot$layers[[i]]$mapping <- …` wrote through to the user's object, and
  `geom_col(aes(y = g, x = n))` came out vertical -- in the reading and in
  the user's next `print()`. The swap now lives on a `ggproto(NULL, layer,
  mapping = …)` child.
- **An empty facet panel (`facet_wrap(~ g, drop = FALSE)`) was handed every
  other panel's rows.** `get_layer_built_data()` returns the (possibly
  empty) panel subset now.
- **Faceted and patchworked layers were handed the geom class as
  `layer_info$type`**, where the single-plot path passes the detected type,
  so a faceted `position = "fill"` bar announced counts (measured: `2 2 1 1`
  against `0.67 0.67 0.33 0.33` unfaceted).
- **A patchwork leaf dropped `orientation`, `stepDirection` and
  `domMapping`** unless it was an expanded violin result. The carry is
  unconditional now, as it is on the facet path.
- **A box plot ignored `scale_x_discrete(labels = )`** and announced the raw
  level while the line reader on the same panel announced the label.
- `is_flipped_layer()` no longer errors on a `NULL` layer index.

### Infrastructure and hygiene

- `reset_global_registry()` failed on an installed package ("cannot change
  value of locked binding"); the registry lives in a package-private
  environment now, and the test that only checked the function *existed*
  checks that it works.
- `readLines()` of the SVG and the bundled assets passes
  `encoding = "UTF-8"`.
- The four `R CMD check` Rd NOTEs on `main` ("Lost braces" from
  `{@code}`/`{@link}`, an escaped `\#`) are gone.
- `Depends: R (>= 4.0.0)` -- the code already uses `str2lang()` (3.6) and
  `marginSums()` (4.0), inside `tryCatch`, so on 3.5 those readers degraded
  silently rather than failing.
- `get_supported_types()` lists `violin` and `pairs`, which the factory
  dispatches and the adapter emits; README gains the matching experimental
  rows and a `contour()` row for the stable base R type it was missing.
- README pointed at `vignette("plot-types")`, which does not exist; it
  points at the examples gallery.
- `use_cdn = NULL` was documented as "auto-detect" for `show()` and
  `save_html()`, which have defaulted to bundled files deliberately since
  the CDN-scaffold bug; the docs, README and vignette now say what the code
  does, including the `lib/` folder `save_html()` writes.
- Test hygiene: `setwd()` restored with `on.exit`, an option restored rather
  than set, a permanently-skipped `test_that()` removed.

## Verification

- Full `testthat` suite locally (R 4.3.3, ggplot2 4.0.3): **8365 passed, 0
  failed, 3 skipped**. Baseline on `main`: 8317 passed, 0 failed.
- `roxygen2::roxygenise()` with the pinned 8.1.0 regenerated `man/`; the
  docs-drift job should stay green.
- `lintr` reports nothing on the changed lines.
- `R CMD check --no-tests --ignore-vignettes` locally: `checking Rd files
  ... OK` (the four NOTEs on `main` are gone); the remaining WARNING and
  NOTEs are the container's missing `en_US.UTF-8` locale, three
  uninstalled Suggests, and installed size.

## Found, verified, and deliberately not fixed here

Each of these is real and reproduced, but changes behaviour that a maintainer
should decide on, or is larger than a review PR should carry. Filed here so
they are not lost:

1. **`barplot()` re-sorts the user's bars alphabetically before drawing
   them.** `SortingPatcher` reorders a named vector or matrix by name, so
   `barplot(c(Jan = 3, Feb = 5, Mar = 2))` draws Feb, Jan, Mar -- on screen,
   not just in the reading -- and per-bar `col`/`density`/`legend.text` are
   not reordered with it. `test-patch-architecture.R` asserts the sort and
   NEWS records it as intended, so it needs a decision, not a patch.
2. **Two high-level plots on one single-panel device** produce one JSON with
   both plots' layers over the SVG of the first (`hist(x); plot(y); show()`).
   The fix interacts with the knitr per-figure hook, so it needs design.
3. **`plot(y ~ g, data = df, subset = g == "a")` fails with "object 'g' not
   found"** through the wrapper, on `main` as well; `boxplot()`'s formula
   path has the caller-frame retry and `barplot`/`lines`/`points` do not.
4. `knit_print` methods are registered by hand from `maidr_on()` rather than
   via `S3method()`; `maidr_print_ggplot()` does not check
   `knitr.in.progress`.
5. The offline widget/knitr iframe inlines the 1.5 MB bundle into a base64
   `data:` URL that is ~2 % under Chromium's 2 MB URL limit; #275 (srcdoc)
   is the fix and this PR does not touch it.
6. `inst/COPYRIGHTS` / `Authors@R` do not list every bundled JS component
   (`immer`, `marked`, `@reduxjs/toolkit`, `@mui/material`, `@emotion`);
   `cran-comments.md` still names bundle 3.75.1 and has unfilled
   placeholders; the NEWS development section has three `## Bug Fixes`
   headings and the version is not bumped to `0.4.0.9000`.
7. The three structural tests (`test-roxygen-orphans.R`,
   `test-plot-type-stability.R`, part of `test-format-utils.R`) skip under
   `R CMD check` and no workflow runs `testthat::test_local()`.
8. `barplot(beside = TRUE, height = m)` (`beside` first) is typed `bar` and
   flattened; five sites read slot 1 instead of `height`.
9. `spectrum()`/`cpgram()` selectors hard-code panel 1; `abline()` extents
   use a 5 % pad where R uses 4 % and ignore `xlim`/`ylim`; the ggplot2
   heatmap reorder factors columns by position rather than mapping; a
   faceted `geom_rug()` with both sides is dropped by the facet combiner;
   `detect_layer_type()` triggers a full `ggplot_build()` per call.
10. Dead code worth removing in a cleanup: the pre-`_doc` candlestick SVG
    helpers, `create_maidr_widget_internal()`, `create_fallback_iframe()`,
    `r_date_format_to_js_function()`, `base_r_state_tracking.R`'s unread
    state, and duplicated rect/polyline walkers across the bar family.

https://claude.ai/code/session_018txuh4mV45nGJY69qBAt9q

---
_Generated by [Claude Code](https://claude.ai/code/session_018txuh4mV45nGJY69qBAt9q)_


jooyoungseo:
Both review points are addressed in a121772:

- `contour()`'s `extract_main_title()` goes through `recorded_main_title()` now, and the assocplot/mosaic inline copies of the same guard are collapsed into it. Added a test: `contour(volcano, main = expression(alpha^2))` reads as a `contour` layer with an empty title instead of failing the save.
- The relative-path `save_html()` test unlinks by `file.path(tempdir(), relative_path)`, since the earlier `on.exit()` has already restored the working directory when the cleanup runs.

Local run of the touched files (`test-base-r-main-title.R`, `test-base-r-contour.R`, `test-base-r-assocplot.R`, `test-base-r-mosaicplot.R`, `test-save-html.R`, `test-contour-level.R`): 0 failures.

---

PR #299, opened 2026-09-06
docs: split the examples gallery by plot family and add canonical, entity and citation metadata to the site

jooyoungseo:
## Description

SEO and generative-engine (Google AI Overviews / AI Mode) optimization of https://r.maidr.ai, following Google's guide [Optimizing your website for generative AI features on Google Search](https://developers.google.com/search/docs/fundamentals/ai-optimization-guide) and the [Googlebot](https://developers.google.com/search/docs/crawling-indexing/googlebot) documentation. Google's position: AI features need nothing beyond standard SEO, pages must be indexed and snippet-eligible, structured data must mirror visible text, and llms.txt neither helps nor hurts.

Googlebot fetches only the first **2 MB** of an HTML file. The live examples article was 9.27 MB, so seven of its eighteen sections (heat map, KDE, regression, multi-layer, multi-panel, facets and all 22 experimental Base R charts) were never indexed. Two things made it that big: gridSVG's inline `gridSVGCoords` / `gridSVGMappings` scripts (about 25% of every chart, read by nothing) and one `filled.contour(volcano)` example that alone was 3.3 MB.

### Changes

- **`R/svg_utils.R`:** `grid.export()` no longer inlines the gridSVG coordinate and mapping tables. Nothing in `maidr.js`, this package or its tests reads them (selectors are built from element ids), so every `show()`, `save_html()`, knitr and Shiny chart gets smaller. NEWS entry under Performance.
- **Examples split:** `vignettes/articles/examples.qmd` is now a hub at the same URL (keyboard table, "When to use" table, links). The 58 examples moved verbatim into seven family articles (bar/pie, distributions, scatter/line, heat map/candlestick, multi-panel/facet, Base R time-series diagnostics, experimental Base R), each with a title, `description:`, and a short intro stating how the family is navigated and whether it is stable or experimental. The volcano example is downsampled to every third row and column. `_pkgdown.yml` switched to an `articles:` config so both the navbar dropdown and `articles/index.html` list them.
- **robots.txt:** `pkgdown/extra/` is never copied by pkgdown (only `pkgdown/assets/` is), so https://r.maidr.ai/robots.txt was a 404. Moved to `pkgdown/assets/robots.txt`. The hand-written `llms.txt` there was also never published and would be overwritten by pkgdown 2.2's generated one, so it is removed.
- **Descriptions and titles:** `description:` on every article (the two Rmd vignettes also had a pkgdown quirk that dropped their og:image and twitter tags without it); the dynamic `date:` lines are gone; home title no longer repeats the brand; sidebar links to the MAIDR core and py-maidr.
- **`<html lang>` (r-lib/pkgdown#2981):** the site already emits `lang="en"` because pkgdown reads the top-level `lang:` field in `_pkgdown.yml` (`get_pkg_lang()`: `_pkgdown.yml` `lang`, then `DESCRIPTION` `Language:`, then `"en"`; `layout.html` writes `<html lang="{{{lang}}}">`). That answers the issue's request for a `template.lang` option. This PR keeps `lang: en` and adds `Language: en` to DESCRIPTION as the documented fallback.
- **Structured data:** JSON-LD moved out of the site-wide include into `pkgdown/templates/in-header-home.html` (home page only, whisker-rendered) with the entity graph shared across maidr.ai, py.maidr.ai and r.maidr.ai: the (x)Ability Design Lab, the R software node with `isBasedOn` the core, the two MAIDR papers by DOI, WebSite, and a WebPage with `relatedLink` to the sibling sites. No `softwareVersion` (main carries a dev version).
- **Post-build step (`tools/seo-postprocess.py`, stdlib only):** pkgdown has no hook for canonical links, so the workflow adds `<link rel="canonical">` and `og:url` to every built page, `noindex` to `404.html` and to reference topics whose Rd carries `\keyword{internal}` (345 of 370 reference pages are orphaned internal topics), descriptions to the pages pkgdown leaves without one, and prunes `404.html`, `LICENSE.html` and the noindexed pages from `sitemap.xml`. A following step fails the build if any page exceeds 1,900,000 bytes and prints the ten largest.
- **README and citation:** the Overview names the lab, MAIDR and py-maidr; new "Related projects" and "Citation" sections; `inst/CITATION` so `citation("maidr")` returns the CHI 2024 and EuroVis 2024 papers plus the package entry. The stability tables are untouched (the stability test parses them).

### Notes for reviewers

- R is not available in the sandbox, so pkgdown, `R CMD check` and testthat could not be run locally; the pkgdown workflow builds this PR (without deploying) and its "Ten largest HTML files" output is the first real measurement of the family pages. The experimental Base R page has the least headroom; if it lands above about 1.6 MB, the fix is a further split, not a threshold change.
- `_pkgdown.yml`, the workflow and all vignette YAML were parsed with PyYAML; the post-process script was run twice on a synthetic pkgdown site (idempotent); `inst/CITATION` was checked by reading.
- Left as is: the reference index still lists an "Internal utilities" section whose pages are now `noindex`; hiding them from the index (`title: internal`) is a content decision for the maintainers.

https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn

---
_Generated by [Claude Code](https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn)_


jooyoungseo:
Thanks for the review, and for independently checking the two claims the PR rests on (nothing reads `gridSVGCoords` / `gridSVGMappings`, and the seven family articles account for all 17 sections).

**`inst/CITATION` hardcoded year: fixed.** The package `bibentry` now uses `format(Sys.Date(), "%Y")`, the `usethis::use_citation()` pattern, so it cannot go stale. The two paper entries keep their literal `2024`, which is the year those papers were published rather than a value that ages.

**First real measurement of the split, from this PR's pkgdown run** ([job](https://github.com/xability/r-maidr/actions/runs/34002502883/job/101403768381)). The build, the post-process step and the size guard all passed:

```
seo-postprocess: 386 pages processed, 67 skipped (not pkgdown or redirect),
386 canonical links added, 345 noindex tags added, 13 descriptions added,
sitemap 386 -> 40 URLs

Ten largest HTML files:
1282168 docs/articles/examples-base-r-experimental.html
 684003 docs/articles/examples-multi.html
 462856 docs/articles/examples-scatter-line.html
 458485 docs/articles/examples-distribution.html
 441431 docs/articles/examples-base-r-timeseries.html
 326484 docs/articles/examples-bar-pie.html
 261065 docs/articles/examples-heatmap-candlestick.html
  72064 docs/reference/Ggplot2ErrorbarLayerProcessor.html
```

The one page you and I both flagged as having the least headroom, `examples-base-r-experimental.html`, came in at 1.28 MB, about 800 KB under Googlebot's cutoff, so no further split is needed. The whole gallery was 9.27 MB on one URL before this; the largest page is now 1.28 MB and every section is inside the first 2 MB that Googlebot reads.

The sitemap going from 386 to 40 URLs is the internal reference topics (345 Rd files carry `\keyword{internal}`) dropping out along with `404.html` and `LICENSE.html`, leaving the 25 public topics, the articles, news and authors.

---

jooyoungseo:
Both nits fixed in `ecf97e7`, and the first one is worth more than tidiness.

**CRAN URL casing.** You're right that the hostname is case-insensitive and both resolve, but `sameAs` is specifically an identity signal, and a consumer comparing it against the `downloadUrl` two lines below would be comparing two strings that differ. Making a reader normalize the host before it can see they match is exactly the kind of friction the field exists to remove. Both now use the lowercase form, matching the pre-existing `downloadUrl`, and I asserted no mixed-case CRAN host remains anywhere in the file rather than just fixing the one line I had written.

**Array layout.** Both new arrays are now one URL per line, matching the Organization and Person nodes.

**On the parsing you could not run:** your manual brace-balance check was correct, and I verified it with a parser on this end. I extract the `application/ld+json` block and run it through `json.loads` after every edit to this file, then assert the node count and that the first line is still exactly the whisker include, since pkgdown's `in-header-home` template is whisker-rendered and a shifted first line would break the include. Nine nodes, valid JSON, first line intact after this change.

---

PR #300, opened 2026-09-06
docs(seo): give reference managers real bibliographic metadata

jooyoungseo:
## Description

Saving any page of r.maidr.ai to Zotero produced a Web Page item titled `<page> • maidr`, with no author and no date. The site emitted only Open Graph and Twitter tags, and Zotero's Embedded Metadata translator looks for Highwire Press `citation_*` first, then Dublin Core, then Open Graph — finding neither of the first two, it fell back to the `<title>`.

### Changes

**`tools/seo-postprocess.py`** now adds a Dublin Core block to every page it already touches: title with the pkgdown site-name suffix stripped, one creator tag per author, publisher, description, canonical identifier, type, format, language, rights, and a date taken from the file the page was built from — `README.md` for the home page, the vignette for an article, the Rd for a reference topic, `NEWS.md` for the changelog.

- **Creators are surname-first** (`Seo, JooYoung`) so Zotero splits them into first and last names rather than storing one opaque string. They match `Authors@R` in DESCRIPTION.
- **`DC.type` is per page**: the home page is `Software` because it stands for the package; articles and reference topics are `Text`. Claiming `Software` site-wide would tell a reference manager that a reference topic is the program.
- **The site name is read from `og:site_name`** and only an exact trailing `<separator><site name>` is removed, so a reference topic rendered as `Show a plot — show • maidr` keeps its em dash and loses only the suffix. An earlier version partitioned on the first separator and would have filed it as `Show a plot`.
- **Redirect stubs keep being skipped** by the existing `http-equiv="refresh"` guard, so a 262-byte stub does not gain a metadata block.
- **Git lookups are cached** by source path and the history check by root. The site has 386 pages and many share a source, so uncached lookups would shell out to git once per page.

**`.github/workflows/pkgdown.yaml`** checks out with `fetch-depth: 0`, and the script refuses to guess when it cannot: `has_history()` treats a single-commit checkout as unusable and omits `DC.date` entirely rather than stamping every page with the build date. See the correction below for why.

**`CITATION.cff`** gives GitHub its "Cite this repository" button alongside the existing `inst/CITATION`, with authors matching `Authors@R`. It is listed in `.Rbuildignore` so `R CMD check` does not report it as a non-standard top-level file.

### Deliberately no `citation_*` tags

Google Scholar's [inclusion guidelines](https://scholar.google.com/intl/en/scholar/inclusion.html) reserve those tags for scholarly articles, require that a site consist "primarily of scholarly articles", and say explicitly that `citation_title` must not carry the name of a repository. Putting them on package documentation misrepresents the page and risks the site being dropped from Scholar, without gaining indexing — Scholar does not index documentation sites. The two MAIDR papers stay cited by DOI in the home page's JSON-LD, in `inst/CITATION` and in the visible citation section, and reach Scholar through ACM and Eurographics. The reasoning is recorded in the module docstring so it is not quietly undone later.

### Correction to an earlier version of this description

An earlier revision of this text claimed the source-date mapping was "confirmed" because `reference/show.html` resolved to `man/show.Rd`'s own commit date rather than the build date. **That verification was invalid.** It ran against this sandbox's clone, which carries 53 commits, and never exercised the environment the workflow produces: `actions/checkout@v4` defaults to depth 1, where the single commit has no parent, so git reports it for every tracked path and every page would have carried the build date while looking correctly per-page.

The reviewer caught it. I reproduced it against a real depth-1 clone before fixing it, and the fix is both the workflow change and the in-script guard above, so re-shallowing the checkout later cannot silently reintroduce it.

### Notes for reviewers

- R is not available in the sandbox, so `R CMD check` and pkgdown could not be run locally; the pkgdown workflow builds this PR and its output is the first real measurement. **Two different `DC.date` values on two pages in that build is the proof the fix works in CI**, and is worth a look before merge.
- Verified on a synthetic pkgdown site: the home page emits `DC.type` `Software` with the suffix stripped; a reference topic emits `Text` and keeps an em dash inside its own title; a page titled only `maidr` survives; a page with no `og:site_name` keeps its whole title; a redirect stub is untouched; a second run changes nothing.
- With history present, `index.html` dates from `README.md` and `reference/show.html` from `man/show.Rd` (`2026-09-06` vs `2026-09-01`). Inside a single-commit checkout the block is emitted with no `DC.date` and a warning naming the fix.
- No tests for `tools/*.py`: this repository has no Python test infrastructure at all (`tests/` is testthat, `tools/` is in `.Rbuildignore`). Adding pytest means a dependency, a CI job and a convention, which should be its own change. The same logic is covered by 35 cases in the sibling xability/py-maidr#770.
- `ruff check` clean. I left the two pre-existing lines `ruff format` wanted to rewrap alone, since they are unrelated to this change.

https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn


jooyoungseo:
You were right, and this was a real defect rather than a theoretical one. Fixed in `1c411f7`.

**Reproduced before fixing.** I cloned a repository at `--depth 1` and ran the same lookups: every path reported the checked-out commit's date, where full history gives them different dates. Your explanation of why is exactly right — the single commit has no parent, so git treats every tracked path as added by it.

**My verification claim was wrong, and you named the reason precisely.** I wrote that `reference/show.html` resolves to `man/show.Rd`'s own commit date "confirming per-page dates rather than one shared timestamp", and that was checked against this sandbox's clone, which carries 53 commits. It never exercised the environment the workflow produces. The claim was true of what I ran and false of what ships, which is the worst shape for a verification claim to have.

**Both of your fix options, not one.** `fetch-depth: 0` on the checkout step, and the script now refuses to guess: `has_history()` checks the commit count once, cached, and `source_date()` returns empty when there is only one commit, so `DC.date` is omitted entirely and a warning prints on stderr. The workflow fix alone would leave the same silent failure waiting for whoever edits that checkout step next.

One refinement to your suggestion. I test the commit count rather than `--is-shallow-repository`, because they turn out to be different questions: this sandbox's clone reports `is-shallow: true` **and** has 53 commits, so it dates pages correctly. Keying on shallowness would have thrown away good dates. Only the single-commit case is actually unusable.

Verified on a synthetic pkgdown site: with history, `index.html` dates from `README.md` and `reference/show.html` from `man/show.Rd` (`2026-09-06` vs `2026-09-01`) — two different values on two pages, which is the check you asked for; run from a single-commit checkout, the block is emitted with no `DC.date` and the warning prints. A second run still adds nothing.

**On your other notes:**

- **Three copies of the author list** (`CREATORS`, `Authors@R`, `CITATION.cff`) — a fair objection, and I left it as three. `CREATORS` is surname-first for Zotero's name splitting, `Authors@R` is R's person objects with roles and emails, and CFF wants split given/family fields; deriving one from another means parsing `Authors@R` from Python at build time, which trades a visible duplication for a parser that fails at site-build time. Worth revisiting if a fourth copy appears.
- **No tests for `tools/*.py`** — accepted, and your point that a script-level test would have caught this one before merge is well taken; the shallow-versus-full distinction is exactly what such a test would encode. I'd rather add that as its own change than grow this PR further.
- Thank you for cross-checking `CITATION.cff` against `inst/CITATION` and `Authors@R`, and for confirming the `.Rbuildignore` anchoring and the SPDX identifier. Those were the parts I could not verify without an R toolchain here.

---

jooyoungseo:
Thanks for confirming the shallow-clone fix traces through cleanly, and for checking the cache keys.

**On the title stripping — I changed my mind and fixed it, in `33f8d69`.** In my previous reply I said I would note it rather than fold another change in. That was the wrong call: the py-maidr reviewer raised the same thing independently, and a reference topic rendered as `Show a plot — show • maidr` would have been filed as `Show a plot`, losing the half that identifies which topic it is. That is not a cosmetic edge case, and the fix was cheap.

It now removes only an exact trailing separator-plus-site-name, reading the site name from the `og:site_name` pkgdown already emits rather than guessing where a suffix begins. Verified on a synthetic site:

| rendered title | recorded |
|---|---|
| `Show a plot — show • maidr` | `Show a plot — show` |
| `Accessible ggplot2 and Base R Plots • maidr` | `Accessible ggplot2 and Base R Plots` |
| `maidr` | `maidr` |
| `A – B` (no `og:site_name`) | `A – B` |

This also brings the three sites into line — maidr and py-maidr strip by exact suffix for the same reason, and describing one set of entities across them is the point of the whole series.

**On tests for `tools/*.py` — you are right that this file has now demonstrated exactly what tests are for, and I have not added them here.** The reason is specific rather than a dodge: this repository has no Python test infrastructure at all. `tests/` is testthat, and `tools/` is in `.Rbuildignore`. Adding pytest means adding a dependency, a CI job and a convention to an R package, which is a change that should stand on its own rather than ride in on a metadata PR.

Where the infrastructure already existed I did take the point: the sibling py-maidr PR now has 27 cases covering exactly the split you describe — a temp repo with one commit versus several, asserting `DC.date` is omitted versus present. That is the same logic as here, so the behaviour is pinned even though this copy is not.

**On the three copies of the author list** — same answer as before, but with a working example now: the py-maidr PR added consistency tests tying its constants back to `pyproject.toml`, so a future author change fails a test. Doing that here needs the same Python test infrastructure, so it belongs with it.

**On your pre-merge check** — agreed, and it is the right last look. The pkgdown run on this commit is the first real one, and two different `DC.date` values on two pages is the proof that history is present. I will read that from the build before merging rather than assume it.

---

jooyoungseo:
**The pkgdown build confirms the shallow-clone fix works in CI.** This is the check you asked for before merge.

```
seo-postprocess: 386 pages processed, 67 skipped (not pkgdown or redirect),
386 canonical links added, 345 noindex tags added (345 pages noindex in total),
13 descriptions added, 386 Dublin Core blocks added, sitemap 386 -> 40 URLs
```

Two things to read out of that run:

**386 Dublin Core blocks added.** The script ran over the real built site and tagged every page it processes, so nothing about the block emission depends on my synthetic fixtures.

**No `single-commit checkout` warning in the log.** That is the direct evidence. `has_history()` prints that line to stderr whenever the commit count is 1, and the run is silent, so the checkout carried real history and `source_date()` actually performed the per-page git lookups rather than short-circuiting. Before `fetch-depth: 0`, that warning would have fired on every page — or, without the guard, the build would have silently stamped all 386 pages with the same date and looked exactly as healthy as this one does.

The one thing this log cannot show is the dates themselves, since a PR build does not deploy and I have no artifact to read the HTML back from. So the proof is by absence of the warning rather than by inspecting two `DC.date` values. If you would rather see the values, the post-merge deploy will publish them and I can pull two pages from the live site — happy to do that as a follow-up check rather than assert it now.

The other counts line up with the existing behaviour: 345 internal reference topics noindexed, the sitemap pruned from 386 URLs to 40, and 67 redirect stubs skipped without gaining a metadata block.

Current state: 12 of 14 checks green, with the macOS and Windows `R CMD check` jobs still running. Nothing red.

---

PR #307, opened 2026-09-15
feat: download the DotPad SDK and carry it with offline documents

jooyoungseo:
## What

maidr.js drives a DotPad tactile display through the vendor's SDK, which it imports from jsDelivr on first connect rather than bundling its 14 MB braille engine. That was the one path a `use_cdn = FALSE` document still took to the network, unless the session named a served copy through `maidr.dotpad_sdk_url`. Dot Inc. have confirmed MAIDR may redistribute the SDK, so the package can now carry a copy on request:

- **`maidr_download_dotpad_sdk()`** (exported) fetches the commit maidr.js is pinned to (`dotincorp/dotpad-sdk-guide@781f230`, the first with an intact `liblouis.data`), verifying every file against its recorded size and MD5, including the LGPL-2.1 notice and liblouis wrapper sources the vendor's README asks redistributors to keep beside the engine. It writes a `manifest.json` naming the commit, is idempotent, and lands in `tools::R_user_dir("maidr", "cache")` unless `maidr.dotpad_sdk_dir` / `MAIDR_DOTPAD_SDK_DIR` says otherwise.
- **`show()` and `save_html()`** then copy the directory into `lib/dotpad-sdk-3.0.2/` beside every `use_cdn = FALSE` document, as a dependency listed ahead of the bundle whose `head` declares the two `window.MAIDR_DOTPAD_*` globals with that relative path. Attached in `create_html_document()` only: the widget, knitr and Shiny paths render into `srcdoc` frames where a relative path has nothing to resolve against, and a self-contained R Markdown document would drop the copied files, so they keep to the URL options.
- A configured URL wins over a downloaded copy; a session that never downloaded the SDK gets exactly what it did before.
- The option docs, README and NEWS no longer say the licence forbids redistribution. The pin mirrors `src/service/dotPadSdk.json` in the maidr repository.

## Checks

- `tests/testthat/test-dotpad-config.R` gains 12 tests: the pin's shape, directory resolution, the download (mocked fetch; verified, idempotent, `force`, a short file refused by size, a same-size corruption refused by digest, nothing written on failure), completeness, and the document (copy lands in `lib/`, globals precede the bundle, `use_cdn = TRUE` and a configured URL leave it out, an undownloaded session is unchanged).
- `roxygen2::roxygenise()` with the pinned 8.1.0 regenerated `man/` and `NAMESPACE`; `lintr` is clean on the touched files.
- Full `testthat` run locally: everything passes except two pre-existing environmental failures (`MASS` not installed for `geom_density_2d`, and the fallback image device in this container), neither touching this change.

https://claude.ai/code/session_01Jbm23mXb8ovppeXmTLFRBc

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jbm23mXb8ovppeXmTLFRBc)_


jooyoungseo:
All three taken in `8e6c898`:

- **Pin ahead of the bundled maidr.js**: option (a). The manifest's docs now say the bundled 4.8.0 still falls back to the earlier commits until its next refresh, and why that does not change what a document loads once a copy is named or downloaded. The bump itself follows xability/maidr#1257's release.
- **`sha256` unverified**: checked now wherever base R can compute it, which is `tools::sha256sum()` from R 4.5 (behind `get0()`, so R 4.0 to 4.4 keep MD5 alone rather than a new dependency), with a test that a wrong SHA-256 is refused there.
- **`on.exit()` in the loop**: one cleanup of every `.part` file under the directory, registered once before the loop.

---

jooyoungseo:
Good catch, and a real one. `aab5e56`:

- **Asset-only URL vs. a downloaded copy**: `maidr_dotpad_local_dependency()` now stays out whenever either `maidr.dotpad_sdk_url` or `maidr.dotpad_asset_base_url` is set, with a test for each combination that also asserts `MAIDR_DOTPAD_ASSET_BASE_URL` is declared exactly once. The same gap existed in py-maidr's `attach_local_dotpad_sdk()` and is fixed there too (xability/py-maidr#784).
- **Unbounded download**: `curl::curl_download()` now runs on a handle with `connecttimeout = 30` and a low-speed abort (under 1 KB/s for 60 s), so a stalled connection fails rather than hangs while a slow link still gets the 14 MB engine.
- **`manifest.json` in `lib/`**: kept on purpose. It is a few hundred bytes naming the commit the copy came from, so a folder found later on a server can be traced to a pin; that is what it is for.
- **Concurrent downloads into one directory**: agreed it is not worth a lock for a one-shot, user-invoked call; noted and left.

---

PR #308, opened 2026-09-16
feat: read the DotPad SDK pins from the shipped manifest and move to 3.0.3

jooyoungseo:
## Summary

Follow-up to #307. The DotPad SDK pin moves from 3.0.2 to 3.0.3, and this package stops carrying its own hand-maintained copy of the pin.

`maidr.js` now ships its manifest in the npm package as `dist/dotpad-sdk.json` (xability/maidr#1259). The bundle refresh copies it to `inst/dotpad-sdk.json`, and `maidr_dotpad_sdk_manifest()` reads every pin from there, so the R side moves with the JavaScript one instead of drifting from it.

3.0.3 is served from `xability/dotpad-sdk-guide`, a mirror Dot Inc. permit MAIDR to keep: they publish this release only as a zip, which a browser cannot import a module out of, so the mirror carries the extracted files. Every one of them was verified byte for byte against the vendor's archive before the pin was written, and the manifest records that archive's repository, commit, path and SHA-256.

## Changes

- `inst/dotpad-sdk.json`: the manifest, as `maidr.js` publishes it.
- `R/dotpad_config.R`: version, commit, base URL, module, asset directory and the per-file sizes and digests all come from that file now (cached after the first read) instead of a hard-coded data frame. Function names and return shapes are unchanged.
- `.github/scripts/fetch-maidr-bundle.sh` and `tools/update-maidr-assets.R`: copy `dist/dotpad-sdk.json` when the release ships one, and say so when it does not, so a refresh against an older `maidr.js` leaves the committed manifest alone.
- `tests/testthat/test-dotpad-config.R`: the expected base URL is derived from the manifest's `repository` field rather than naming a repository, and a new test reads the JSON independently and checks the accessors against it. The `liblouis.data` size assertion stays.
- Docs, `NEWS.md`, `README.md` and the vignette: version-generic wording, and the pins' new home.

## Verification

Run locally in this checkout:

```
Rscript -e 'pkgload::load_all(); testthat::test_local(filter = "dotpad")'   # passes
Rscript -e 'lintr::lint_package()'                                          # no lints on the changed files
Rscript -e 'roxygen2::roxygenise()'                                         # no drift
bash -n .github/scripts/fetch-maidr-bundle.sh                               # ok
```

One test is skipped as designed: `tools::sha256sum()` arrived in R 4.5, and the digest check falls back to MD5 below that.

Two suites fail in this container, `test-contour-level.R` and `test-fallback-rendering.R`, and both fail identically on `main` without this change: the `MASS` package is not installed here, so `geom_density_2d()` cannot compute its stat. They are not related to this diff.

https://claude.ai/code/session_01Jbm23mXb8ovppeXmTLFRBc

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jbm23mXb8ovppeXmTLFRBc)_


jooyoungseo:
Both points were right, and both are fixed in `bf3fdff`.

The caching claim was mine and the code did not back it. `maidr_dotpad_sdk_manifest()` now parses `inst/dotpad-sdk.json` once into a package environment, which matters a little more than the file's size suggests: four functions call it independently on a render path, so it was four parses per document rather than one.

The per-file gap is the same one a review caught on the Python side, and I have closed it the same way in both places. The `jq` gate now checks every entry's positive `bytes` and its digests' lengths, and the read is split into `maidr_dotpad_read_manifest()`, which validates every field it publishes and names the offending file rather than letting `vapply` complain about a length mismatch. Your reading of what would have happened was exactly it.

Verified: the gate accepts the shipped manifest and rejects six malformed ones (missing digest, short digest, non-hex md5, string `bytes`, empty `files`, empty `repository`); the R side rejects the same shapes with the file named; `testthat::test_local(filter = "dotpad")` passes with the usual R < 4.5 skip in this container; `lintr` finds nothing in either changed file; `roxygen2::roxygenise()` is in sync with the new `man/` page committed alongside.

Thanks for running the real `rcmdcheck` on R 4.6.1 with the `sha256sum` path exercised. That is the one thing this container cannot tell me.

---

jooyoungseo:
Finding 1 was a real bug and the worst kind, since it reads as success. Fixed in `bd88a37`, with your reproduction as the test.

I confirmed it before fixing it: an array-shaped `files` gave a manifest of eight files as zero rows and no error, and `all(logical(0))` is `TRUE`, so `maidr_dotpad_sdk_available()` would have called an empty directory a complete SDK, skipped the download and carried nothing into an offline document silently. That is the same all-of-nothing trap a reviewer caught on the Python side of this work, in the same shape, which is a fair sign it deserved the belt and braces you suggested:

- `files` must be an object keyed by unique, non-empty paths; an array is refused by name.
- A file entry that is not an object names the file rather than raising `subscript out of bounds`.
- `maidr_dotpad_sdk_available()` answers `FALSE` outright with no files to check, so the same trap cannot be reached by another route.

Finding 2 is in too. Both values are pasted straight onto a file's path and neither is built in R any more, so both must end in a slash, in `maidr_dotpad_read_manifest()` and in the `jq` gate. Your reading of the failure mode is right: joined URLs and a 404 per file, with nothing saying why.

The nit is done as well; the helper takes `file_path` now.

Verified: the R side rejects an array `files`, a scalar entry, and a missing slash on either field, each naming what is wrong, and the real manifest still reads eight files. The gate rejects the same four. `testthat::test_local(filter = "dotpad")` passes, `lintr` finds nothing in the changed files, and roxygen has no drift.

---

jooyoungseo:
Closed in `c2473ed`. You are right that it is the odd one out: every other field is checked down to digest length, and the one relationship between fields was not checked at all. `maidr_dotpad_read_manifest()` now requires `module` to be a key of `files`, the `jq` gate does the same, and a test covers it.

The consequence is worth spelling out for anyone reading later, because it is the quiet kind. The download would report success, `maidr_dotpad_sdk_available()` would say the copy is complete, and the saved document would point `sdk_url` at a file that was never fetched. The reader connects a DotPad, the SDK import fails, and the braille line drops to grade 1 with nothing saying why.

You are also right that the current generator cannot produce it: maidr's repin script builds `files` from the same selection that must contain the module, and refuses an archive without it. So this is the consumer agreeing with the producer rather than a live bug.

Worth noting for the record: the Python binding has the same shape of gap, where only a test over the shipped manifest asserts the module is listed rather than the parser doing it. That one is already merged, so I have left it; it is a small follow-up if you want the two bindings identical.

Verified: the real manifest passes both checks, a manifest naming `DotPadSDK-9.9.9.js` is rejected by both, `testthat::test_local(filter = "dotpad")` passes, `lintr` is clean on the changed files, and roxygen has no drift.

---

jooyoungseo:
Good catch, and you are right that it is the asymmetry that makes it worth fixing: this pull request checks a digest to the character and required a trailing slash because a missing one costs a 404, yet the one field that decides where bytes land on disk was unchecked. Fixed in `e05077d`.

`maidr_dotpad_read_manifest()` and the `jq` gate now both refuse a file path that is absolute, that names a home directory or a drive, that carries a backslash, or that has an empty, `.` or `..` segment. Verified against seven such paths in both the R reader and the gate, including `../../../.ssh/authorized_keys` and `lib/../../outside.js`, and against the eight paths the real manifest names, which all still pass.

I agree with your framing of the severity. Nothing user-authored reaches this today: the manifest is committed and only a maintainer regenerates it from a tarball whose checksum was already verified. What made me take it rather than note it is that the download is the one place this package writes files it did not name itself, so the check belongs there whatever the current trust chain looks like.

The `commit` nit was deliberate, and I have left it. The pin that matters is `baseUrl`, which every fetch is built from and which is checked; `commit` is there so a person can trace the files back to a tree, and the shipped manifest's format is pinned by a test.

The same gap exists in the Python binding, where `download_dotpad_sdk()` joins the manifest's path segments the same way. That one merged earlier today, so I am opening a follow-up there rather than leaving the two bindings different.

---

jooyoungseo:
Thank you for downloading the real SDK from the mirror and re-fetching `liblouis.data` outside the package to check it byte for byte. That file is the reason this pin exists at all, and an independent confirmation that the mirror serves the vendor's bytes is worth more than any assertion in the diff.

Your three points are all confirmations rather than asks, and I agree with each reading. The lowercase-hex assumption is shared deliberately between the R check and the shell gate rather than duplicated by accident; the cache has no invalidation hook because the value cannot change within a session, and the tests bypass it by reading a temp file through `maidr_dotpad_read_manifest()` directly; and `upstream` is provenance for a person, unread by any code path, which is why the reader does not validate it.

Merging: 14 checks green, six rounds of review handled.

---

PR #310, opened 2026-09-16
feat(ggplot2): read a rectangle layer as a gantt when its author declares one

### LINKED ISSUE #197: `geom_rect()` draws gantt charts too, but the rule for telling one from an annotation needs deciding

jooyoungseo:
Closes #197.

The `annotate()` half landed already. What was still open is the question the issue reserved: **is a `geom_rect()` layer whose bands partition a schedule?** "I would rather have this settled than guess at it, since guessing wrong announces decoration as data."

## Why the author is asked

The measurement posted on the issue is confirmed rather than revisited: five structural rules against eight charts, ggplot2 3.4.4.

```
chart                            R1     R2     R3     R6     R7     wanted
------------------------------------------------------------------------------
gantt (the target)               TRUE   TRUE   TRUE   TRUE   TRUE   gantt
gantt, tasks overlapping in x    TRUE   TRUE   TRUE   TRUE   TRUE   gantt
gantt, varying bar heights       TRUE   TRUE   FALSE  TRUE   FALSE  gantt
single annotate('rect')          TRUE   FALSE  FALSE  FALSE  FALSE  NOT
two-region highlight (1 layer)   TRUE   TRUE   TRUE   TRUE   TRUE   NOT
waterfall (lanes on x)           FALSE  FALSE  FALSE  FALSE  FALSE  NOT
heatmap via geom_rect            TRUE   TRUE   TRUE   FALSE  FALSE  NOT
heatmap with one cell missing    TRUE   TRUE   TRUE   TRUE   TRUE   NOT
------------------------------------------------------------------------------
correct                          4      5      4      6      5      / 8
```

All three of the issue's structural candidates claim a `geom_rect()`-drawn heatmap, which the issue lists as a must-not-claim. The best rule available, R6, still claims a gappy heatmap and a two-region highlight. A monotone waterfall and a one-task-per-lane schedule are the same rectangles, so there is nothing in the geometry left to separate.

**Rule 4 is the only candidate the measurement does not falsify**, so this takes it. The table is kept in a comment above the reading, so the reasoning survives the next person who wonders why a geometry rule is not used.

## `maidr_gantt()`

`geom_rect()` with one thing added: the author saying these rectangles are intervals in lanes.

```r
ggplot(tasks) +
  maidr_gantt(aes(xmin = start, xmax = end, ymin = lane - 0.4, ymax = lane + 0.4)) +
  scale_y_continuous(breaks = 1:3, labels = c("design", "build", "test"))
```

**Nothing about the picture changes.** Measured, the declared layer's built data is `identical()` to the bare `geom_rect()` layer's and the panel's `x.range`/`y.range` match, so swapping one call for the other moves nothing on the page. No discrete scale is trained, no band height is imposed, and `check.aes`/`check.param` stay on so a typo is reported exactly as `geom_rect()` reports it.

**Every chart already written keeps exactly the reading it has today.** An undeclared rect layer reaches `unread_layer_type()` — which is where `detect_layer_type()` already fell through to — pinned by a ten-fixture table asserting after == before, covering `annotate('rect')`→skip, bare `geom_rect`→unknown, heatmap/waterfall/highlight→unknown, and `geom_tile`/`geom_col`/`geom_segment`/`geom_point` untouched.

## Two carriers, because they fail in opposite directions

Both measured on ggplot2 3.4.4:

- `ggplot2::layer()` records `maidr_gantt` in `layer$constructor` — through one and two levels of user wrapper. A helper that called `geom_rect()` internally would record ggplot2's own name, which is why this is built with `layer()` directly.
- A field on the layer instance answers under `do.call(maidr_gantt, ...)`, where the closure at `constructor[[1]]` raises "cannot coerce type 'closure'".

Read field-first with the constructor as fallback, each in `tryCatch`. `GeomRect$maidr_type` stays NULL — the shared prototype is untouched, so no other rect layer is tagged.

## The processor half had to land in the same change

`segment_lane_axis()` is keyed on `x`/`xend`/`y`/`yend`, which a rect frame does not carry. Routing a declared layer to `"gantt"` without this would have shipped a **confident empty schedule** (`data lanes: 0 | lanes: NULL | selectors: 0`) — a false claim of a different kind, and one not visible from the detection diff alone.

`rect_gantt_frame()` renames the bounds into the frame the segment gantt already reads (`x = xmin`, `xend = xmax`, `y = yend = midpoint`) and then calls the landed functions unchanged. That also gives the degenerate guard for free: a zero-width rect normalises to level-on-both-axes and `segment_lane_axis()` refuses it.

Lanes are named by the single explicit tick drawn inside a band, and by position otherwise — the rule py-maidr#533 settled. A tick whose label is a rendering of its own number is a coordinate, not a name: on the default scale the panel labels are `NA, 1, 2, 3, NA`, each its own break written out.

## What this trusts — written down, and pinned by a test asserting the wrong answer

`maidr_gantt()` over heatmap coordinates is announced as a schedule. The author declared it and the package believes them. Any guard strong enough to catch that is the rule the measurements above ruled out, so this is a consequence of the decision rather than an oversight — and `test_that("a declared heatmap is announced as a schedule, and that is the hole", ...)` keeps it read rather than discovered.

## Also fixed, in scope

- `lane_ticks()` indexed `panel_params` by the **data** axis, but `panel_params` is keyed by the **drawn** axis. Under `coord_flip()` a declared schedule was named from the wrong axis — announcing `Jan Feb Mar` for lanes actually drawn `design build test`. Pinned, with the `geom_segment()` negative case asserted unchanged.
- The `GeomRect` branch is gated on `stat_class` so it cannot shadow the candlestick `StatRectCS`/`StatLinerangeBC` arms below it.
- The bar lookup matches on the `^geom_rect\.` **name** prefix. A class-keyed search resolves position 1 to `plot.background..rect.N`, because the theme draws rects too.

## The tidyquant claims are now measured, not cited

The two comments justifying `class(geom)[1]` over `inherits()`, and the `stat_class` guard, originally said they were read off tidyquant 1.0.12's source because tidyquant is a Suggests neither the dev environment nor CI installs. Installed it and measured (e70f8c9). Every claim holds, and one was vaguer than the source warrants:

```
inherits(GeomRectCS, "GeomRect")        TRUE
class(GeomRectCS)[1]                    "GeomRectCS"
inherits(GeomLinerangeBC, "GeomRect")   FALSE

geom_candlestick()  GeomLinerangeBC / StatLinerangeBC
                    GeomRectCS      / StatRectCS
geom_barchart()     GeomLinerangeBC / StatLinerangeBC
                    GeomSegmentBC   / StatSegmentLeftBC
                    GeomSegmentBC   / StatSegmentRightBC
```

`StatRectCS` is `geom_candlestick()`'s alone — `geom_barchart()` draws its bodies as segments, not rects — which the old wording ran together. With tidyquant present the existing candlestick tests stop skipping, and they are what covers the guard against real charts: `detect_layer_type()` still answers `candlestick` for the body layer and `skip` for the wick.

## Deliberately deferred

- **A generic `maidr_declare(layer, type =)`** was considered. It would have to wrap a layer built elsewhere, and a wrapper around `geom_rect()` loses the constructor carrier — so the generic form rides on the field alone. There is also only one ambiguous geom so far; the point to generalise is when a second one needs declaring.
- **`geom_grob_prefix()` counts by geom class**, so on `geom_col() + geom_rect()` the rect layer resolves to the col layer's grob. Pre-existing, it loses highlighting rather than mis-highlighting, and fixing it would widen this change. Two tests in `test-gantt-rect.R` now demonstrate it concretely, so it is worth its own issue.
- **`coord_flip()` still emits `orientation: "horz"`.** That is the package's settled position (it rotates the coordinate system and leaves the data layout alone), so it is consistent rather than new.

This is the package's first *per-layer* author declaration — every other `maidr_*` export acts on the session or the installation — so once exported it cannot be withdrawn without deprecation. Flagging that explicitly for review.

## Verification

- `testthat::test_local()` — **0 failures**. The absolute warning count is environment-dependent (12 where this was developed, 49 on the CI runner, which has more Suggests installed and a different ggplot2); what holds either way is the **delta: zero new warnings**, checked by diffing the warning set against `main` rather than comparing totals. New file `test-gantt-rect.R`, 26 `test_that` blocks.
- `tests/testthat/test-ggplot2-annotations.R:92`, which pins bare `geom_rect` at `"unknown"`, still passes **unchanged**.
- `Rscript tools/document.R` — exit 0, zero roxygen warnings, no drift.
- End-to-end: a declared chart renders as a real gantt with lanes named and selectors resolving in the exported SVG; the same schedule undeclared still falls back, which is the cost the issue measured.

https://claude.ai/code/session_01CAfHGG5HPt4qyz1sfMzkfB


jooyoungseo:
On the one point that named a gap — the `stat_class` guard being "currently unverified by any test — only synthetic layers ... are tested" — that isn't quite the position. The coverage exists; it was dormant.

`tests/testthat/test-ggplot2-candlestick-layer-processor.R:64` and `:77` already ask `detect_layer_type()` of a **real** `tidyquant::geom_candlestick()` body and wick layer and assert `"candlestick"` / `"skip"`. Those are exactly the assertions a wrong guard would break. They skip in CI only because tidyquant is a Suggests CI does not install.

So I installed it and ran them. tidyquant 1.0.12, ggplot2 still 3.4.4:

```
body  geom class[1]: GeomRectCS      | stat: StatRectCS      | detect -> candlestick
wick  geom class[1]: GeomLinerangeBC | stat: StatLinerangeBC | detect -> skip
```

`class(geom)[1]` is `GeomRectCS`, never `GeomRect`, so a real candlestick does not reach the new branch at all — the guard only ever mattered for the synthetic pairing. Full suite with those ~28 previously-skipped tests live: **0 failures**, and the only new warning sources are the three tidyquant files that had been skipping all along.

Pushed in e70f8c9, which also replaces the two "read off tidyquant's source rather than measured" comments with the measurement, since that caveat no longer had to stand. It turned up one thing the citation had run together:

```
geom_candlestick()  GeomLinerangeBC / StatLinerangeBC
                    GeomRectCS      / StatRectCS
geom_barchart()     GeomLinerangeBC / StatLinerangeBC
                    GeomSegmentBC   / StatSegmentLeftBC
                    GeomSegmentBC   / StatSegmentRightBC
```

`StatRectCS` is `geom_candlestick()`'s alone — `geom_barchart()` draws its bodies as segments, not rects. The old wording implied both produced it.

On the other two notes: agreed on both, and neither needs a change here. The `find_rect_name()` miscount stays out of scope, and the headless `maidr.js` `TypeError` you reproduced against a pre-existing `geom_segment()` gantt matches what I'd expect — it isn't this branch's.

---

